PR #7, opened 2024-07-15
feat: add llama vendor

kasrakhosravi:


vendors/llama.go
```diff
@@ -0,0 +1,44 @@
...
+	bodyMap, ok := jrr.(*common.JsonRpcResponse)
+	if !ok {
+		return nil
+	}
+
+	err := bodyMap.Error
+	code := err.OriginalCode()
+	msg := err.Message
+
+	if code == 0 && strings.Contains(msg, "error code: 1015") {
```
aramalipoor, line 32: Not sure if by default non-json responses (i.e. Llama's rate limits) will be properly hydrated into the error object. 

A unit test can show how it works. Maybe to have a fallback behavior when body is non-parsable to use the body as-is in the "message" field. That way this logic here will work.

vendors/llama.go
```diff
@@ -0,0 +1,44 @@
...
+		return common.NewErrEndpointUnauthorized(
+			common.NewErrJsonRpcException(code, common.JsonRpcErrorCapacityExceeded, msg, nil),
```
aramalipoor, line 34: There's a dedicated endpoint error for capacity. 'NewErrEndpointUnauthorized' is not correct for here.

PR #8, opened 2024-07-15
test: empty response retries and network tests

kasrakhosravi:


erpc/networks.go
```diff
@@ -186,10 +186,17 @@ func (n *Network) Forward(ctx context.Context, req *upstream.NormalizedRequest)
...
+			lg.Debug().Msg("Entered GetWithExecution")
+
 			n.upstreamsMutex.RLock()
 			upsList := n.Upstreams
 			n.upstreamsMutex.RUnlock()
 
+			if upsList == nil {
+				lg.Error().Msg("upsList is nil")
+				return nil, errors.New("upsList is nil")
+			}
```
aramalipoor, line 198: This shouldn't happen here, because u.Upstreams must never be null at this stage. We can have this check during network initialization.

erpc/networks_test.go
```diff
@@ -1129,3 +1147,394 @@ func TestNetwork_WeightedRandomSelect(t *testing.T) {
...
+	gock.New("http://rpc1.localhost").
+		Post("").
+		Reply(200).
+		JSON(json.RawMessage(`{"result":[]}`)).
+		Delay(500 * time.Millisecond)
+
+	gock.New("http://rpc2.localhost").
+		Post("").
+		Reply(200).
+		JSON(json.RawMessage(`{"result":{"hash":"0x64d340d2470d2ed0ec979b72d79af9cd09fc4eb2b89ae98728d5fb07fd89baf9","fromHost":"rpc2"}}`)).
```
aramalipoor, line 1309:
```suggestion
		JSON(json.RawMessage(`{"result":[{"logIndex":5,"fromHost":"rpc2"}]}`)).
```
Perhaps better to mimic array to be close to getLogs response?

erpc/networks_test.go
```diff
@@ -1129,3 +1147,394 @@ func TestNetwork_WeightedRandomSelect(t *testing.T) {
...
+		return shouldMakeRealCall
+	})
+
+	var requestBytes = json.RawMessage(`{"jsonrpc": "2.0","method": "eth_getLogs","params":[{"address":"0x1234567890abcdef1234567890abcdef12345678"}],"id": 1}`)
+
+	gock.New("http://rpc1.localhost").
+		Post("").
+		Reply(200).
+		JSON(json.RawMessage(`{"result":[]}`)).
+		Delay(500 * time.Millisecond)
```
aramalipoor, line 1304: We don't need a delay here, it should work without the delay

erpc/networks_test.go
```diff
@@ -1129,3 +1147,394 @@ func TestNetwork_WeightedRandomSelect(t *testing.T) {
...
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	// Setup test network
+	clr := upstream.NewClientRegistry()
+	fsCfg := &common.FailsafeConfig{
+		Hedge: &common.HedgePolicyConfig{
+			Delay:    "200ms",
+			MaxCount: 1,
+		},
```
aramalipoor, line 1321: We shouldn't have any policies for this test it can contaminate the results. This one is passing perhaps because of hedged request, not the actual "empty response handling" logic.

erpc/networks_test.go
```diff
@@ -1129,3 +1147,394 @@ func TestNetwork_WeightedRandomSelect(t *testing.T) {
...
+		t.Fatalf("Expected hash to be string, got %T", result["hash"])
+	}
+
+	expectedHash := "0x64d340d2470d2ed0ec979b72d79af9cd09fc4eb2b89ae98728d5fb07fd89baf9"
+	if hash != expectedHash {
+		t.Errorf("Expected hash to be %q, got %q", expectedHash, hash)
+	}
+}
+
+func TestNetwork_ForwardEthGetLogsBothEmptyArrayResponse(t *testing.T) {
```
aramalipoor, line 1421: Same comments above apply here.

upstream/failsafe.go
```diff
@@ -41,11 +41,16 @@ func CreateFailSafePolicies(scope Scope, component string, fsCfg *common.Failsaf
...
-		policies = append(policies, timeoutPolicy)
+		// For network-level we want the timeout to apply to the overall lifecycle
+		if scope == ScopeNetwork {
+			policies = append(policies, timeoutPolicy)
+		}
+
+		// For upstream-level we want the timeout to apply to each individual request towards upstream
+		if scope == ScopeUpstream {
+			policies = append(policies, timeoutPolicy)
+		}
```
aramalipoor, line 53: If policy is moved here it no more "apply to individual requests", is the original code is not working as expected?

aramalipoor: Oh I think I see original code for upstream-level is using wrong condition, maybe that was the root cause?

PR #10, opened 2024-07-16
feat: grafana charts

kasrakhosravi:
Screenshot 2024-07-17 at 09 36 13


aramalipoor (approved):
LGTM.

If it's easy/quick to make labels less verbose for easier reading that would be great.

e.g. doesn't need to say `{ method: 'eth_blockNumber...` probably just`eth_blockNumber` is enough, cuz it's obvious is the method.

PR #15, opened 2024-07-24
fix: alchemy rpc endpoint

KONFeature:
Correct the alchemy rpc endpoint rebuilt url for type `evm-alchemy` or for url starting with `evm+alchemy://xxx` or `alchemy://xxx` endpoints.

The `alchemyapi.io` endpoint don't work on all the chains, for example with `arb-sepolia` or `base-mainnet` it's ko

### DIFF
--- a/upstream/alchemy_http_json_rpc_client.go
+++ b/upstream/alchemy_http_json_rpc_client.go
@@ -142,7 +142,7 @@ func (c *AlchemyHttpJsonRpcClient) getOrCreateClient(network common.Network) (Ht
 		return nil, fmt.Errorf("unsupported network chain ID for Alchemy: %d", chainID)
 	}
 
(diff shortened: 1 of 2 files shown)


aramalipoor:
Good catch, haven't looked into it yet but does this domain work for all their chains? If not we can have a mapping for these different chains.

KONFeature:
On my side, on the Alchemy multi-chain dashboard, it's this endpoint asked for every chains 

But maybe we can go generic, and put an additional config param like alchemy endpoint suffix, for alchemy upstream config?

image

aramalipoor (approved):
LGTM 🚀

aramalipoor:
I think your change makes sense, I also don't see `alchemyapi.io` anymore. We can add that feature any time later.

Thanks for the PR mate 🙌

PR #25, opened 2024-08-02
test: retry on finalized and non-finalized blocks

kasrakhosravi:
- [x] unit test to not retry when block is finalized
- [x] unit test to retry when block is not finalized
- [x]  unit test to retry when block finalization is not available

### DIFF
--- a/erpc/evm_block_tracker.go
+++ b/erpc/evm_block_tracker.go
@@ -99,6 +99,8 @@ func (e *EvmBlockTracker) LatestBlock() uint64 {
 }
 
 func (e *EvmBlockTracker) FinalizedBlock() uint64 {
+	e.updateMu.Lock()
+	defer e.updateMu.Unlock()
 	return e.finalizedBlockNumber
 }
 
--- a/erpc/evm_block_tracker.go
+++ b/erpc/evm_block_tracker.go
@@ -14,7 +14,7 @@ import (
 type EvmBlockTracker struct {
 	network *Network
 
-	RWMutex              sync.RWMutex
+	mu              sync.RWMutex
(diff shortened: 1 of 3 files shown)


erpc/evm_block_tracker.go
```diff
@@ -99,6 +99,8 @@ func (e *EvmBlockTracker) LatestBlock() uint64 {
 }
 
 func (e *EvmBlockTracker) FinalizedBlock() uint64 {
+	e.updateMu.Lock()
+	defer e.updateMu.Unlock()
```
aramalipoor, line 103: Prob a RWMutex is better to use here because there might be multiple readers at the same time.

erpc/evm_block_tracker.go
```diff
@@ -14,7 +14,7 @@ import (
 type EvmBlockTracker struct {
 	network *Network
 
-	updateMu             sync.Mutex
+	RWMutex              sync.RWMutex
```
aramalipoor, line 17:
```suggestion
	mu              sync.RWMutex
```

erpc/networks_test.go
```diff
@@ -378,9 +378,597 @@ func TestNetwork_Forward(t *testing.T) {
...
+		}
+
+		result, ok := jrr.Result.([]interface{})
+		if !ok {
+			t.Fatalf("Expected Result to be []interface{}, got %T", jrr.Result)
+		}
+
+		if len(result) != 0 {
+			t.Fatalf("Expected empty result array")
+		}
```
aramalipoor, line 566: This must assert response is definitely from host 1 and not host 2

aramalipoor (approved):
LGTM 🚀

PR #36, opened 2024-08-13
feat: add Etherspot upstream for erc4337 requests

kasrakhosravi:
Screenshot 2024-08-14 at 15 04 07


upstream/clients.go
```diff
@@ -92,6 +93,12 @@ func (manager *ClientRegistry) CreateClient(ups *Upstream) (ClientInterface, err
 					clientErr = fmt.Errorf("failed to create Pimlico client for upstream: %v", cfg.Id)
 				}
 
+			case common.UpstreamTypeEvmEtherspot:
+				newClient, err = NewEtherspotHttpJsonRpcClient(ups, parsedUrl)
```
aramalipoor, line 97: We also need a condition in upstream.go "Upstream.guessUpstreamType()" for cases when type is not defined by the user.

PR #38, opened 2024-08-21
feat: thirdweb upstream vendor

kasrakhosravi:


upstream/thirdweb_http_json_rpc_client.go
```diff
@@ -0,0 +1,237 @@
...
+	"net/url"
+	"strconv"
+	"strings"
+	"sync"
+
+	"github.com/erpc/erpc/common"
+)
+
+var thirdwebNetworkSubdomains = map[int64]struct{}{
+	1:           {}, // Ethereum Mainnet
```
aramalipoor, line 15: Since thirdweb endpoints are predictable (using the numeric subdomain) we don't need to hardcode this list. We can simply make a request for eth_chainId and if response is valid we can consider that chain supported. There's an example of that in Pimlico upstream.

PR #41, opened 2024-08-23
fix: invalid normalization for eth_getStorageAt

xJonathanLEI:
The normalization code tries to normalize the block number for `eth_getStorageAt` but incorrectly thinks that the block number is the second parameter. It's actually the third. This causes any storage request to fail.


aramalipoor:
Thanks @xJonathanLEI for spotting, will merge and release 🙌

PR #42, opened 2024-08-24
fix: cache never used for `eth_chainId`

xJonathanLEI:
When setting response cache, `eth_chainId` has block ref set to `all`:

https://github.com/erpc/erpc/blob/f214de4266d0d86ac33356668ef0e4f32bf3a1aa/common/evm_block_ref.go#L158-L162

However, when fetching cache entry, `eth_chainId` falls under this branch:

https://github.com/erpc/erpc/blob/f214de4266d0d86ac33356668ef0e4f32bf3a1aa/erpc/evm_json_rpc_cache.go#L74-L76

This results in the chain ID request successfully getting cached but never used. This PR fixes it by setting the same block ref as in the response.


aramalipoor:
Good catch, I realized we only had this in response extraction, it makes more sense here in request extraction.

PR #58, opened 2024-09-05
test: prepare upstreams with unsupported network

kasrakhosravi:
if `SupportsNetwork` for certain upstreams can throw an error (e.g a bad response from `eth_chainId` method to validate the network support), we shouldn't throw and stop the flow and must go to the next upstream.

### DIFF
--- a/erpc/networks_test.go
+++ b/erpc/networks_test.go
@@ -4993,7 +4993,6 @@ func TestNetwork_Forward(t *testing.T) {
 		defer gock.CleanUnmatchedRequest()
 
 		var requestBytes = json.RawMessage(`{"jsonrpc": "2.0","method": "eth_getLogs","params":[{"address":"0x1234567890abcdef1234567890abcdef12345678"}],"id": 1}`)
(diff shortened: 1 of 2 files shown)


erpc/networks_test.go
```diff
@@ -4942,6 +4942,137 @@ func TestNetwork_Forward(t *testing.T) {
...
 		assert.Equal(t, "upstream-a", sortedUpstreamsCall[0].Config().Id, "Expected upstream-a to be preferred for eth_call in Phase 1")
 	})
+
+	t.Run("ForwardEnvioUnsupportedNetwork", func(t *testing.T) {
+		defer gock.Off()
+		defer gock.Clean()
+		defer gock.CleanUnmatchedRequest()
+
+		var requestBytes = json.RawMessage(`{"jsonrpc": "2.0","method": "eth_getLogs","params":[{"address":"0x1234567890abcdef1234567890abcdef12345678"}],"id": 1}`)
+		emptyResponse := json.RawMessage(`{"jsonrpc": "2.0","id": 1,"result":[]}`)
```
aramalipoor, line 4952: Probably it's better to test with a non-empty result and also the `fromHost` technique to make sure it's coming from rpc1

PR #65, opened 2024-09-17
fix: improve thread-safety of request handling for high load

aramalipoor:


zytek:
@aramalipoor is this mutex handling logic correct ? 

https://github.com/erpc/erpc/pull/65/files#diff-9312b7975e3419b1d373974839f59deffed6ba93c18a8551e078450450e0b6dbL44

looks like you added unnecessary lock/unlock in that function?

aramalipoor:
The main use of mutex in here was due to 'potential' concurrent result parsing e.g. when reading the result when setting the cache + reading the result when returning back to client.

I think the RLock/RUnlock part probably is redundant 🤔 

In the new PR I'm working on the approach is changed so this part of code is updated: https://github.com/erpc/erpc/pull/66/files#diff-9312b7975e3419b1d373974839f59deffed6ba93c18a8551e078450450e0b6dbR101
Your comment here is a good reminder for me to double check thread-safety there too. It would be amazing if you can look at that and let me know your thoughts :)

zytek:
I just re-read the change and saw you're using read lock and only then a read+write lock, didn't spot it first. 🤦 Might be the correct way after all.

BTW perhaps you could think of setting up CI with race detector? https://go.dev/doc/articles/race_detector

PR #67, opened 2024-10-08
fix: add blockHash support for certain eth_methods (eip-1898)

kasrakhosravi:
This PR updates the `ExtractEvmBlockReferenceFromRequest` for methods (referenced [here](https://eips.ethereum.org/EIPS/eip-1898)) that currently accept a default block parameter (`blockNumber` HEX string or tag such as `latest`) to additionally support block hashes (as an object).

* `"earliest"`
* `"0x0"`
* `{ "blockNumber": "0x0" }`
* `{ "blockHash": "0xd4e56740f876aef8c010b86a40d5f56745a118d0906a34e69aec8c0db1cb8fa3" }`

### DIFF
--- a/common/evm_block_ref.go
+++ b/common/evm_block_ref.go
@@ -104,13 +104,16 @@ func ExtractEvmBlockReferenceFromRequest(r *JsonRpcRequest) (string, int64, erro
 		if len(r.Params) > 1 {
 			switch secondParam := r.Params[1].(type) {
 			case string:
-				// Handle the second parameter as a HEX String
-				if strings.HasPrefix(secondParam, "0x") {
+				// Handle the 2nd parameter as a blockNumber HEX String
+				if len(secondParam) < 66 && strings.HasPrefix(secondParam, "0x") {
 					bni, err := HexToInt64(secondParam)
 					if err != nil {
 						return secondParam, 0, err
 					}
 					return strconv.FormatInt(bni, 10), bni, nil
+					// Handle the 2nd parameter as a blockHash HEX String
+				} else if len(secondParam) == 66 && strings.HasPrefix(secondParam, "0x") { // 32 bytes * 2 + "0x"
+					return secondParam, 0, nil
 				}
 				return "", 0, nil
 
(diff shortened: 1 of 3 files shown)


common/evm_block_ref.go
```diff
@@ -101,6 +97,54 @@ func ExtractEvmBlockReferenceFromRequest(r *JsonRpcRequest) (string, int64, erro
...
+
+				if blockHash, exists := secondParam["blockHash"]; exists {
+					if bh, ok := blockHash.(string); ok && strings.HasPrefix(bh, "0x") && len(bh) == 66 { // 32 bytes * 2 + "0x"
+						return bh, 0, nil
+					}
+					return "", 0, fmt.Errorf("unsupported 2nd parameter for method %s: %+v", r.Method, r.Params)
+				}
+
+				// If neither blockNumber nor blockHash is provided
+				return "", 0, fmt.Errorf("missing blockNumber or blockHash in second parameter object for ExtractEvmBlockReferenceFromRequest: %+v", r.Params)
```
aramalipoor, line 138:
```suggestion
				return "", 0, fmt.Errorf("missing blockNumber or blockHash in second parameter object for %s method: %+v",  r.Method, r.Params)
```

common/evm_block_ref.go
```diff
@@ -101,6 +97,54 @@ func ExtractEvmBlockReferenceFromRequest(r *JsonRpcRequest) (string, int64, erro
...
 
+	case "eth_getBalance",
+		"eth_getTransactionCount",
+		"eth_getCode",
+		"eth_call":
+		if len(r.Params) > 1 {
+			switch secondParam := r.Params[1].(type) {
+			case string:
+				// Handle the second parameter as a HEX String
+				if strings.HasPrefix(secondParam, "0x") {
```
aramalipoor, line 108: Do we need to consider block hash as well here?

e.g.:
```suggestion
				if len(secondParam) < 10 && strings.HasPrefix(secondParam, "0x") {
```

Or even further extend to return blockHash if len is 66, etc

common/evm_block_ref.go
```diff
@@ -130,16 +174,44 @@ func ExtractEvmBlockReferenceFromRequest(r *JsonRpcRequest) (string, int64, erro
...
+						bni, err := HexToInt64(bns)
+						if err != nil {
+							return bns, 0, err
+						}
+						return strconv.FormatInt(bni, 10), bni, nil
+					}
+					return "", 0, fmt.Errorf("unsupported 3rd parameter for method %s: %+v", r.Method, r.Params)
+				}
+
+				if blockHash, exists := thirdParam["blockHash"]; exists {
```
aramalipoor, line 202: A unit test would be valuable here, it's first time second param is an object

common/evm_block_ref.go
kasrakhosravi: if I understood correctly from (https://eips.ethereum.org/EIPS/eip-1898), the string can also be one of the below values (prob for backward. compatibility):

* HEX String - an integer block number
* String “earliest” for the earliest/genesis block
* String “latest” - for the latest canonical block
* String “pending” - for the pending state/transactions
* String “safe” - for the most recent safe block
* String “finalized” - for the most recent finalized block

Seems like `blockHash` is only supported in as part of the object, hence making if the param is either a reference to Data (`blockHash`) or Quantity (`blockNumber`)

aramalipoor: I think it's better to include this logic, to be robust against other client behavior, since it won't impact the existing expected flow and won't add too much cost.

Basically assume some unknown client/chain might still allow blockHash here.

Issue #51, opened 2024-08-28
Upstream master/backup


fdmota:
Does ERPC support upstream selection based on master/backup logic. The idea being that the requests should only be forwarded to the master unless the master is unavailable or as fall behind following the chain head or any other serious issues.
The use case is that the master is a local cheap node and latency is not critical.

aramalipoor:
With the current scoring mechanism only health aspects are considered (latency, rate limits, failures, etc). These scores will be used as a 'weight' when selecting an upstream, so a 'weighted random selection' happens using these scores.

I think your use-case is pretty common that you'd want to give higher weight to a particular upstream. There are few approaches I'm thinking about:
* **(1)** Allow users assign a weight (0.0 to 1.0) which intentionally skews the scores in favor of an upstream.
* or, **(2)** Introduce a concept like 'priority tiers' where you can define 'low' and 'high' tiers for example, and as long as even 1 upstream is is 'alive' stick with the same tier. Definition of 'alive' can be simple or complex (error rate threshold, latency threshold, etc)

Generally 'upstream selection' always brings up these questions:
* When exactly would you consider an upstream broken?
* How often and with what strategy would want to 'give an upstream a chance' to prove themselves?
* Do you sort upstreams and try from best ones? Or do you treat them on/off and completely disable the bad ones?

Basically a trade off of `Cost` (e.g. redundant/hedged calls, cpu/mem pressure, 3rd party provider billing) vs `Performance` (how fast requests are served) and `Resiliency` (how often we finally get a good response).

aramalipoor:
Also have thought about another idea to introduce a `standByRules` config on upstream-level:
```yaml
      - id: my-cheap-node-1
         endpoint: http://my-veeery-cheap-node1.svc.cluster.local
      - id: my-cheap-node-2
         endpoint: http://my-veeery-cheap-node2.svc.cluster.local
      - id: my-cheap-node-3
         endpoint: http://my-veeery-cheap-node2.svc.cluster.local

      - id: my-semi-somewhat-affordable-node
        endpoint: https://www.blastapi.io/xxxxxxxxx
        standByRules:
          - bringAsBackupIf:
              minAvailableUpstreamsLessThan: 3 # bring in if any of cheap ones are down
              errorRateHigherThan: 0.5 # OR if error rates are >=50% on other upstreams
              latencyP90HigherThan: 3s # OR if current upstreams have too high p90 latency

      - id: my-veeeery-expensive-node
[...]

PR #69, opened 2024-10-09
feat: add selection policy for upstreams 

aramalipoor:
* Allow influencing how upstreams are selected to serve (or not) traffic.
* A selection policy is defined on network-level, and is responsible to return list of upstreams that must be active.
* The main purpose of selectionPolicy is to define what is an accepted performance and/or what are the required conditions to select a node.
* selectionPolicy can be configured to run per-method and network, or per-network only.
* We'll create a "default" selectionPolicy that is activated if there are at least 1 upstream assigned to "fallback" group. This policy has a basic logic for errorRate and blockLag which can be also tuned via ROUTING_POLICY_MAX_ERROR_RATE, ROUTING_POLICY_MAX_BLOCK_HEAD_LAG, ROUTING_POLICY_MIN_HEALTHY_THRESHOLD. These envs allow you to tune this default logic without having to re-write the policy function.
* This PR also introduces Typescript/Javascript support for the whole config (as well as selection policy evaluation function)

Solves #51

## Use-cases
* When a node is lagging behind more X blocks, disable them entirely, until they're synced back.
* When an upstream has X% error-rate disable them and check them again on some interval to see if they're back up.
* Use "cheap" nodes first and disable "fast" nodes by default, and only if cheap nodes are 'broken' use fast nodes.

# Example

```yaml

projects:
- id: main 

  upstreams:
   - endpoint: cheap-1.com
   - endpoint: cheap-2.com
   - endpoint: fast-1.com
     # Each upstream can have an arbitrary group name which is used in metrics, as well as 
     # useful when writing an eval function in selectionPolicy below.
     group: fallback
   - endpoint: fast-2.com
     group: fallback

  networks:
    - evm:
        chainId: 1

      # Determines when to include or exclude upstreams depending on their health, performace
      # relative to other upstreams.
      selectionPolicy:
        # Every 1 minute evaluate which upstreams must be included,
        # based on the arbitrary logic (e.g. <90% error rate and <10 block lag):
        evalInterval: 1m
        # Freeform typescript-based logic to select upstreams to be included by returning them:
        evalFunction: |
          (upstreams, method) => {
            const defaults = upstreams.filter(u => u.config.group !== 'fallback')
            const fallbacks = upstreams.filter(u => u.config.group === 'fallback')
            
            const maxErrorRate = parseFloat(process.env.ROUTING_POLICY_MAX_ERROR_RATE || '0.7')
            const maxBlockHeadLag = parseFloat(process.env.ROUTING_POLICY_MAX_BLOCK_HEAD_LAG || '10')
            const minHealthyThreshold = parseInt(process.env.ROUTING_POLICY_MIN_HEALTHY_THRESHOLD || '1')
            
            const healthyOnes = defaults.filter(
              u => u.metrics.errorRate < maxErrorRate && u.metrics.blockHeadLag < maxBlockHeadLag
            )
            
            if (healthyOnes.length >= minHealthyThreshold) {
              return healthyOnes
            }

            return [...healthyOnes, ...fallbacks]
          }
        # To isolate selection evaluation and result to each "method" separately change this flag to true
        evalPerMethod: false
        # When an upstream is excluded you can give it a chance on a regular basis,
        # to handle certain amount of sample requests again, so that metrics are refreshed.
        # For example to see if error rate is improving after 5 minutes, or still too high.
        # This is conceptually similar to how a circuit-breaker works in "half-open" state.
        # Resampling is not always needed because "evm state poller" component will still make
        # requests for "latest" block which still updates errorRate.
        resampleExcluded: false
        resampleInterval: 5m
        resampleCount: 100

```


WesleyCharlesBlake:
so some comments here of "nice-to-haves"

1. for a specific upstream thats particularly "mixed" ie with archival/full nodes, and maybe some public nodes with rate limits, it would be nice if an archival/trace/debug request was routed to a node full node and returns an empty response, it gets "cordoned" for that method in the future, and will never be included to serve those heavy types of calls. (this would be the automated way)
2. second best would be to to mark each upstream as either full/archive , that way we can bundle all upstreams to one network, and only archival/debug/trace calls would be served by the corresponding upstreams (ie archive=true)
3. following this approach, maybe we can configure a block range limit per each upstream, so this would allow eth_getLogs on full nodes, provided its within that nodes block range

PR #70, opened 2024-10-09
feat: add interceptors to capture or override request and/or responses

aramalipoor:
### For debugging

Many times as a user I want to capture a specific request and/or response for debugging purpose. Using logs I end up having either too much noise, or not enough for the specific goal I have.

For example I want to capture only certain eth_* methods only if a certain response is being returned. Or I want to capture timeouts only for certain network, etc. With this module we're able to specific rules (field-based or script-based) to write those request/responses to a file.

---

### For overrides

Additionally we can allow script-based interceptors to override a response (or request) body for example when covering an edge-case on a certain chain or upstream. This allows encapsulating all the RPC-related logic in 1 place (i.e. eRPC config) in a web3 project, rather than spreading these logic across multiple places (backend, frontend, erpc, etc)

An example would be to parse EVM event topics/data or transactions call data arguments, adding USD values, etc, as part of the response, to avoid having extra logic on client-side.

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -19,13 +19,14 @@ var (
 
 // Config represents the configuration of the application.
 type Config struct {
-	LogLevel     string             `yaml:"logLevel" json:"logLevel"`
-	Server       *ServerConfig      `yaml:"server" json:"server"`
-	Database     *DatabaseConfig    `yaml:"database" json:"database"`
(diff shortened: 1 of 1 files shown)


aramalipoor:
Need to prioritize other features at the moment. Might come back to this later.

PR #92, opened 2024-11-13
feat: skip full-node upstreams when block is present in req and historical

kasrakhosravi:
Skip full-node upstreams when the request includes a block and is marked as historical, optimizing request handling.

For more efficiency, we can mark upstream node type (full, archive, light) automatically as part of the `detectfeatures` functionality, but will be handled in an upcoming PR.


upstream/upstream.go
```diff
@@ -655,10 +667,102 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+	if err == nil {
+		return false
+	}
+	errMsg := err.Error()
+	// Check for common error messages indicating missing historical state
+	return strings.Contains(errMsg, "missing trie node") ||
+		strings.Contains(errMsg, "required historical state unavailable") ||
+		strings.Contains(errMsg, "historical state unavailable") ||
+		strings.Contains(errMsg, "state unavailable") ||
+		strings.Contains(errMsg, "not found")
```
aramalipoor, line 726: We have a list in error normalization logic, probably it's best to extract these in common/evm.go and have common.EvmIsMissingDataError so we can use in both places

upstream/upstream.go
```diff
@@ -655,10 +667,102 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+
+func (u *Upstream) detectNodeType() (common.EvmNodeType, error) {
+	// using zero address for universality
+	address := "0x0000000000000000000000000000000000000000"
+
+	// Step 1: Attempt to get balance at block 0x1
+	balanceReqBlock1 := common.NewNormalizedRequest([]byte(fmt.Sprintf(
+		`{"jsonrpc":"2.0","id":1,"method":"eth_getBalance","params":["%s", "0x1"]}`,
+		address,
+	)))
```
aramalipoor, line 737: Probably eth_getBlockByNumber is better

upstream/upstream.go
```diff
@@ -655,10 +667,102 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+	errMsg := err.Error()
+	// Check for common error messages indicating missing historical state
+	return strings.Contains(errMsg, "missing trie node") ||
+		strings.Contains(errMsg, "required historical state unavailable") ||
+		strings.Contains(errMsg, "historical state unavailable") ||
+		strings.Contains(errMsg, "state unavailable") ||
+		strings.Contains(errMsg, "not found")
+}
+
+func (u *Upstream) detectNodeType() (common.EvmNodeType, error) {
```
aramalipoor, line 729: It's better if we get the last block and check these:
* latest block - 32 
* latest block - 128
* latest  block - 1024
* latest  block / 2
* block 1

and perhaps a "trace" or "receipts" call to get an idea if they're archival or not... we can start from furthest block and go up.

upstream/upstream.go
```diff
@@ -655,10 +667,102 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+		return fmt.Errorf("failed to get json-rpc request: %w", err), false
+	}
+
+	// sample of methods with block param
+	methodsWithBlockParam := map[string]int{
+		"eth_getBalance":          1,
+		"eth_getCode":             1,
+		"eth_getTransactionCount": 1,
+		"eth_call":                1,
+	}
```
aramalipoor, line 681: It's better to move these to common/evm.go as well, we're using similar info in evm block extraction, so let's see what are the ways to have this list in once place...

Also let's define this array as global const so it's not created per request, it's unnecessary memory usage

upstream/upstream.go
```diff
@@ -625,7 +625,19 @@ func (u *Upstream) detectFeatures() error {
 			u.supportedNetworkIds[util.EvmNetworkId(cfg.Evm.ChainId)] = true
 			u.supportedNetworkIdsMu.Unlock()
 		}
-		// TODO evm: check full vs archive node
+
+		if cfg.Evm.NodeType == "" {
+			nodeType, err := u.detectNodeType(&common.NormalizedRequest{})
```
aramalipoor, line 630: Any reason to create the req here? It can be done in isolation within "detectNodeType"

upstream/upstream.go
```diff
@@ -655,10 +667,101 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+						return fmt.Errorf("failed to get request block number: %w", ebn), false
+					}
+
+					if bn < lb-int64(u.config.Evm.MaxAvailableRecentBlocks) {
+						return fmt.Errorf("block number %d is out of range (must be >= %d)", bn, lb-int64(u.config.Evm.MaxAvailableRecentBlocks)), true
+					}
+				}
+			}
+		}
+	}
```
aramalipoor, line 703: Looking at body of req.EvmBlockNumber() do we even need to manually check with MethodsWithBlockParam?

This method seems to be already taking care of all edge-cases and blockNumber extraction. We can just check if "ebn" is above 0 and then try to fetch the "latestBlock" and so on...

upstream/upstream.go
```diff
@@ -655,10 +667,101 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+	for _, blockHeight := range blockHeights {
+		// Normalize the block height to hex format
+		hexBlockHeight, err := common.NormalizeHex(fmt.Sprintf("%x", blockHeight))
+
+		if err != nil {
+			return "", err
+		}
+
+		balanceReq := common.NewNormalizedRequest([]byte(fmt.Sprintf(
+			`{"jsonrpc":"2.0","id":1,"method":"eth_getBlockByNumber","params":["%s", false]}`,
```
aramalipoor, line 733: Better to use util.RandomID() instead of 1 for the "id"

upstream/upstream.go
```diff
@@ -654,11 +666,91 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+			return fmt.Errorf("failed to detect node type: %w", err), false
+		}
+		nodeType = common.EvmNodeType(detectedNodeType)
+	}
+
+	if nodeType == common.EvmNodeTypeFull {
+		lb := req.Network().EvmStatePollerOf(u.Config().Id).LatestBlock()
+
+		if bn < lb-u.config.Evm.MaxAvailableRecentBlocks {
+			return fmt.Errorf("block number %d is out of range (must be >= %d)", bn, lb-u.config.Evm.MaxAvailableRecentBlocks), true
```
aramalipoor, line 688: I think we can return common.MissingData... error here, so it's properly handled downstream

upstream/upstream.go
```diff
@@ -654,11 +666,91 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+		balanceReq := common.NewNormalizedRequest([]byte(fmt.Sprintf(
+			`{"jsonrpc":"2.0","id": %d,"method":"eth_getBlockByNumber","params":["%s", false]}`,
+			util.RandomID(),
+			hexBlockHeight,
+		)))
+
+		_, errBlock := u.Forward(context.Background(), balanceReq)
+		if errBlock == nil {
+			// No error, store response for this block height
+			responses[blockHeight] = true
```
aramalipoor, line 729: Probably it's better to also check if result exists and e.g. has result[hash] of that block, sometimes nodes might give no error but result is nil

erpc/networks_test.go
```diff
@@ -6068,6 +6068,37 @@ func TestNetwork_InFlightRequests(t *testing.T) {
 	})
 }
 
+func TestNetwork_SkippedUpstreams(t *testing.T) {
+	t.Run("ValidBlockNumberForFullNodeUpstream", func(t *testing.T) {
```
aramalipoor, line 6072: Good to add other cases for archive nodes, and unknown node types (default behavior, which must be 'not' skip imo)

upstream/upstream.go
```diff
@@ -654,11 +658,83 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
 
-	// TODO evm: if block can be determined from request and upstream is only full-node and block is historical skip
+	// if block can be determined from request and upstream is only full-node and block is historical skip
+	bn, ebn := req.EvmBlockNumber()
+	if ebn != nil || bn <= 0 {
+		return fmt.Errorf("failed to get valid request block number: %w", ebn), false
+	}
+
+	if u.config.Evm.NodeType == common.EvmNodeTypeFull {
+		lb := req.Network().EvmStatePollerOf(u.Config().Id).LatestBlock()
```
aramalipoor, line 668: We probably should check for nil of req.Network() (and maybe evm state poller too)

On initialization when we're detecting chain id of an upstream network is not assigned yet because that's what we're detecting. I guess this situation happens when upstream is defined but evm.chainId is not defined in the config.

upstream/upstream.go
```diff
@@ -654,11 +658,103 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
 	return nil, false
 }
 
+func (u *Upstream) detectNodeType() (common.EvmNodeType, error) {
+	// Step 1: Get the latest block number using the state poller
+	req := &common.NormalizedRequest{}
+	if req.Network() == nil || req.Network().EvmStatePollerOf(u.Config().Id) == nil {
+		return "", common.NewErrUpstreamClientInitialization(fmt.Errorf("network or state poller is nil"), u.config.Id)
+	}
+	lb := req.Network().EvmStatePollerOf(u.Config().Id).LatestBlock()
```
aramalipoor, line 688: We need to make sure by this point state poller is initialized and latest block is fetched at least once. If not done or straightforward we need to revise the related design.

upstream/upstream.go
```diff
@@ -654,11 +658,103 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+				responses[blockHeight] = true
+			} else {
+				responses[blockHeight] = false
+			}
+		} else if common.EvmIsMissingDataError(errForward) {
+			// Error due to missing historical state, continue checking
+			responses[blockHeight] = false
+		} else {
+			// Log the error for this specific block height
+			u.Logger.Debug().Err(errForward).Int64("blockHeight", blockHeight).Msg("Error getting balance")
```
aramalipoor, line 739:
```suggestion
			u.Logger.Debug().Err(errForward).Int64("blockHeight", blockHeight).Msg("could not get block during node type detection")
```

upstream/upstream.go
```diff
@@ -654,11 +658,103 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+	isFullNode := !isArchiveNode && (responses[lb-128] || responses[lb-32])
+
+	if isArchiveNode {
+		return "archive", nil
+	} else if isFullNode {
+		return "full", nil
+	}
+
+	// If none of the checks succeeded, it's likely a light node
+	return "light", nil
```
aramalipoor, line 755: Better to use the const instead of hardcoded strings. It is common.EvmNodeTypeFull or something like that iirc

upstream/upstream.go
```diff
@@ -654,7 +658,30 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
 		}
 	}
 
-	// TODO evm: if block can be determined from request and upstream is only full-node and block is historical skip
+	// if block can be determined from request and upstream is only full-node and block is historical skip
+	bn, ebn := req.EvmBlockNumber()
+	if ebn != nil || bn <= 0 {
+		return fmt.Errorf("failed to get valid request block number: %w", ebn), false
```
aramalipoor, line 664: same here https://github.com/erpc/erpc/pull/92/files#r1842575938

upstream/upstream.go
```diff
@@ -654,7 +658,30 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+		statePoller := ntw.EvmStatePollerOf(u.Config().Id)
+		if statePoller == nil {
+			return fmt.Errorf("state poller is nil"), false
+		}
+
+		lb := statePoller.LatestBlock()
+
+		if bn < lb-u.config.Evm.MaxAvailableRecentBlocks {
+			return common.NewErrEndpointMissingData(fmt.Errorf("block number %d is out of range (must be >= %d)", bn, lb-u.config.Evm.MaxAvailableRecentBlocks)), true
+		}
```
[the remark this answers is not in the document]
aramalipoor: For nil cases it's more robust if we return 'nil err and no skip' 🤔 ?

upstream/upstream.go
kasrakhosravi: totally. as discussed we want to have a Provider/Endpoint layer to handle detecting features at that level probably. we will handle this functionality in an upcoming PR

upstream/http_json_rpc_client.go
```diff
@@ -798,6 +776,16 @@ func extractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
...
+		} else if common.EvmDataIsOutOfRangeForFullNodeError(err) {
+			return common.NewErrEndpointDataOutOfRangeForFullNode(
+				common.NewErrJsonRpcExceptionInternal(
+					int(code),
+					common.JsonRpcErrorMissingData,
+					err.Message,
+					nil,
+					details,
+				),
+			)
```
aramalipoor, line 788: Will this be necessary? This error is not coming from upstreams themselves (from client) it happens on higher level, on network.forward

common/evm.go
```diff
@@ -25,3 +28,31 @@ type EvmStatePoller interface {
...
+		strings.Contains(err.Error(), "transaction not found") ||
+		strings.Contains(err.Error(), "cannot find transaction") ||
+		strings.Contains(err.Error(), "after last accepted block") ||
+		strings.Contains(err.Error(), "is greater than latest") ||
+		strings.Contains(err.Error(), "No state available")
+}
+
+func EvmDataIsOutOfRangeForFullNodeError(err error) bool {
+	return strings.Contains(err.Error(), "missing trie node")
+}
```
aramalipoor, line 58: In here why do we need to distinguish between these two groups?

erpc/networks_test.go
```diff
@@ -5844,6 +5844,124 @@ func TestNetwork_InFlightRequests(t *testing.T) {
...
+		gock.New("http://rpc1.localhost").
+			Post("/").
+			Persist().
+			Filter(func(request *http.Request) bool {
+				return strings.Contains(util.SafeReadBody(request), "eth_getBalance")
+			}).
+			Reply(200).
+			BodyString(`{"jsonrpc":"2.0","id":1,"result":"0x0"}`)
+		req := common.NewNormalizedRequest(requestBytes)
+		resp, err := network.Forward(context.Background(), req)
```
aramalipoor, line 5913:
```suggestion
		resp, err := network.Forward(ctx, req)
```
So that when test is finished request is also forced to cancel (without it we sometimes have gock race conditions)

erpc/networks_test.go
```diff
@@ -5908,3 +6026,136 @@ func setupTestNetwork(t *testing.T, ctx context.Context, upstreamConfig *common.
 
 	return network
 }
+
+func setupTestNetworkWithFullNodeUpstream(t *testing.T, ctx context.Context, upstreamConfig *common.UpstreamConfig, networkConfig *common.NetworkConfig) *Network {
```
aramalipoor, line 6030: Should we have 2 upstreams here (one full one archive) to properly that full nodes are skipped?

If we use upstream then we should expect the new "error" we created

common/errors.go
```diff
@@ -1567,6 +1567,20 @@ var NewErrEndpointMissingData = func(cause error) error {
 	}
 }
 
+type ErrEndpointDataOutOfRangeForFullNode struct{ BaseError }
+
+const ErrCodeEndpointDataOutOfRangeForFullNode = "ErrCodeEndpointDataOutOfRangeForFullNode"
```
aramalipoor, line 1572:
```suggestion
const ErrCodeUpstreamNodeTypeMismatch = "ErrUpstreamNodeTypeMismatch"
```
* I used "Endpoint" in errors that are returned from node itself (not a logic within erpc itself). So "Upstream" is probably a better one as it's because erpc itself thinks this node type is not good one (without even sending a request).
* If we use NodeType error, then we can have one or two fields in "details" of the error like (expectedNodeType: "archive" actualNodeType: "full") then this same error can be used in other places
* The word "Code" doesn't need to appear in the code string

common/errors.go
```diff
@@ -1567,6 +1567,20 @@ var NewErrEndpointMissingData = func(cause error) error {
...
+
+var NewErrEndpointDataOutOfRangeForFullNode = func(cause error) error {
+	return &ErrEndpointDataOutOfRangeForFullNode{
+		BaseError{
+			Code:    ErrCodeEndpointDataOutOfRangeForFullNode,
+			Message: "remote endpoint data is out of range for full node",
+			Cause:   cause,
+		},
+	}
+}
```
aramalipoor, line 1582: * I used "Endpoint" in errors that are returned from node itself (not a logic within erpc itself). So "Upstream" is probably a better one as it's because erpc itself thinks this node type is not good one (without even sending a request).
* If we use NodeTypeMismatch error, then we can have one or two fields in "details" of the error like (expectedNodeType: "archive" actualNodeType: "full") then this same error can be used in other places
* The word "Code" doesn't need to appear in the code string
```suggestion
type ErrUpstreamNodeTypeMismatch struct{ BaseError }

const ErrCodeUpstreamNodeTypeMismatch = "ErrUpstreamNodeTypeMismatch"

var NewErrUpstreamNodeTypeMismatch = func(cause error, expected EvmNodeType, actual EvmNodeType) error {
	return &ErrUpstreamNodeTypeMismatch{
		BaseError{
			Code:    ErrCodeEndpointDataOutOfRangeForFullNode,
			Message: "remote endpoint data is out of range for full node",
			Cause:   cause,
			Details: // ... {expectedNodeType: expected, ...
		},
	}
}
```

aramalipoor:
One place we can also update is https://github.com/erpc/erpc/blob/main/common/errors.go#L737-L738 where we're summarizing all upstream errors, this can be considered part of "X missing data". But we could also create a new category like "X node type mismatch" so that at least we take advantage of this new error for more clarity for the user?

upstream/upstream.go
```diff
@@ -722,7 +726,28 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
...
+			return nil, false
+		}
+
+		statePoller := ntw.EvmStatePollerOf(u.Config().Id)
+		if statePoller == nil {
+			return nil, false
+		}
+
+		if lb := statePoller.LatestBlock(); bn < lb-u.config.Evm.MaxAvailableRecentBlocks {
+			return common.NewErrUpstreamNodeTypeMismatch(fmt.Errorf("block number (%d) in request will not yield result for a fullNodeType upstream since it is not recent enough (must be >= %d)", bn, lb-u.config.Evm.MaxAvailableRecentBlocks), common.EvmNodeTypeArchive, common.EvmNodeTypeFull), true
```
aramalipoor, line 747: Should we also check if latest block exists? i.e. lb > 0

Also what happens if blockNumber is after the last recorded latestBlock? e.g. when state poller still is not aware of the new block, but user is making a request to get the data? 

Logically it should allow the request I think

upstream/upstream.go
```diff
@@ -722,7 +726,28 @@ func (u *Upstream) shouldSkip(req *common.NormalizedRequest) (reason error, skip
 		}
 	}
 
-	// TODO evm: if block can be determined from request and upstream is only full-node and block is historical skip
+	// if block can be determined from request and upstream is only full-node and block is historical skip
+	if u.config.Evm != nil && u.config.Evm.NodeType == common.EvmNodeTypeFull {
```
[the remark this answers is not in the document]
kasrakhosravi: hmm yeah that makes sense

upstream/upstream.go
```diff
@@ -693,7 +693,11 @@ func (u *Upstream) detectFeatures() error {
 			u.supportedNetworkIds[util.EvmNetworkId(cfg.Evm.ChainId)] = true
 			u.supportedNetworkIdsMu.Unlock()
 		}
-		// TODO evm: check full vs archive node
+
+		if cfg.Evm.MaxAvailableRecentBlocks == 0 && cfg.Evm.NodeType == common.EvmNodeTypeFull {
+			cfg.Evm.MaxAvailableRecentBlocks = 128
```
aramalipoor, line 698: Is this number universally correct? on all major eth clients etc

If this number is too high is better than too low, as we'll just failover if node is indeed missing the data

aramalipoor:
It's good to add/explain these 2 variables in the Upstream docs as well, so users know how they can mark nodes as archive/full

upstream/upstream.go
kasrakhosravi: `geth: By default, state for the last 128 blocks kept in memory`
https://github.com/ethereum/go-ethereum/releases/tag/v1.8.0

`Also known as pruned node, a full node maintains the blockchain's current state and data on the network, excluding trace data for transactions that are beyond the most recent 128 blocks`
https://erigon.gitbook.io/erigon/basic-usage/usage/type-of-node#full-node-or-pruned-node

kasrakhosravi: there was an interesting comment in `Erigon` docs e.g. related to full-nodes. could it be that the full will be marked as full, but can store beyond 128 blocks using `prune` command?

erpc/networks_test.go
kasrakhosravi: agreed. created a flexible test fixture builder for nodes types with `full`, `archive` and `unknown` type.

docs/pages/config/projects/upstreams.mdx
```diff
@@ -55,10 +55,16 @@ projects:
...
-          chainId: 42161]
+          # chainId is optional and will be detected from the endpoint (eth_chainId),
+          # but it is recommended to set it explicitly, for faster initialization.
+          chainId: 42161
+          # statePollerInterval is required and can be used to change the interval for checking the node's liveness. Default is 30s.
+          statePollerInterval: 5s
+          # nodeType is optional and you can manually set it to "full" or "archive"
+          nodeType: full
+          # maxAvailableRecentBlocks is optional and can be used to limit the maximum number of recent blocks to be served by this upstream. Default is 128 for full nodes.
+          maxAvailableRecentBlocks: 128
```
aramalipoor, line 67: was thinking maybe to capitalize optional
```suggestion
          # (OPTIONAL) chainId is optional and will be detected from the endpoint (eth_chainId),
          # but it is recommended to set it explicitly, for faster initialization.
          chainId: 42161
          # (OPTIONAL) statePollerInterval used to periodically fetch the latest/finalized/sync states. DEFAULT: 30s.
          statePollerInterval: 30s
          # (OPTIONAL) nodeType is optional and you can manually set it to "full" or "archive". DEFAULT: archive
          nodeType: full
          # (OPTIONAL) maxAvailableRecentBlocks limits the maximum number of recent blocks to be served by this upstream. DEFAULT: 128 (for "full" nodes).
          maxAvailableRecentBlocks: 128
```

upstream/upstream.go
aramalipoor: I see, ok so 128 is safest default then

erpc/networks_test.go
```diff
@@ -5844,6 +5844,194 @@ func TestNetwork_InFlightRequests(t *testing.T) {
...
+		requestBytes := []byte(`{"jsonrpc":"2.0","method":"eth_getBalance","params":["0x0000000000000000000000000000000000000000", "0x1"]}`)
+
+		gock.New("http://rpc1.localhost").
+			Post("").
+			Filter(func(request *http.Request) bool {
+				body := util.SafeReadBody(request)
+				return strings.Contains(body, "eth_getBalance")
+			}).
+			Reply(500).
+			JSON([]byte(`{"error":{"code":-32000,"message":"Internal error"},"id":1}`))
```
aramalipoor, line 5908: To be semantically correct:
```suggestion
			Reply(200).
			JSON([]byte(`{"result":[{"value":0x1,"fromHost":"rpc1"}]}`))```

File: common/config.go [hunk 28]
```diff
@@ -217,9 +217,10 @@ type JsonRpcUpstreamConfig struct {
 }
 
 type EvmUpstreamConfig struct {
-	ChainId             int         `yaml:"chainId" json:"chainId"`
-	NodeType            EvmNodeType `yaml:"nodeType,omitempty" json:"nodeType"`
-	StatePollerInterval string      `yaml:"statePollerInterval,omitempty" json:"statePollerInterval"`
+	ChainId                  int         `yaml:"chainId" json:"chainId"`
+	NodeType                 EvmNodeType `yaml:"nodeType,omitempty" json:"nodeType"`
+	StatePollerInterval      string      `yaml:"statePollerInterval,omitempty" json:"statePollerInterval"`
+	MaxAvailableRecentBlocks int64       `yaml:"maxAvailableRecentBlocks,omitempty" json:"maxAvailableRecentBlocks"`
```
aramalipoor, line 223: We can add `defaults.go` and `validation.go` for these fields.

docs/pages/config/projects/upstreams.mdx
kasrakhosravi: I thought based on this (https://github.com/erpc/erpc/blob/9fbb437d58e8aaca8c934a32f89ce130540d56b5/common/validation.go#L369), `statePollerInterval` is required?

common/defaults.go
```diff
@@ -252,6 +252,19 @@ func (e *EvmUpstreamConfig) SetDefaults() {
...
+	}
+
+	if e.MaxAvailableRecentBlocks == 0 {
+		switch e.NodeType {
+		case "full":
+			e.MaxAvailableRecentBlocks = 128
+		case "archive":
+			e.MaxAvailableRecentBlocks = 0
+		}
+	}
```
aramalipoor, line 267: This must be outside the "if"
```
		case "archive":
			e.MaxAvailableRecentBlocks = 0
```

PR #80, opened 2024-10-22
Feat: Redis modifications

OBlackmon3:
## What?
This PR adds TTL to specific methods and adds connection pool size configuration.
## Why?
TTL has been useful for our specific use case here and we added the connection pool size configuration to help debug a latency issue we are experiencing with batched requests.
## How?
This accomplished by adding SetTTL() and HasTTL() methods to the connector interface.
## Testing?
We've been running this in our environments for a few weeks with no issues.

Most of the contributions were completed by @dshiell and I helped with some debugging and testing.

PR #97, opened 2024-11-15
feat: add per-method and per-network caching policies and ttl

aramalipoor:
This PR is created off of the work by @dshiell and @OBlackmon3 in #80 

New structure supports:
```yaml
# ...
database:
  evmJsonRpcCache:
    connectors:
      - id: string
        driver: memory | redis | postgresql | dynamodb
        # ... (driver specific config)
    policies:
      - network: string
        method: string
        params: []any
        finality: finalized | unfinalized | unknown
        empty: ignore | allow | only
        minItemSize: string # xB | xKB | xMB
        maxItemSize: string # xB | xKB | xMB
        connector: string
        ttl: duration
```


common/config.go
```diff
@@ -59,24 +59,117 @@ type AdminConfig struct {
...
+			c.Policies = append(
+				c.Policies,
+				&CachePolicyConfig{
+					Network:   "*",
+					Method:    "*",
+					Finality:  DataFinalityStateUnknown,
+					TTL:       30 * time.Second,
+					Connector: "default",
+				},
+				&CachePolicyConfig{
```
kasrakhosravi, line 102: why this instance of `CachePolicyConfig` has been appended 2 times?

common/config.go
```diff
@@ -59,24 +59,117 @@ type AdminConfig struct {
...
+			{
+				Network:   "*",
+				Method:    "*",
+				Finality:  DataFinalityStateFinalized,
+				TTL:       0,
+				Connector: "default",
+			},
+		}
+		// PostgreSQL might not be very efficient for these short default TTLs
+		if rawBackward.Driver != DriverPostgreSQL {
```
kasrakhosravi, line 92: how about `DDB`?

common/evm_block_ref.go
```diff
@@ -7,6 +7,11 @@ import (
 	"strings"
 )
 
+// List of methods that static result is expected for. i.e. result will not change over time
+var EvmStaticMethods = map[string]bool{
+	"eth_chainId": true,
```
kasrakhosravi, line 12: how about these methods?

`net_version`:  Returns the current network ID
`web3_clientVersion`:  Returns the current client version.
`eth_protocolVersion`: Returns the current Ethereum protocol version.

common/evm_block_ref.go
```diff
@@ -7,6 +7,11 @@ import (
...
+}
+
 func ExtractEvmBlockReference(rpcReq *JsonRpcRequest, rpcResp *JsonRpcResponse) (string, int64, error) {
```
kasrakhosravi, line 15: is it not common to have named returned values in go? e.g it is easy to understand the requests of this function but not so much on the returned values

common/response.go
```diff
@@ -246,6 +266,46 @@ func (r *NormalizedResponse) EvmBlockNumber() (int64, error) {
...
+			// Static methods are not expected to change over time so we can consider them finalized
+			finality = DataFinalityStateFinalized
+			return
+		}
+	}
+
+	if r.request != nil {
+		ntw := r.request.Network()
+		if ntw != nil {
+			evmBn, _ := r.EvmBlockNumber()
```
kasrakhosravi, line 287: is it ok to ignore the `error` here?

common/evm_block_ref.go
aramalipoor: I thought about adding these as well, the biggest question is "if it correct to cache data from 'one of' upstreams" as you might have multiple upstreams each different values for some of these methods?

common/config.go
aramalipoor: DDB with native TTL is efficient to be used by default. In PGs case we have to run periodic intervals so is better to let user opt-in knowing the cost.

common/evm_block_ref.go
kasrakhosravi: hmm that is indeed an interesting problem. they are static in the context of 1 upstream node, but here it could be different for different nodes.

data/cache_policy.go
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	// When fetching data we need to iterate over all policies as we don't know finality of the data when originally written
+	// We will iterate from first to last policy (matched on network/method) to see which one has the data
+	// Therefore it is recommended to put the fastest most up-to-date policy first
+	return true
+}
+
+func (p *CachePolicy) GetConnector() Connector {
+	return p.connector
```
kasrakhosravi, line 54: noob go question. why we are not using a pointer to `p` here?

data/dynamodb.go
```diff
@@ -196,6 +202,19 @@ func createTableIfNotExists(
...
 
+	_, err = client.UpdateTimeToLive(&dynamodb.UpdateTimeToLiveInput{
+		TableName: aws.String(cfg.Table),
+		TimeToLiveSpecification: &dynamodb.TimeToLiveSpecification{
+			AttributeName: aws.String(cfg.TTLAttributeName),
+			Enabled:       aws.Bool(true),
+		},
+	})
+
+	if err != nil && !strings.Contains(err.Error(), "already enabled") {
```
kasrakhosravi, line 213: shouldn't this be `strings.Contains(err.Error(), "already enabled")`?

data/memory.go
```diff
@@ -54,50 +105,56 @@ func (m *MemoryConnector) Get(ctx context.Context, index, partitionKey, rangeKey
...
-
-	key := fmt.Sprintf("%s:%s", partitionKey, rangeKey)
-	m.cache.Remove(key)
-	return nil
-}
-
-func (m *MemoryConnector) deleteWithWildcard(_ context.Context, _, partitionKey, rangeKey string) error {
-	prefixPK := strings.TrimSuffix(partitionKey, "*")
-	prefixRK := strings.TrimSuffix(rangeKey, "*")
+func (m *MemoryConnector) cleanupExpired() {
```
kasrakhosravi, line 140: shouldn't we return an `error` result here from this func?

data/cache_policy.go
aramalipoor: p is a pointer already, so we're just getting a attribute

data/dynamodb.go
aramalipoor: The only allowed error is if TTL attr already exists

data/memory.go
```diff
@@ -54,50 +105,56 @@ func (m *MemoryConnector) Get(ctx context.Context, index, partitionKey, rangeKey
...
+	m.logger.Debug().Str("key", key).Msg("getting item from memory")
+
+	item, ok := m.cache.Get(key)
 	if !ok {
 		return "", common.NewErrRecordNotFound(fmt.Sprintf("PK: %s RK: %s", partitionKey, rangeKey), MemoryDriverName)
 	}
-	return value, nil
+
+	// Check if item has expired
+	if item.expiresAt != nil && time.Now().After(*item.expiresAt) {
```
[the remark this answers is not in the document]
aramalipoor: "not" "before" "expiry" vs "after" "expiry"

for me the latter is more intuitive :D

data/memory.go
aramalipoor: no error is really expected from here afaict 🤔

data/memory.go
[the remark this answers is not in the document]
kasrakhosravi: yeah I think the only thing I was thinking was if `time.Now()` equals `item.expiresAt`. I don't think `After` would cover that? also an edge-case

data/memory.go
kasrakhosravi: understood. but I saw an error was expected for the same func in `PG` driver, that is why I was curious, but maybe there is different

PR #102, opened 2024-11-20
fix: improve defaults for cache and selection policy

kasrakhosravi:


common/defaults.go
```diff
@@ -496,7 +496,7 @@ const DefaultPolicyFunction = `
 			return healthyOnes
 		}
 
-		return [...fallbacks, ...healthyOnes]
+		return [...fallbacks, ...upstreams]
```
aramalipoor, line 499:
```suggestion
		return upstreams
```

docs/pages/config/projects/networks.mdx
```diff
@@ -72,8 +72,8 @@ projects:
                 return healthyOnes
               }
 
-              // Otherwise, return all healthy upstreams plus all fallback ones.
-              return [...healthyOnes, ...fallbacks]
+              // Otherwise, return all upstreams plus all fallback ones
+              return [...fallbacks, ...upstreams]
```
aramalipoor, line 76:
```suggestion
              return upstreams
```

common/defaults.go
```diff
@@ -41,15 +41,15 @@ func (c *Config) SetDefaults() {
 						Method:      "*",
 						Finality:    DataFinalityStateUnknown,
 						MaxItemSize: util.StringPtr("1mb"),
-						TTL:         30 * time.Second,
+						TTL:         5 * time.Second,
```
aramalipoor, line 44: also we have a docs section that says "30s"

PR #103, opened 2024-11-21
fix: monitoring and metrics improvements (cache/latency)

kasrakhosravi:
Metrics: 

- [x]  More granular time buckets for histogram metrics (`upstream_request_duration_seconds` and `network_request_duration_seconds`)

Monitoring:
- [x] `Network-level cache total success (hits/misses)`
- [x] `Network-level cache total errors`
- [x] `Network-level cache set success`
- [x] `Network-level cache get success (hits/misses)`

Screenshot 2024-11-22 at 11 51 21


health/metrics.go
```diff
@@ -17,11 +17,24 @@ var (
...
+			0.5,    // 500 ms
+			1,      // 1 s
+			2.5,    // 2.5 s
+			5,      // 5 s
+			10,     // 10 s
+			30,     // 30 s
+			60,     // 60 s
+			300,    // 5 min
+			1800,   // 30 min
+			3600,   // 1 hour
```
aramalipoor, line 37: Don't think we need these 2? Very unlikely requests take that long?

health/metrics.go
```diff
@@ -17,11 +17,24 @@ var (
 		Name:      "upstream_request_duration_seconds",
 		Help:      "Duration of actual requests towards upstreams.",
 		Buckets: []float64{
-			0.1, // 100ms
-			0.5, // 500ms
-			1,   // 1s
-			5,   // 5s
-			10,  // 10s
+			0.0005, // 0.5 ms
+			0.001,  // 1 ms
```
aramalipoor, line 21: Also these are too short don't know if they're gonna be actually used

health/metrics.go
kasrakhosravi: hmm makes sense

aramalipoor:
We also have new metrics for cache set/get/hit/miss that include policy and few more things (finality state etc). It's good to have few panels for those.

PR #106, opened 2024-11-22
feat: Chained builder for the config

KONFeature:
Bringing a new chained builder for the typescript package.

Global modifications:

- Review Dockerfile stage, for a lighter end image (only keeping prod dependencies in the final one)
- Accept `erpc.js` file as valid config file
- Fix uint field parsing from js / ts
- Review the typescript package export (switch from `esbuild + tsc` to `tsup`, output dir from `lib/` to `dist/`, remove built file from git)
- Add more instruction in the typescript config package readme (builder instruction + performance consideration when using typescript for lighter image)

And in the core typescript package config, startup a new chained builder variant to easily build erpc config using strongly typed component.
Here is the methods added:

- `initErpcConfig(partial<Config> => Builder)`: Create a new config builder
- `Builder.addRateLimiters(rateLimiter | (context) => rateLimiter)`: Add a new rate limiter to the builder, providing strongly typed `rateLimiterBudget` in each other methods
- `Builder.decorate(decorationArgs | (context) => decorationArgs)`: Add some variable in the context (atm only `upstreams` and `networks` context are accepted, failsafe, circuit breaker etc coming later)
- `Builder.addProject(project | (context) => project)`: Add a new project to the config
- `Builder.build() => Config`: Craft the erpc config from the builder


common/script/compiler.go
```diff
@@ -17,7 +17,7 @@ func CompileTypeScript(entrypoint string) (string, error) {
 		Platform:    api.PlatformNode,
 		GlobalName:  "exports",
 		EntryPoints: []string{entrypoint},
-		Sourcemap:   api.SourceMapInline,
+		Sourcemap:   api.SourceMapNone,
```
aramalipoor, line 20: Would we have a better error messages with inline source maps?

KONFeature: Yep definitly, rolled back this change on the latest commit

PR #118, opened 2024-12-05
fix(selection-policy): use fallback rpc if healthy

somehowchris:
By reading the section on [Default fallback policy](https://docs.erpc.cloud/config/projects/selection-policies#default-fallback-policy) I would have assumed that my fallbacks are used as fallbacks but seemingly all upstreams are used.

With this change, for the default policy, if none of the non-fallback RPCs are considered healthy and there are RPCs considered as fallbacks; the same health-checking logic is applied and healthy fallbacks returned.

If none of these measures apply the same full `upstreams` list is returned. (per my understanding shouldn't even be that as I would rather get an error than serve old data but thats more than just a breaking change)

----
Following this pattern assures that defaults are tried, fallbacks are used when healthy and in the worst case use any of the rpcs considered by the scoring system, thus making use of the fallback group.


aramalipoor:
Thanks @somehowchris for looking into this.

This is an interesting and nuanced feature.

Probably I should've used "backup" wording in this default policy, vs 'fallback'.

The main reason this default policy is returning all upstreams instead of only fallbacks is to cover the use-case of 'if my cheap nodes are degraded, "add" few backup expensive nodes' in such case you actually don't want to be too harsh and completely exclude the cheap nodes. This would be fine because eRPC will still do failover on per request basis, and give a chance to cheap nodes.

In other use-cases such as 'i want to exclude weak/degraded nodes' you can change the policy function completely and say exclude nodes with error rate > 40%. (that's why we made it freeform function in the config)

For sure feel free to customize the policy in your setup, but regarding this default behavior wanted to explain it, what are your thoughts? 🤔

aramalipoor:
Sharing some convo done in Telegram for future reference:

image

---------

image

aramalipoor (approved):
LGTM 🚀

docs/pages/config/projects/networks.mdx
```diff
@@ -184,6 +198,15 @@ export default createConfig({
...
                 }
 
+                if (fallbacks.length > 0) {
+                  let healthyFallbacks = fallbacks.filter(
+                    u => u.metrics.errorRate < maxErrorRate && u.metrics.blockHeadLag < maxBlockHeadLag
+                  )
+                  if (healthyFallbacks.length > 0) {
+                    return healthyFallbacks
+                  }
+                }
```
aramalipoor, line 208:
```suggestion
                // If there are fallbacks defined, try to use them
                if (fallbacks.length > 0) {
                  // Apply same health filtering as default rpcs
                  let healthyFallbacks = fallbacks.filter(
                    u => u.metrics.errorRate < maxErrorRate && u.metrics.blockHeadLag < maxBlockHeadLag
                  )
                  
                  // If there are healthy fallbacks use them
                  if (healthyFallbacks.length > 0) {
                    return healthyFallbacks
                  }
                }
```

PR #123, opened 2024-12-16
fix: support js config file + more types

KONFeature:
Fixes:
 - Support `erpc.js` config file
 - Correct `uint` type parsing from js/ts
 - Split docker build to have lighter prod image
 - Correct `Duration` from go type in js/ts (previously it was a string, but wasn't parsed at runtime)
 
Features:
 - Add more generic types for better ts code completion (on upstreams, auth, rate limits, db connector)


aramalipoor (approved):
LGTM 🚀

PR #139, opened 2024-12-29
feat: config validation bin

KONFeature:
Adding a CLI binary to validate config file + provide a few stats / warn about the config

Stats provided:
 - Orphan rate limits budget (warn)
 - Rate limits + associated resources (debug)
 - Total number of projects (info)
 - Projects info (upstreams, networks) (info)

The CLI is exported via goreleaser for now, so it can be used by downloading the executable matching the computer spec on the release page

It could also be added to the typescript package via a small install script that would download the right executable depending on the dev computer

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,11 +19,11 @@ setup:
 
 .PHONY: run
 run:
-	@go run ./cmd/erpc/main.go
+	@go run ./cmd/erpc/main.go ./cmd/erpc/validate.go
 
 .PHONY: run-pprof
 run-pprof:
-	@go run ./cmd/erpc/main.go ./cmd/erpc/pprof.go
+	@go run ./cmd/erpc/main.go ./cmd/erpc/pprof.go ./cmd/erpc/validate.go
 
 .PHONY: run-fake-rpcs
 run-fake-rpcs:
--- a/Makefile
+++ b/Makefile
@@ -35,8 +35,7 @@ run-k6:
 
 .PHONY: build
 build:
-	@CGO_ENABLED=0 go build -ldflags="-w -s" -o ./bin/erpc-server ./cmd/erpc/main.go
-	@CGO_ENABLED=0 go build -ldflags="-w -s" -o ./bin/erpc-validate ./cmd/erpc/validate
+	@CGO_ENABLED=0 go build -ldflags="-w -s" -o ./bin/erpc-server ./cmd/erpc/main.go ./cmd/erpc/validate.go
 	@CGO_ENABLED=0 go build -ldflags="-w -s" -tags pprof -o ./bin/erpc-server-pprof ./cmd/erpc/*.go
 
 .PHONY: test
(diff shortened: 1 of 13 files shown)


test/fake_erpc.go
```diff
@@ -284,10 +273,9 @@ func prepareERPCConfig(fs afero.Fs, config StressTestConfig) (string, string, er
 // 	return upstreamsCfg
 // }
 
-func initializeERPC(fs afero.Fs, configPath string) error {
-	args := []string{"erpc-test", configPath}
+func initializeERPC(fs afero.Fs, cfg *common.Config) error {
```
aramalipoor, line 276:
```suggestion
func initializeERPC(cfg *common.Config) error {
```

erpc/http_server.go
```diff
@@ -37,20 +37,23 @@ type HttpServer struct {
 }
 
 func NewHttpServer(ctx context.Context, logger *zerolog.Logger, cfg *common.ServerConfig, admin *common.AdminConfig, erpc *ERPC) *HttpServer {
-	reqMaxTimeout, err := time.ParseDuration(*cfg.MaxTimeout)
-	if err != nil {
-		if cfg.MaxTimeout != nil && *cfg.MaxTimeout != "" {
-			logger.Error().Err(err).Msgf("failed to parse max timeout duration using 30s default")
+	err := error(nil)
```
aramalipoor, line 40:
```suggestion
	err := nil
```
should work I suppose?

cmd/erpc/main.go
```diff
@@ -30,16 +32,53 @@ func init() {
 func main() {
 	logger := log.With().Logger()
 
-	logger.Info().Msgf("starting eRPC version: %s, commit: %s", common.ErpcVersion, common.ErpcCommitSha)
+	// Define the flag for the config file
+	configFileFlag := &cli.StringFlag{
+		Name:     "config",
+		Usage:    "Config file to use (by default checking erpc.js, erpc.ts, erpc.yaml, erpc.yml)",
+		Required: false,
+	}
```
aramalipoor, line 40: To be backward compatible I think we should support arg-mode as well, like:
```
erpc myconf.yml
```
Perhaps we should deprecate it though on next 'major' release.

Makefile
```diff
@@ -35,7 +35,7 @@ run-k6:
 
 .PHONY: build
 build:
-	@CGO_ENABLED=0 go build -ldflags="-w -s" -o ./bin/erpc-server ./cmd/erpc/main.go
+	@CGO_ENABLED=0 go build -ldflags="-w -s" -o ./bin/erpc-server ./cmd/erpc/main.go ./cmd/erpc/validate.go
```
aramalipoor, line 38: Don't think we need this as the main() func only exists in main.go

aramalipoor: Hmm in that case how about moving it to erpc package? Keeping the main.go only entrypoint

PR #149, opened 2025-01-22
feat: allow removing cors headers for unknown origins

### LINKED ISSUE #148: bug: enabling cors setting leads to unwanted request filtering
**Describe the bug**
When enabling cors settings, erpc also starts filtering requests based on the whitelisted origins and returns a 403 error for non whitelisted origins.
This is not how cors is intended to work and leads to unwanted side effects with tools that use a Origin header, but do not enforce cors.

eg. Chrome extensions, which are not bound to cors, but send requests with `chrome-extension://<extension-id>` as origin.

Ideally you should not filter requests based on cors headers, just not reply any cors headers if the origin doesn't match, so the client can decide itself if its in a situation where cors is necessary.
If you rely on that filtering for some reason, you should at least make it an optional feature, which can be disabled while keeping cors headers active.

At the end, we'd like to prevent arbitrary websites from using our rpc,  but we don't want to over-block all kind of external applications.

**To Reproduce**
1. Use this erpc.yaml config:
```yaml
logLevel: debug

server:
  httpHostV4: "0.0.0.0"
  listenV6: true
  httpHostV6: "[::]"
  httpPort: 4000
  aliasing:
    rules:
      - matchDomain: "*"
        serveProject: "main"
        serveArchitecture: "evm"
        serveChain: "7078815900"

projects:
  - id: main
    cors:
      allowedOrigins:
        - "https://*.ethpandaops.io"
      allowedMethods: 
        - "GET"
        - "POST"
        - "OPTIONS"
      allowedHeaders:
        - "Content-Type"
    upstreamDefaults:
      ...
```

2. Make this request:
```bash
curl --location 'http://localhost:4000/' \
--header 'Content-Type: application/json' \
--header 'Origin: https://non-whitelisted-site.com' \
--data '{
    "params": [
        "0x1203319",
        false
    ],
    "id": 9199,
    "jsonrpc": "2.0"
}'
```

**Expected behavior**
The request should still be processed. The response should not contain cors headers as the origin isn't whitelisted.

kasrakhosravi:
Fixes #148


common/defaults.go
```diff
@@ -362,8 +362,9 @@ func (a *AdminConfig) SetDefaults() {
 	if a.CORS == nil {
 		// It is safe to enable CORS of * for admin endpoint since requests are protected by Secret Tokens
 		a.CORS = &CORSConfig{
-			AllowedOrigins:   []string{"*"},
-			AllowCredentials: util.BoolPtr(false),
+			AllowedOrigins:      []string{"*"},
+			BlockForeignOrigins: util.BoolPtr(true),
```
aramalipoor, line 366: Let's make it false by default

common/config.go
```diff
@@ -288,12 +288,13 @@ type NetworkDefaults struct {
...
 
 type CORSConfig struct {
-	AllowedOrigins   []string `yaml:"allowedOrigins" json:"allowedOrigins"`
-	AllowedMethods   []string `yaml:"allowedMethods" json:"allowedMethods"`
-	AllowedHeaders   []string `yaml:"allowedHeaders" json:"allowedHeaders"`
-	ExposedHeaders   []string `yaml:"exposedHeaders" json:"exposedHeaders"`
-	AllowCredentials *bool    `yaml:"allowCredentials" json:"allowCredentials"`
-	MaxAge           int      `yaml:"maxAge" json:"maxAge"`
+	AllowedOrigins      []string `yaml:"allowedOrigins" json:"allowedOrigins"`
+	BlockForeignOrigins *bool    `yaml:"blockForeignOrigins" json:"blockForeignOrigins"`
```
aramalipoor, line 292:
```suggestion
	NoHeadersForUnknownOrigins *bool    `yaml:"noHeadersForUnknownOrigins" json:"noHeadersForUnknownOrigins"`
```
probably more accurate name

common/validation.go
```diff
@@ -487,6 +487,14 @@ func (c *CORSConfig) Validate() error {
 	if c.AllowedOrigins == nil || len(c.AllowedOrigins) == 0 {
 		return fmt.Errorf("*.cors.allowedOrigins is required, add at least one allowed origin")
 	}
+	for _, origin := range c.AllowedOrigins {
```
kasrakhosravi, line 490: @aramalipoor with this in context (https://github.com/erpc/erpc/blob/3526fec18f815681cdf4bc469116fb56a7db632f/erpc/http_server.go#L585), is it valid to have this strict validation here?

aramalipoor: I don't think this validation is correct, there could be cases when user wants to allow the RPC endpoint be used anywhere. Without "*" they can't do it as the browser will essentially require them to provide all the sites in the world.

PR #151, opened 2025-01-26
feat: add initializer for all cache connectors

kasrakhosravi:
- Redis
    - [x] Initialization logic
    - [x] Unit test
    - [x] Integration test
- Postgres
    - [x] Initialization logic
    - [x] Unit test
    - [x] Integration test
- DynamoDB
    - [x] Initialization logic
    - [x] Unit test
    - [x] Integration test


util/initializer.go
```diff
@@ -487,8 +487,12 @@ func (i *Initializer) autoRetryLoop(ctx context.Context) {
 			return
 		}
 		if err != nil {
-			i.logger.Warn().Err(err).Int("state", int(state)).Msgf("initialization auto-retry failed, will retry in %v", delay)
+			i.logger.Warn().
+				Err(err).
+				Int("state", int(state)).
+				Msgf("initialization auto-retry failed with this error: %v, will retry in %v", err, delay)
```
aramalipoor, line 493: The `err` is already added above, it should be already printing

PR #155, opened 2025-01-28
feat: add support for proxies on upstream clients

kasrakhosravi:
Proxy Pools in eRPC enable centralized management of http(s)/socks5 proxies and round robin load balancing used by multiple upstreams. 

Screenshot 2025-01-28 at 15 39 58

### DIFF
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -29,7 +29,8 @@ type GenericHttpJsonRpcClient struct {
 	Url     *url.URL
 	headers map[string]string
 
-	proxyPool string
+	proxyPool         string
+	proxyPoolRegistry *ProxyPoolRegistry
 
 	projectId  string
 	upstreamId string
@@ -62,13 +63,15 @@ func NewGenericHttpJsonRpcClient(
 	upstreamId string,
 	parsedUrl *url.URL,
 	jsonRpcCfg *common.JsonRpcUpstreamConfig,
+	proxyPoolRegistry *ProxyPoolRegistry,
 ) (HttpJsonRpcClient, error) {
 	client := &GenericHttpJsonRpcClient{
-		Url:        parsedUrl,
-		appCtx:     appCtx,
-		logger:     logger,
-		projectId:  projectId,
-		upstreamId: upstreamId,
+		Url:               parsedUrl,
+		appCtx:            appCtx,
+		logger:            logger,
+		projectId:         projectId,
+		upstreamId:        upstreamId,
+		proxyPoolRegistry: proxyPoolRegistry,
 	}
 
 	// Default fallback transport (no proxy)
@@ -192,15 +195,8 @@ func (c *GenericHttpJsonRpcClient) shutdown() {
 }
 
 func (c *GenericHttpJsonRpcClient) getHttpClient() *http.Client {
-	proxyPoolRegistry, err := NewProxyPoolRegistry(common.GetConfig().ProxyPools, c.logger)
-	if err != nil {
-		c.logger.Warn().Str("pool", c.proxyPool).
-			Err(err).Msg("could not create proxy pool registry; using fallback httpClient")
-		return c.httpClient
-	}
-
-	if c.proxyPool != "" && proxyPoolRegistry != nil {
-		pool, err := proxyPoolRegistry.GetPool(c.proxyPool)
+	if c.proxyPool != "" && c.proxyPoolRegistry != nil {
+		pool, err := c.proxyPoolRegistry.GetPool(c.proxyPool)
 		if err != nil {
 			// If the pool does not exist, fallback
 			c.logger.Warn().Str("pool", c.proxyPool).
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -119,9 +119,7 @@ func NewGenericHttpJsonRpcClient(
 			client.headers = jsonRpcCfg.Headers
 		}
 
-		if jsonRpcCfg.ProxyPool != "" {
-			client.proxyPool = proxyPool
-		}
+		client.proxyPool = proxyPool
 	}
 
 	go func() {
@@ -194,8 +192,13 @@ func (c *GenericHttpJsonRpcClient) shutdown() {
 }
 
 func (c *GenericHttpJsonRpcClient) getHttpClient() *http.Client {
-	if c.proxyPool != nil && len(c.proxyPool.clients) > 0 {
-		return c.proxyPool.GetClient()
+	if c.proxyPool != nil {
+		client, err := c.proxyPool.GetClient()
+		if err != nil {
+			c.logger.Error().Err(err).Msgf("failed to get client from proxy pool")
+			return c.httpClient
+		}
+		return client
 	}
 
 	return c.httpClient
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -101,8 +101,8 @@ func NewGenericHttpJsonRpcClient(
 			client.headers = jsonRpcCfg.Headers
 		}
 
-		if jsonRpcCfg.ProxyUrl != "" {
-			client.proxyUrl = jsonRpcCfg.ProxyUrl
+		if jsonRpcCfg.ProxyPool != "" {
+			client.proxyUrl = jsonRpcCfg.ProxyPool
 		}
 
 		transport := &http.Transport{
@@ -112,8 +112,8 @@ func NewGenericHttpJsonRpcClient(
 		}
 
 		// If ProxyUrl is set, parse and attach it to the transport
-		if jsonRpcCfg.ProxyUrl != "" {
-			proxyURL, err := url.Parse(jsonRpcCfg.ProxyUrl)
+		if jsonRpcCfg.ProxyPool != "" {
+			proxyURL, err := url.Parse(jsonRpcCfg.ProxyPool)
 			if err != nil {
 				return nil, fmt.Errorf("invalid proxy URL: %w", err)
 			}
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -1155,3 +1155,7 @@ func getVendorSpecificErrorIfAny(
 
 	return vn.GetVendorSpecificErrorIfAny(rp, jr, details)
 }
+
+func (c *GenericHttpJsonRpcClient) UnderlyingHttpClient() *http.Client {
+	return c.getHttpClient()
+}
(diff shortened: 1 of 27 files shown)


clients/proxy_pool_registry.go
```diff
@@ -0,0 +1,114 @@
...
+		return r, nil
+	}
+
+	// Initialize each proxy pool
+	for _, poolCfg := range cfg {
+		pool, err := createProxyPool(poolCfg)
+		if err != nil {
+			return nil, err
+		}
+		r.pools.Store(poolCfg.ID, pool)
```
aramalipoor, line 60: Since we're writing only during initialization I don't think sync.Map is needed, it just adds overhead?

clients/http_json_rpc_client.go
```diff
@@ -524,14 +557,11 @@ func getJsonRpcResponseFromNode(rootNode ast.Node) (*common.JsonRpcResponse, err
...
-				int(common.JsonRpcErrorParseException),
-				cause,
-				"",
-			),
-		}
+		jrResp.Error = common.NewErrJsonRpcExceptionExternal(
+			int(common.JsonRpcErrorParseException),
+			cause,
+			"",
+		)
```
aramalipoor, line 564: Probably it's good to somehow pass the rawResult in the error, maybe via "Details" in case users need to investigate what was the broken result?

clients/http_json_rpc_client.go
```diff
@@ -181,6 +191,28 @@ func (c *GenericHttpJsonRpcClient) shutdown() {
 	c.processBatch(true)
 }
 
+func (c *GenericHttpJsonRpcClient) getHttpClient() *http.Client {
+	proxyPoolRegistry, err := NewProxyPoolRegistry(common.GetConfig().ProxyPools, c.logger)
```
aramalipoor, line 195: This seems to be running on 'every request' so it's not optimal to create a new pool registry everytime.

The better approach is to create the registry once during Client creation/init, and then here just to "access" it.

clients/http_json_rpc_client.go
```diff
@@ -181,6 +191,28 @@ func (c *GenericHttpJsonRpcClient) shutdown() {
...
+	}
+
+	if c.proxyPool != "" && proxyPoolRegistry != nil {
+		pool, err := proxyPoolRegistry.GetPool(c.proxyPool)
+		if err != nil {
+			// If the pool does not exist, fallback
+			c.logger.Warn().Str("pool", c.proxyPool).
+				Err(err).Msg("could not find proxy pool in registry; using fallback httpClient")
+			return c.httpClient
+		}
```
aramalipoor, line 209: We better explicitly fail during config validation if a ProxyPool ID does not exist (validations.go) GetPool shouldn't really fail practically.

clients/proxy_pool_registry.go
kasrakhosravi: yeah I think so. we just write to this in the initialization phase, so a basic lock should be enough there

clients/http_json_rpc_client.go
kasrakhosravi: Isn't it already passed in `cause`?

thirdparty/pimlico.go
```diff
@@ -209,7 +209,12 @@ func (v *PimlicoVendor) OwnsUpstream(ups *common.UpstreamConfig) bool {
 }
 
 func (v *PimlicoVendor) createClient(ctx context.Context, logger *zerolog.Logger, parsedURL *url.URL) (clients.HttpJsonRpcClient, error) {
-	client, err := clients.NewGenericHttpJsonRpcClient(ctx, logger, "n/a", "n/a", parsedURL, nil)
+	proxyPoolRegistry, err := clients.NewProxyPoolRegistry(common.GetConfig().ProxyPools, logger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create proxy pool registry: %v. will use fallback http client", err)
+	}
+
+	client, err := clients.NewGenericHttpJsonRpcClient(ctx, logger, "n/a", "n/a", parsedURL, nil, proxyPoolRegistry)
```
aramalipoor, line 217: We should have a singleton proxyPoolRegistry created on top-level (not on every vendor here), so probably outside of ERPC object (similar to RateLimitersRegistry)

Also we can already fetch and pass the "Pool" only to clients, they don't need to receive the full registry.

clients/http_json_rpc_client.go
```diff
@@ -91,24 +111,21 @@ func NewGenericHttpJsonRpcClient(
...
-				MaxIdleConns:        1024,
-				MaxIdleConnsPerHost: 256,
-				IdleConnTimeout:     90 * time.Second,
-			},
+		if jsonRpcCfg.ProxyPool != "" {
+			client.proxyPool = jsonRpcCfg.ProxyPool
+		}
+
+		if fetchedProxyPool != nil {
+			client.fetchedProxyPool = fetchedProxyPool
```
aramalipoor, line 128: Do we even need ProxyPool name? Why not just a `proxyPool *ProxyPool` passed object?

clients/http_json_rpc_client.go
```diff
@@ -358,8 +383,9 @@ func (c *GenericHttpJsonRpcClient) processBatch(alreadyLocked bool) {
 
 	c.logger.Debug().Str("host", c.Url.Host).RawJSON("request", requestBody).Interface("headers", httpReq.Header).Msgf("sending json rpc POST request (batch)")
 
-	// Make the HTTP request
-	resp, err := c.httpClient.Do(httpReq)
+	// pick the client from the proxy pool registry (if configured) or fallback
+	actualClient := c.getHttpClient()
+	resp, err := actualClient.Do(httpReq)
```
aramalipoor, line 388:
```suggestion
	resp, err := c.getHttpClient().Do(httpReq)
```
?

clients/http_json_rpc_client.go
```diff
@@ -587,7 +610,9 @@ func (c *GenericHttpJsonRpcClient) sendSingleRequest(ctx context.Context, req *c
 	}
 	c.logger.Debug().Str("host", c.Url.Host).RawJSON("request", requestBody).Interface("headers", httpReq.Header).Msg("sending json rpc POST request (single)")
 
-	resp, err := c.httpClient.Do(httpReq)
+	// pick the client from the proxy pool registry (if configured) or fallback
+	actualClient := c.getHttpClient()
+	resp, err := actualClient.Do(httpReq)
```
aramalipoor, line 615:
```suggestion
	resp, err := c.getHttpClient().Do(httpReq)
```

clients/proxy_pool_registry.go
```diff
@@ -0,0 +1,119 @@
...
+
+	// Initialize each proxy pool
+	for _, poolCfg := range cfg {
+		pool, err := createProxyPool(poolCfg)
+		if err != nil {
+			return nil, err
+		}
+		r.mu.Lock()
+		r.pools[poolCfg.ID] = pool
+		r.mu.Unlock()
```
aramalipoor, line 62: Since we only write on init and only read during operations, we probably don't even need a mutex. Shouldn't be any race conditions

thirdparty/pimlico.go
```diff
@@ -209,9 +212,17 @@ func (v *PimlicoVendor) OwnsUpstream(ups *common.UpstreamConfig) bool {
...
 func (v *PimlicoVendor) createClient(ctx context.Context, logger *zerolog.Logger, parsedURL *url.URL) (clients.HttpJsonRpcClient, error) {
-	client, err := clients.NewGenericHttpJsonRpcClient(ctx, logger, "n/a", "n/a", parsedURL, nil)
-	if err != nil {
-		return nil, err
+	if v.proxyPoolRegistry != nil {
+		proxyPool, err := v.proxyPoolRegistry.GetPool(parsedURL.Host)
+		if err != nil {
+			return nil, err
+		}
+		client, err := clients.NewGenericHttpJsonRpcClient(ctx, logger, "n/a", "n/a", parsedURL, nil, proxyPool)
```
aramalipoor, line 220: We don't need to pass proxyPoolRegistry to vendors at all, none of these clients are used for critical traffic. These are mainly one-off used to detect chainId.

If these vendors needed proxy in future we can make sure to pass the ProxyPool object maybe as part of VendorSettings.

For now I suggest to remove them.

clients/proxy_pool_registry.go
```diff
@@ -0,0 +1,119 @@
...
+	ID      string
+	clients []*http.Client
+	counter uint64 // Add atomic counter for round-robin
+}
+
+// returns a round-robin client from the pool.
+func (p *ProxyPool) GetClient() *http.Client {
+	if len(p.clients) == 0 {
+		// Fallback: if for some reason the pool has zero clients
+		return &http.Client{Timeout: 60 * time.Second}
```
aramalipoor, line 26: This path shouldn't occur I suggest to "panic()" because if it for some reason happens, it'll be hidden + creating too many Client objects in memory, and causing OOM or high CPU usage.

clients/http_json_rpc_client.go
```diff
@@ -91,24 +110,17 @@ func NewGenericHttpJsonRpcClient(
...
-	} else {
-		client.httpClient = &http.Client{
-			Timeout: 60 * time.Second,
-			Transport: &http.Transport{
-				MaxIdleConns:        1024,
-				MaxIdleConnsPerHost: 256,
-				IdleConnTimeout:     90 * time.Second,
-			},
+		if jsonRpcCfg.ProxyPool != "" {
+			client.proxyPool = proxyPool
```
aramalipoor, line 123:
```suggestion
		client.proxyPool = proxyPool
```
do we need this check?

clients/http_json_rpc_client.go
```diff
@@ -181,6 +193,14 @@ func (c *GenericHttpJsonRpcClient) shutdown() {
 	c.processBatch(true)
 }
 
+func (c *GenericHttpJsonRpcClient) getHttpClient() *http.Client {
+	if c.proxyPool != nil && len(c.proxyPool.clients) > 0 {
+		return c.proxyPool.GetClient()
```
aramalipoor, line 198: Probably it's better if internals are not exposed here, so GetClient() can return nil and we check that (vs `len(clients)` which is an internal detail)

clients/http_json_rpc_client.go
kasrakhosravi: good point. the proxyPoolsRegistry is already setup at a higher level, so this check does not accomplish anything

common/config.go
```diff
@@ -33,6 +33,7 @@ type Config struct {
 	Projects     []*ProjectConfig   `yaml:"projects" json:"projects"`
 	RateLimiters *RateLimiterConfig `yaml:"rateLimiters" json:"rateLimiters"`
 	Metrics      *MetricsConfig     `yaml:"metrics" json:"metrics"`
+	ProxyPools   []ProxyPoolConfig  `yaml:"proxyPools" json:"proxyPools"`
```
aramalipoor, line 36:
```suggestion
	ProxyPools   []*ProxyPoolConfig  `yaml:"proxyPools" json:"proxyPools"`
```

PR #170, opened 2025-02-05
test: add concurrent eth_getLogs benchmark test

kasrakhosravi:


erpc/networks_bench_test.go
```diff
@@ -297,3 +297,109 @@ func BenchmarkNetworkForward_RetryFailures(b *testing.B) {
...
+		"id": 1
+    }`)
+
+	b.SetParallelism(500)
+
+	b.ResetTimer()
+	b.RunParallel(func(pb *testing.PB) {
+		for pb.Next() {
+			fakeReq := common.NewNormalizedRequest(requestBytes)
+			_, err := ntw.Forward(ctx, fakeReq)
```
aramalipoor, line 398: This might not give us exact 500 concurrent requests based on GOMAXPROCS

So maybe having a for-loop around line 398 and trigger goroutines can make it happen?

I'm not fully sure though maybe `SetParallelism(500)` already gives us exactly what we need (i.e. in the logs you see ~500 multiplex requests)

PR #171, opened 2025-02-07
fix: skip getLogs range hook when blockHash (eip-234) parameter is provided

kasrakhosravi:
EIP: https://github.com/ethereum/EIPs/blob/master/EIPS/eip-234.md

### DIFF
--- a/architecture/evm/eth_getLogs.go
+++ b/architecture/evm/eth_getLogs.go
@@ -504,6 +504,10 @@ func requestHasBlockHash(r *common.NormalizedRequest) bool {
 		return false
 	}
 
+	if len(jrq.Params) == 0 {
+		return false
+	}
+
 	filter, ok := jrq.Params[0].(map[string]interface{})
 	if !ok {
 		return false
--- a/architecture/evm/eth_getLogs.go
+++ b/architecture/evm/eth_getLogs.go
@@ -20,23 +20,18 @@ import (
 // when looking for data too close to the lower-end of Full nodes, because they might have pruned the data already.
 var LowerBoundBlocksSafetyMargin int64 = 10
 
-func BuildGetLogsRequest(fromBlock, toBlock int64, address interface{}, topics interface{}, blockHash string) (*common.JsonRpcRequest, error) {
-	filter := make(map[string]interface{})
-
-	if blockHash != "" {
-		// Per EIP-234, if blockHash is present, fromBlock/toBlock should not be included.
-		filter["blockHash"] = blockHash
-	} else {
-		fb, err := common.NormalizeHex(fromBlock)
-		if err != nil {
-			return nil, err
-		}
-		tb, err := common.NormalizeHex(toBlock)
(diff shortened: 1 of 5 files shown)


architecture/evm/eth_getLogs.go
```diff
@@ -82,6 +87,39 @@ func upstreamPreForward_eth_getLogs(ctx context.Context, n common.Network, u com
...
 		return false, nil, nil
 	}
+
+	blockHash, ok := filter["blockHash"].(string)
+	if ok && blockHash != "" {
+		// EIP-234 => if blockHash is present, fromBlock or toBlock is NOT allowed
+		if _, hasFB := filter["fromBlock"]; hasFB {
+			jrq.RUnlock()
+			return true, nil, common.NewErrEndpointServerSideException(
+				fmt.Errorf("EIP-234 violation: blockHash and fromBlock cannot both be present for eth_getLogs"),
```
aramalipoor, line 97: These are client-side exceptions

aramalipoor:
When blockHash is provided we probably don't even need these eth_getLogs hooks, probably it's less opinionated if we just skip is such cases? i.e. if from/to block are provided and they're 0x to be handled

architecture/evm/eth_getLogs.go
kasrakhosravi: had it first as a client-side exception :) will fix

architecture/evm/hooks.go
```diff
@@ -50,6 +50,10 @@ func HandleUpstreamPreForward(ctx context.Context, n common.Network, u common.Up
 
 	switch method {
 	case "eth_getLogs":
+		// If the request has a blockhash, we can do a simple forward to upstream without sending it to the eth_getLogs hooks
+		if requestHasBlockHash(r) {
+			return false, nil, nil
+		}
```
aramalipoor, line 56: better move this logic inside the upstreamPreForward_eth_getLogs itself... this hook orchestration code doesn't need any logic leaked into it

architecture/evm/eth_getLogs.go
```diff
@@ -82,6 +87,31 @@ func upstreamPreForward_eth_getLogs(ctx context.Context, n common.Network, u com
 		jrq.RUnlock()
 		return false, nil, nil
 	}
+
+	blockHash, ok := filter["blockHash"].(string)
+	if ok && blockHash != "" {
```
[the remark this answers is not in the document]
aramalipoor: and therefore ethGetLogsSubRequest.blockHash will also never be used...?

architecture/evm/eth_getLogs.go
```diff
@@ -301,6 +306,7 @@ type ethGetLogsSubRequest struct {
 	toBlock   int64
 	address   interface{}
 	topics    interface{}
+	blockHash string
```
aramalipoor, line 309: Is this being used anywhere?

PR #173, opened 2025-02-11
fix: proper validation for duplicate upstreams

kasrakhosravi:
- [x] Validate duplicate upstreams in registry and return fatal error on duplicates.
- [x] Update initializer to mark tasks as fatal and transition to StateFatal without further retries.


common/initializer.go
```diff
@@ -122,7 +124,7 @@ func (t *BootstrapTask) Wait(ctx context.Context) error {
 				return ctx.Err()
 			case <-ch.(chan struct{}):
 				// The attempt ended. Check if we failed.
-				if TaskState(t.state.Load()) == TaskFailed {
+				if TaskState(t.state.Load()) == TaskFailed || TaskState(t.state.Load()) == TaskFatal {
```
aramalipoor, line 127:
```suggestion
				s := TaskState(t.state.Load())
				if s == TaskFailed || s == TaskFatal {
```

common/initializer.go
```diff
@@ -339,11 +349,17 @@ func (i *Initializer) State() InitializationState {
...
+		Int("fatal", fatal).
 		Msg("calculating initialization state")
 
 	if total == succeeded {
 		return StateReady
 	}
+
+	if fatal > 0 {
+		return StateFatal
+	}
```
aramalipoor, line 361: Don't think we want to mark the whole initializer as Fatal if only 1 task is Fatal... I'd suggest to consider it like Partial if others are successful

Imagine from 10 upstreams 2 are fatally broken, but others are fine, then the Network initializer must be considered "Partial"

common/initializer.go
```diff
@@ -391,16 +407,18 @@ func (i *Initializer) Stop(destroyFn func() error) error {
...
+	i.tasksMu.Unlock()
 
-	// Wait for auto-retry goroutine to finish
 	i.autoRetryWg.Wait()
 
-	// Now, wait for any tasks that might still be running to finish or fail.
+	// After auto-retry has finished, lock again for the rest
+	i.tasksMu.Lock()
+	defer i.tasksMu.Unlock()
+
```
aramalipoor, line 421: Any specific reason the locking strategy must change?

common/initializer.go
```diff
@@ -485,6 +503,12 @@ func (i *Initializer) autoRetryLoop(ctx context.Context) {
 		return
 	}
 
+	if i.State() == StateFatal {
+		i.logger.Error().Msg("initialization fatal state, stopping auto-retry")
+		i.autoRetryActive.Store(false)
+		return
+	}
```
aramalipoor, line 510: We should avoid retrying a specific task not the whole initializer (e.g. 2 upstreams are fatal but 5 upstreams have transient errors)

PR #195, opened 2025-03-06
fix: enhance eth_getBlockReceipts request reference handling

kasrakhosravi:
`eth_getBlockReceipts` supports passing an object containing `blockNumber` or `blockHash` as param as well. 

We were only supporting a basic blockNumber or blockTag as string, but with this change, we also support this as well.

```
{
  "jsonrpc": "2.0",
  "method": "eth_getBlockReceipts",
  "params": [
    {
      "blockHash": "0xe869f24fc3596bfaf3d046e5db65c0fc879f6e634fa702bc16f88b37032b19a8"
    }
  ],
  "id": 703
}
```

```
{
    "jsonrpc": "2.0",
    "id": 703,
    "result": [
        {
            "blockHash": "0xe869f24fc3596bfaf3d046e5db65c0fc879f6e634fa702bc16f88b37032b19a8",
            "blockNumber": "0x14f87d8",
            "contractAddress": null,
            ...
}
```

### DIFF
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -294,7 +294,11 @@ var DefaultWithBlockCacheMethods = map[string]*CacheMethodConfig{
 		ReqRefs: SecondParam,
 	},
 	"eth_getBlockReceipts": {
-		ReqRefs:  FirstParam,
+		ReqRefs: [][]interface{}{
+			{0},
+			{0, "blockHash"},
+			{0, "blockNumber"},
+		},
 		RespRefs: BlockNumberOrBlockHashParam,
 	},
 	"trace_block": {
--- a/erpc/projects.go
+++ b/erpc/projects.go
@@ -147,7 +147,7 @@ func (p *PreparedProject) Forward(ctx context.Context, networkId string, nq *com
 
 	if err == nil || common.IsClientError(err) || common.HasErrorCode(err, common.ErrCodeEndpointExecutionException) {
 		if err != nil {
-			lg.Error().Err(err).Msgf("finished forwarding request for network with some client-side exception")
+			lg.Info().Err(err).Msgf("finished forwarding request for network with some client-side exception")
 		} else {
 			if lg.GetLevel() == zerolog.TraceLevel {
 				lg.Info().Object("response", resp).Msgf("successfully forwarded request for network")
--- a/erpc/projects.go
+++ b/erpc/projects.go
@@ -147,7 +147,7 @@ func (p *PreparedProject) Forward(ctx context.Context, networkId string, nq *com
 
 	if err == nil || common.IsClientError(err) || common.HasErrorCode(err, common.ErrCodeEndpointExecutionException) {
 		if err != nil {
-			lg.Info().Err(err).Msgf("finished forwarding request for network with some client-side exception")
+			lg.Error().Err(err).Msgf("finished forwarding request for network with some client-side exception")
 		} else {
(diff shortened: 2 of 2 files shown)


aramalipoor:
The client-side exceptions are intentionally info to prevent noise, as there will be many cases like tx reverts or random client calls (exploiters) that does not mean anything is broken on erpc/providers

Any reason to make them error?

kasrakhosravi:
> The client-side exceptions are intentionally info to prevent noise, as there will be many cases like tx reverts or random client calls (exploiters) that does not mean anything is broken on erpc/providers
> 
> Any reason to make them error?

Good point. The reasoning I had in mind was this error I encountered for `envio`:

```
[
  {
    "level": "info",
    "component": "proxy",
    "projectId": "main",
    "networkId": "evm:1",
    "body": {
      "jsonrpc": "2.0",
      "method": "eth_getBlockReceipts",
      "params": [
        {
          "blockHash": "0xe869f24fc3596bfaf3d046e5db65c0fc879f6e634fa702bc16f88b37032b19a8"
        }
      ],
      "id": 703
    },
    "time": 1741282162568,
    "message": "received http request"
  },
  {
    "level": "info",
    "projectId": "main",
    "component": "proxy",
    "networkId": "evm:1",
    "method": "eth_getBlockReceipts",
    "id": 703,
    "ptr": "0x4000238210",
    "error": {
      "code": "ErrUpstreamsExhausted",
      "message": "all upstream attempts failed (1 user errors)",
      "cause": [
        {
          "code": "ErrUpstreamRequest",
          "message": "failed to make request to upstream",
          "details": {
            "method": "eth_getBlockReceipts",
            "upstreamId": "envio-evm:1",
            "attempts": 1,
            "retries": 0,
            "hedges": 0,
            "durationMs": 151,
            "networkId": "evm:1"
          },
          "cause": {
            "code": "ErrEndpointClientSideException",
            "message": "client-side error when sending request to remote endpoint",
            "cause": {
              "code": "ErrJsonRpcExceptionInternal",
              "message": "Invalid params: invalid type: map, expected BlockNumber, 'latest', or 'earliest'",
              "details": {
                "statusCode": 200,
                "headers": {
                  "content-type": "application/json",
                  "date": "Thu, 06 Mar 2025 17:29:22 GMT"
                },
                "data": null,
                "originalCode": -32602,
                "normalizedCode": -32602
              },
              "cause": null
            }
          }
        }
      ],
      "details": {
        "attempts": 19,
        "durationMs": 151,
        "hedges": 14,
        "method": "eth_getBlockReceipts",
        "networkId": "evm:1",
        "projectId": "main",
        "retries": 4
      }
    },
    "time": 1741282162720,
    "message": "finished forwarding request for network with some client-side exception"
  }
]
```

I thought as a dev, I need to know that this upstream does not support sending `blockHash` as param to `eth_getBlockReceipts`. If it was at the `Error` level, probably I would have caught it easier rather than `Info`.

One idea for me is to set an alert in Grafana on client side exceptions; there could be interesting cases that are not merely user error but lack of standard in different upstreams for how they should treat different requests.

Reverting this change for now.

aramalipoor:
I see, for such error we actually another problem as well, which is since it's a Client error it'll not be retried properly towards other upstreams. I remember another similar case as well, a 4xx that must've been retried towards other upstreams but not the current one.

Probably it's good to have a task to create a mechanism for such retryable client-side errors.

In terms of observability I think having it as INFO is enough since user will also see it in responses as well during debugging sessions.

PR #203, opened 2025-03-10
fix: retryable client side errors mechanism

kasrakhosravi:
Since when we have client error it'll not be retried properly towards other upstreams, it would be useful to have a mechanism for such retryable client-side errors.

### DIFF
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -333,20 +333,12 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 		// 11) "Invalid Argument / Params" errors
 		//----------------------------------------------------------------
 
-		if code == -32602 ||
-			strings.Contains(msg, "param is required") ||
-			strings.Contains(msg, "Invalid Request") ||
-			strings.Contains(msg, "validation errors") ||
-			strings.Contains(msg, "invalid argument") ||
-			strings.Contains(msg, "invalid params") ||
-			strings.Contains(msg, "tx of type") ||
-			strings.Contains(msg, "invalid type") {
-
-			// by default, we retry this type of client-side exception, as the root cause
-			// might be an invalid argument, that another upstream might support.
-			//
-			// Examples:
-			// - Envio not supporting { blockNumber: XXX, blockHash: YYY } for eth_getBlockReceipts
+		// Even though these errors have invalid argument or params, they are more about a lack of standard
+		// for certain args/params for certain methods, among different providers or clients. We will retry them.
+		//
+		// Examples:
+		if strings.Contains(msg, "tx of type") || // Should be retried toward upstreams that support this tx type
+			strings.Contains(msg, "invalid type: map, expected BlockNumber, 'latest', or 'earliest'") { // Envio not supporting { blockNumber: XXX, blockHash: YYY } for eth_getBlockReceipts
 			return common.NewErrEndpointClientSideException(
 				common.NewErrJsonRpcExceptionInternal(
 					int(code),
@@ -356,6 +348,26 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					details,
 				),
 				true, // retryable
+			)
+		} else if code == -32602 ||
+			strings.Contains(msg, "param is required") ||
+			strings.Contains(msg, "Invalid Request") ||
+			strings.Contains(msg, "validation errors") ||
+			strings.Contains(msg, "invalid argument") ||
+			strings.Contains(msg, "invalid params") ||
+			strings.Contains(msg, "tx of type") {
+
+			// For invalid args/params errors, there is a high chance that the error is due to a mistake that the user
+			// has done, and retrying another upstream would not help.
+			return common.NewErrEndpointClientSideException(
+				common.NewErrJsonRpcExceptionInternal(
+					int(code),
+					common.JsonRpcErrorInvalidArgument,
+					err.Message,
+					nil,
+					details,
+				),
+				false, // not retryable
 			)
 		}
 
@@ -381,8 +393,8 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					}
 				}
 			}
-			// We are going to retry this type of client-side exception, as the root cause
-			// might be an invalid request, that another upstream might support.
+			// For invalid request errors, there is a high chance that the error is due to a mistake that the user
+			// has done, and retrying another upstream would not help.
 			return common.NewErrEndpointClientSideException(
 				common.NewErrJsonRpcExceptionInternal(
 					int(code),
@@ -391,7 +403,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 					details,
 				),
-				true, // retryable
+				false, // not retryable
 			)
 		}
 
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -159,7 +159,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 					details,
 				),
-				true, // retryable
+				true, // retryable towards network
 			)
 		}
 
@@ -235,7 +235,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 					details,
 				),
-				false, // not retryable
+				false, // not retryable towards network
 			)
 		}
 
@@ -301,7 +301,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 						nil,
 						details,
 					),
-					true, // retryable
+					true, // retryable towards network
 				)
 			}
 		}
@@ -348,7 +348,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 					details,
 				),
-				true, // retryable
+				true, // retryable towards network
 			)
 		} else if code == -32600 {
 			if dt, ok := err.Data.(map[string]interface{}); ok {
@@ -363,7 +363,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 								nil,
 								details,
 							),
-							true, // retryable
+							true, // retryable towards network
 						)
 					}
 				}
@@ -386,7 +386,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 					details,
 				),
-				false, // not retryable
+				false, // not retryable towards network
 			)
 		}
 
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -169,9 +169,6 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 			strings.HasPrefix(msg, "Safe block not found") ||
 			strings.HasPrefix(msg, "finalized block not found") ||
 			strings.HasPrefix(msg, "Finalized block not found") {
-
-			details["blockTag"] = strings.ToLower(strings.SplitN(err.Message, " ", 2)[0])
-
 			// by default, we retry this type of clien-side exception as other upstreams might
 			// have/support this specific block tag data.
 			return common.NewErrEndpointClientSideException(
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -278,6 +278,20 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 				)
 			}
+			// Special case for envio's handling of "eth_getBlockReceipts" where they don't support object of
+			// type "BlockNumber" in the "blockHash" field.
+			if strings.Contains(msg, "invalid type: map, expected BlockNumber, 'latest', or 'earliest'") {
+				return common.NewErrEndpointClientSideException(
+					common.NewErrJsonRpcExceptionInternal(
+						int(code),
+						common.JsonRpcErrorInvalidArgument,
+						err.Message,
+						nil,
+						details,
+					),
+					true,
+				)
+			}
 			return common.NewErrEndpointClientSideException(
 				common.NewErrJsonRpcExceptionInternal(
 					int(code),
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -169,6 +169,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 			strings.HasPrefix(msg, "Safe block not found") ||
 			strings.HasPrefix(msg, "finalized block not found") ||
 			strings.HasPrefix(msg, "Finalized block not found") {
+
 			// by default, we retry this type of clien-side exception as other upstreams might
 			// have/support this specific block tag data.
 			return common.NewErrEndpointClientSideException(
@@ -350,7 +351,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 			if dt, ok := err.Data.(map[string]interface{}); ok {
 				if innerMsg, ok := dt["message"]; ok {
 					if strings.Contains(innerMsg.(string), "validation errors in batch") {
-						// Return a server-side error so the caller might retry or split the batch.
+						// Return a retryable client-side error so the caller might retry or split the batch.
 						return common.NewErrEndpointClientSideException(
 							common.NewErrJsonRpcExceptionInternal(
 								int(code),
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -56,7 +56,31 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 			msg += " Data: " + fmt.Sprintf("%v", err.Data)
 		}
 
-		// Infer from known status codes
+		//----------------------------------------------------------------
+		// 1) "Unsupported" errors
+		//----------------------------------------------------------------
+
+		if r.StatusCode == 415 || code == common.JsonRpcErrorUnsupportedException || // By HTTP status code or explicit JSON-RPC error code
+			code == -32004 || code == -32001 || // direct codes from upstream
+			strings.Contains(msg, "Unsupported method") ||
+			strings.Contains(msg, "not supported") ||
+			strings.Contains(msg, "method is not whitelisted") ||
+			strings.Contains(msg, "is not included in your current plan") {
+			return common.NewErrEndpointUnsupported(
+				common.NewErrJsonRpcExceptionInternal(
+					int(code),
+					common.JsonRpcErrorUnsupportedException,
+					err.Message,
+					nil,
+					details,
+				),
+			)
+		}
+
+		//----------------------------------------------------------------
+		// 2) "Request-too-large / range-too-large" errors
+		//----------------------------------------------------------------
+
 		if strings.Contains(msg, "Try with this block range") ||
 			strings.Contains(msg, "block range is too wide") ||
 			strings.Contains(msg, "this block range should work") ||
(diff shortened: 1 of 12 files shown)


common/errors.go
```diff
@@ -1958,6 +1968,23 @@ func HasErrorCode(err error, codes ...ErrorCode) bool {
...
+			// If “retriable: true” is present in the details, treat it as "retryable"
+			if details := e.Details(); details != nil {
+				if r, has := details["retriable"].(bool); has && r {
+					return true
+				}
+			}
+			// Otherwise, it’s not retriable
+			return false
+		}
+	}
```
aramalipoor, line 1986: Probably can be replaced by something like:
```go
if rt, ok := err.DeepSearch("retriable").(bool); ok && !rt {
  return false
}
```

Although the `DeepSearch` needs to be updated to return `interface{}` or `any` meaning any type and not only strings!

aramalipoor: I also suggest to put this lookup 'after' the HasErrorCode logic below, because that one is faster and more often hit vs this one (I'm guessing)

architecture/evm/error_normalizer.go
```diff
@@ -299,6 +299,20 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 					nil,
 				)
 			}
+			// Special case for envio's handling of "eth_getBlockReceipts" where they don't support object of
+			// type "BlockNumber" in the "blockHash" field.
+			if strings.Contains(msg, "invalid type: map, expected BlockNumber, 'latest', or 'earliest'") {
```
aramalipoor, line 304:
```suggestion
			if strings.Contains(msg, "invalid type:") {
```
How about this to make it more generic invalid arg err?

aramalipoor:
@kasrakhosravi the implementation is good imo (with suggested changes) we need to think about all client-side errors that we currently have to decide which ones must be retriable.

The current behavior it retries all client-side errors (check failsafe.go), for example Tenderly returns a 4xx on certain features they don't support which must be retried on other upstreams.

Probably it's good to go over all current instances of ClientSideException and decide one by one if they must be marked as retriable, and then to be safe the default ClientSideException (where we don't know exactly what is the reason) to set it to 'true'. The downside would be extra retries which is better than prematurely failing-fast without trying other upstreams if we really must.

architecture/evm/error_normalizer.go
kasrakhosravi: I also thought keeping them more specific is better. also not sure if all `invalid type` errors fall under the `retryable` category. needs to be investigated.

architecture/evm/error_normalizer.go
```diff
@@ -276,29 +324,29 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
...
-			}
+			strings.Contains(msg, "invalid params") ||
+			strings.Contains(msg, "tx of type") ||
+			strings.Contains(msg, "invalid type") {
+
+			// by default, we retry this type of client-side exception, as the root cause
+			// might be an invalid argument, that another upstream might support.
+			//
+			// Examples:
+			// - Envio not supporting { blockNumber: XXX, blockHash: YYY } for eth_getBlockReceipts
```
aramalipoor, line 349: For majority of the cases this is probably not correct. Validation errors or invalid params can happen a lot if user provides a bad HEX number for example. I suggest to treat these vendor-specific ones explicitly vs retrying all 4xx errors.

A counter-argument is we don't expect properly designed prod dApps to generate these kinds of errors as they'll have higher level validation checks before calling RPC. Although it's in theory, best is to check big RPC clients how many 4xx do they see.

What do you think?

architecture/evm/error_normalizer.go
```diff
@@ -307,14 +355,19 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
...
+		if code == -32600 {
 			if dt, ok := err.Data.(map[string]interface{}); ok {
-				if msg, ok := dt["message"]; ok {
-					if strings.Contains(msg.(string), "validation errors in batch") {
-						// Intentionally return a server-side error for failed requests in a batch
-						// so they are retried in a different batch.
-						// TODO Should we split a batch instead on json-rpc client level?
+				if innerMsg, ok := dt["message"]; ok {
+					if strings.Contains(innerMsg.(string), "validation errors in batch") {
+						// Return a server-side error so the caller might retry or split the batch.
```
aramalipoor, line 370: Since we have retryable client errors I think we can remove this edge-case? There's a unit test for that I expect that should pass

thirdparty/quicknode.go
```diff
@@ -60,17 +60,31 @@ func (v *QuicknodeVendor) GetVendorSpecificErrorIfAny(resp *http.Response, jrr i
...
 			)
 		} else if strings.Contains(msg, "failed to parse") {
+			// We do not retry on parse errors, as retrying another upstream would not help.
 			return common.NewErrEndpointClientSideException(
 				common.NewErrJsonRpcExceptionInternal(code, common.JsonRpcErrorParseException, msg, nil, details),
+				false, // not retryable
 			)
-		} else if code == -32010 {
+		} else if code == -32010 { // Transaction cost exceeds current gas limit
+			// We do not retry low gas limit errors, as retrying another upstream would not help.
```
aramalipoor, line 69: If i'm not wrong the gas limit is a node-specific config, if that's true then trying on another upstream might be actually a correct behavior

kasrakhosravi: This indeed seems to be from [Parity](https://github.com/openethereum/parity-ethereum/blob/55c90d4016505317034e3e98f699af07f5404b63/rpc/src/v1/helpers/errors.rs#L452)

aramalipoor: I mean the "max gas" is probably a node-specific config (geth, reth, etc), based on this code it seems to be: https://github.com/openethereum/parity-ethereum/blob/v2.7.2-stable/miner/src/pool/verifier.rs#L174

kasrakhosravi: seems like it. if I understood correctly, each client can configure their own `gas limit` and also always make sure it is always lower than the `block gas limit overall`.

So one client (e.g reth) might have configured a gas limit higher than parity (but still below block gas limit) and therefore retrying another upstream would make sense under this logic.

architecture/evm/error_normalizer.go
kasrakhosravi: totally agreed. I think the invalid arg, params, request is too broad to always retry. would be nice though to get as many of these invalid `arg -> still retryable` error messages as possible.

thirdparty/quicknode.go
```diff
@@ -60,17 +60,32 @@ func (v *QuicknodeVendor) GetVendorSpecificErrorIfAny(resp *http.Response, jrr i
...
+				return common.NewErrEndpointClientSideException(
+					common.NewErrJsonRpcExceptionInternal(code, common.JsonRpcErrorInvalidArgument, msg, nil, details),
+					false, // not retryable
+				)
+			} else if strings.Contains(msg, "eth_getLogs and eth_newFilter are limited") {
+				// retrying on limited range errors for eth_getLogs and eth_newFilter toward other upstreams would be helpful.
+				return common.NewErrEndpointClientSideException(
+					common.NewErrJsonRpcExceptionInternal(code, common.JsonRpcErrorInvalidArgument, msg, nil, details),
+					true, // retryable
+				)
```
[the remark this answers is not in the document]
kasrakhosravi: indeed. there was a mistake in quicknode's error docs: https://www.quicknode.com/docs/base/error-references

sending a large range `eth_getLogs` request to quicknode actually returns another error code:

```
{
    "jsonrpc": "2.0",
    "id": 1,
    "error": {
        "code": -32614,
        "message": "eth_getLogs is limited to a 10,000 range"
    }
}
```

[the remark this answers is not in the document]
aramalipoor: lol, alright let's include both messages then, and both to be TooLarge error

architecture/evm/error_normalizer.go
kasrakhosravi: I think the edge-case itself. cannot be removed cause it seems a specific one that carries a certain error in `data.message`. but I created a branch for it in our now organized `Invalid Request` category which would return a client-side exception which is retryable.

kasrakhosravi:
> We have a "ignoreClientErrors" flag in `failsafe.retry` policy, with this change we need to revisit if that flag is needed at all 🤔

don't think we need that flag anymore. and now our code is also cleaner and easier to reason about as we either explicitly define which client errors are retryable and which are not.

bin/erpc-server
aramalipoor, line 1: probably we should add these to .gitignore

common/errors.go
```diff
@@ -938,7 +935,10 @@ func (e *ErrUpstreamsExhausted) DeepestMessage() string {
 }
 
 func (e *ErrUpstreamsExhausted) UpstreamId() string {
-	return e.DeepSearch("upstreamId")
+	if val := e.DeepSearch("upstreamId"); val != nil {
+		return val.(string)
```
aramalipoor, line 939: To avoid unexpected panics maybe:
```suggestion
		if s, ok := val.(string); ok { return s }
```
?

architecture/evm/error_normalizer.go
```diff
@@ -56,7 +56,31 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
...
+			return common.NewErrEndpointUnsupported(
+				common.NewErrJsonRpcExceptionInternal(
+					int(code),
+					common.JsonRpcErrorUnsupportedException,
+					err.Message,
+					nil,
+					details,
+				),
+			)
+		}
```
aramalipoor, line 78: In recent commit had to move this to avoid premature detection when message is only "not found" e.g. from Tenderly, we had to move this in different location.

https://github.com/erpc/erpc/pull/204/commits/6df5524559940926102bf3f62ffed6c2e6dae2ce

kasrakhosravi: gotcha. I moved the `unsupported` error normalization logic below `not found` block and also put a comment for why we want to have it there.

common/errors.go
```diff
@@ -1524,12 +1526,22 @@ type ErrEndpointClientSideException struct{ BaseError }
 
 const ErrCodeEndpointClientSideException = "ErrEndpointClientSideException"
 
-var NewErrEndpointClientSideException = func(cause error) error {
+func NewErrEndpointClientSideException(cause error, retryableTowardNetwork ...bool) error {
```
aramalipoor, line 1529: How about making this an array of flags/options?

```suggestion
func NewErrEndpointClientSideException(cause error, opts ...ExceptionOption) error {
```

then usage will be: 
```golang
common.NewErrEndpointClientSideException(cause, RetryableTowardNetwork)
```

kasrakhosravi: ah gotcha. so more explicit and also if we want to have other flags such as `RetryableTowardUpstream`, we can have that as well.

kasrakhosravi: Exactly. with the logic that we have in `isRetryableTowardNetwork`, if we add it explicitly to any of the exceptions details param; we would explicitly set the retyable behavior.

common/errors.go
```diff
@@ -1945,8 +1967,39 @@ func HasErrorCode(err error, codes ...ErrorCode) bool {
...
+			errs := exher.Errors()
+			if len(errs) > 0 {
+				for _, e := range errs {
+					if IsRetryableTowardsUpstream(e) {
+						return true
+					}
+				}
+			}
+		}
+	}
```
aramalipoor, line 1995: Shouldn't this be inside IsRetryableTowardsUpstream ?

common/errors.go
```diff
@@ -1973,7 +2024,17 @@ func IsRetryableTowardsUpstream(err error) bool {
...
 		ErrCodeEndpointRequestTooLarge,
-	)
+	) {
+		return false
+	}
+
+	// If the upstream is hitting capacity limits -> no retry
+	if IsCapacityIssue(err) {
+		return false
+	}
```
aramalipoor, line 2034: Is this a breaking change compared to current behavior?

common/errors.go
kasrakhosravi: in the current logic, we are checking this against network scope

https://github.com/erpc/erpc/blob/671178233890f88c0d1918d2b209eeec2bba214b/upstream/failsafe.go#L323

common/errors.go
```diff
@@ -90,13 +90,23 @@ type StandardError interface {
...
 
+type ExceptionOption func(e StandardError)
+
+func WithRetryableTowardNetwork(retryable bool) ExceptionOption {
+	return func(e StandardError) {
+		if e.Base() != nil {
+			e.Base().Details["retryableTowardNetwork"] = retryable
+		}
+	}
+}
```
aramalipoor, line 108:
```suggestion
func (e BaseError) WithRetryableTowardNetwork(retryable bool) ExceptionOption {
// ..
}
```

If we make this a func of the BaseError then we can easily make any error instance retryable or not, either when initializing such errors or when used...

The usage will be simpler:
```golang
common.NewErrEndpointClientSideException(...).WithRetryableTowardNetwork(true)
```

common/errors.go
```diff
@@ -90,13 +90,23 @@ type StandardError interface {
...
 	Base() *BaseError
 	MarshalZerologObject(v *zerolog.Event)
 }
 
+type ExceptionOption func(e StandardError)
+
+func WithRetryableTowardNetwork(retryable bool) ExceptionOption {
+	return func(e StandardError) {
+		if e.Base() != nil {
+			e.Base().Details["retryableTowardNetwork"] = retryable
```
aramalipoor, line 105: On a different note, should we make sure Details exist before setting a value?

common/errors.go
kasrakhosravi: we have this referenced in two places in the current code and encapsulating it inside the `isRetryableTowardUpstream` would make our code cleaner

here since we are returning `not retryable` for `IsCapacityIssue `, should cover the case here

https://github.com/erpc/erpc/blob/671178233890f88c0d1918d2b209eeec2bba214b/upstream/failsafe.go#L312

and same logic here

https://github.com/erpc/erpc/blob/671178233890f88c0d1918d2b209eeec2bba214b/erpc/networks.go#L279

common/errors.go
```diff
@@ -1945,8 +1967,39 @@ func HasErrorCode(err error, codes ...ErrorCode) bool {
...
+					if IsRetryableTowardsUpstream(e) {
+						return true
+					}
+				}
+			}
+		}
+	}
+
+	// If the error is retryable towards upstream, then it is also retryable towards network
+	return IsRetryableTowardsUpstream(err)
```
aramalipoor, line 1998: If this is true, then does it mean we're doing a redundant `IsRetryableTowardsUpstream` in failsafe.go, for the flow of network scope? Is there any chance to simplify that?

common/errors.go
aramalipoor: Makes sense, I'm wondering why we didn't have it inside before. My guess is because we didn't have network-level retryable logic, am I correct?

common/errors.go
aramalipoor: That was probably a different reason (maybe cuz we didn't expect this code for upstream errors).

Logically it does seem like a IsRetryableTowardsUpstream case, what do you think?

common/errors.go
```diff
@@ -1524,12 +1541,15 @@ type ErrEndpointClientSideException struct{ BaseError }
...
 const ErrCodeEndpointClientSideException = "ErrEndpointClientSideException"
 
-var NewErrEndpointClientSideException = func(cause error) error {
+func NewErrEndpointClientSideException(cause error) RetryableError {
 	return &ErrEndpointClientSideException{
-		BaseError{
+		BaseError: BaseError{
 			Code:    ErrCodeEndpointClientSideException,
-			Message: "client-side error when sending request to remote endpoint",
+			Message: "client-side error, etc.",
```
aramalipoor, line 1548: This message indicates the error is initiated by the upstream, after request was sent to endpoint. The new message is too vague and wouldn't be useful for narrowing down.

In general not sure how to feel about error msgs, maybe we wouldn't really need them... could be a refactor for another day

kasrakhosravi: hmm makes sense. do you think error code is descriptive enough?

common/errors.go
aramalipoor: Ah nice find!

upstream/failsafe.go
```diff
@@ -293,47 +293,16 @@ func createRetryPolicy(scope common.Scope, cfg *common.RetryPolicyConfig) (fails
...
-						}
-						if !cfg.IgnoreClientErrors && common.IsClientError(err) {
-							return true
-						}
-					}
-				}
+			return common.IsRetryableTowardsUpstream(err)
+		} else if scope == common.ScopeNetwork && err != nil {
+			if rt := common.IsRetryableTowardNetwork(err); rt != nil {
+				return *rt
```
aramalipoor, line 305: I'm not sure if this approach gives us what we need, because in case result is `true` we don't wanna short-circuit and return, we actually want to continue in case for some other reason we should return false (e.g. MissingData is retryable, unless it's from an already synced upstream)

So probably instead of pointer-bool approach it should look like:
```suggestion
			if !common.IsRetryableTowardNetwork(err) {
				return false
```

Although, please go over the full logic to make sure what i say makes sense, because we might want the other way `if common.IsRetryableTowardNetwork(err) { return true }`

common/errors.go
aramalipoor: In general I think so, I don't remember any two errors that had same code but diff message, which means even a reference code of Code -> Explanation would be sufficient vs bloating the memory + code with long strings (Imagine on a 10k RPS instance thousands of copies of such long strings are created in memory). The overall mb might be negligible.

Anyway we can think about removing Message concept in a separate refactor if it made sense

thirdparty/quicknode.go
```diff
@@ -60,17 +61,21 @@ func (v *QuicknodeVendor) GetVendorSpecificErrorIfAny(resp *http.Response, jrr i
...
+			).WithRetryableTowardNetwork(false)
+		} else if code == -32010 { // Transaction cost exceeds current gas limit
+			// retrying on gas limit exceeded errors toward other upstreams would be helpful, as max gas limit
+			// can be defined per client (reth, geth, parity, etc.) (still needs to be lower than overall block gas limit)
 			return common.NewErrEndpointClientSideException(
 				common.NewErrJsonRpcExceptionInternal(code, common.JsonRpcErrorClientSideException, msg, nil, details),
 			)
-		} else if code == -32602 {
+		} else if code == -32602 && strings.Contains(msg, "cannot unmarshal hex string") {
+			// we do not retry on invalid argument errors, as retrying another upstream would not help.
```
aramalipoor, line 75: We probably should separate these two, an 'invalid argument' could be because Quicknode does not support certain param/tag... which we should retry on other nodes.

The 'cannot unmarshal hex string' is most probably safe to not retry.

kasrakhosravi: Am I correct that the `invalid argument (32602)` is a bit too general to be considered retryable?

https://github.com/erpc/erpc/pull/203/files#diff-0b005aa95e710cc341ee8cddc3fb690c956286ea19b90299a23e1c416470fceeR366

kasrakhosravi: And also the logic here `else if code == -32602 && strings.Contains(msg, "cannot unmarshal hex string")`

is specific enough to be checked here against quicknode (usage of `&`) and if it is just a `32602 `, it will be checked anyway in `error_normalizer`

architecture/evm/error_normalizer.go
```diff
@@ -257,66 +301,78 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
...
 					}
 				}
 			}
+		} else if code == -32602 || code == -32600 ||
+			strings.Contains(msg, "param is required") ||
+			strings.Contains(msg, "Invalid Request") ||
+			strings.Contains(msg, "validation errors") ||
+			strings.Contains(msg, "invalid argument") ||
+			strings.Contains(msg, "invalid params") ||
+			strings.Contains(msg, "tx of type") {
```
aramalipoor, line 372: This specific error is probably for tx type support, if yes then we need to retry it on network-level

aramalipoor: I see it's already covered above so actually let's remove it from here

architecture/evm/error_normalizer.go
```diff
@@ -257,66 +301,78 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
...
+		// for certain args/params for certain methods, among different providers or clients. We will retry them.
+		//
+		// Examples:
+		if strings.Contains(msg, "tx of type") || // Should be retried toward upstreams that support this tx type
+			strings.Contains(msg, "invalid type: map, expected BlockNumber, 'latest', or 'earliest'") { // Envio not supporting { blockNumber: XXX, blockHash: YYY } for eth_getBlockReceipts
+			return common.NewErrEndpointClientSideException(
 				common.NewErrJsonRpcExceptionInternal(
 					int(code),
-					common.JsonRpcErrorUnsupportedException,
+					common.JsonRpcErrorInvalidArgument,
```
aramalipoor, line 343: Still common.JsonRpcErrorUnsupportedException is more accurate is there a reason to make it InvalidArgument?

architecture/evm/error_normalizer.go
kasrakhosravi: good catch. fixed

architecture/evm/error_normalizer.go
kasrakhosravi: hmm I think for the first one (`tx of type`) the `unsupported` makes more sense while for the second one the `invalid argument`. maybe best to separate those here?

thirdparty/quicknode.go
aramalipoor: I think the code is too generic to not retry (i.e. it's better we retry such generic code) but specific texts about param required/unmarshal errors probably is safe to not retry.

In general when in doubt let's retry.

architecture/evm/error_normalizer.go
aramalipoor: Even for second is basically a tag that they don't support, so it should be fine imo

PR #212, opened 2025-03-28
fix: create client per chainId for dynamic providers

kasrakhosravi:
- [x] thirdweb
- [x] pimlico
- [x] envio

### DIFF
--- a/erpc/networks_registry.go
+++ b/erpc/networks_registry.go
@@ -137,7 +137,34 @@ func (nr *NetworksRegistry) GetNetwork(networkId string) (*Network, error) {
 		return nil, fmt.Errorf("network %s is not properly initialized yet", networkId)
 	}
 
-	return ntw.(*Network), nil
+	// Check if upstreams are available for this network
+	network := ntw.(*Network)
+	upstreams := nr.upstreamsRegistry.GetNetworkUpstreams(networkId)
+	if len(upstreams) == 0 {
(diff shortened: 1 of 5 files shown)


erpc/networks_registry.go
```diff
@@ -137,7 +137,34 @@ func (nr *NetworksRegistry) GetNetwork(networkId string) (*Network, error) {
...
+	if len(upstreams) == 0 {
+		// Instead of returning an error, wait a short time for upstreams to initialize
+		// This helps with race conditions where the network is created but upstreams aren't ready yet
+		ctx, cancel := context.WithTimeout(context.Background(), 1000*time.Millisecond)
+		defer cancel()
+
+		ticker := time.NewTicker(50 * time.Millisecond)
+		defer ticker.Stop()
+
+		for {
```
aramalipoor, line 152: I'm not sure if this is the best remedy for this, we need to investigate why such ordering/race condition happens and fix that if it's something like incorrect Initializer dependencies etc

aramalipoor:
Pimlico provider also needs the same treatment I believe

erpc/networks_registry.go
kasrakhosravi: yeah as we discussed this was not the main issue. by waiting for at least one upstream in `PrepareUpstreamsForNetwork`, we are correctly waiting for the initialziation of upstreams for network

PR #222, opened 2025-04-07
fix: tolerate block head rollback only to a certain degree

kasrakhosravi:


aramalipoor:
Probably the better approach for this is to introduce such max deviation tolerance in shared state variables (as we discussed to align when remotely stored values go out of sync).

The evm state poller then can stay what it is, do you see any reason we have to change the logic there?

Also cordoning should only happen from policy evaluations, the main purpose of of having such feature is to avoid hard-coding logic around cordoning/uncordoning. Would that be a problem for the goal of this PR?

data/shared_state_registry.go
```diff
@@ -14,7 +14,7 @@ import (
 )
 
 type SharedStateRegistry interface {
-	GetCounterInt64(key string) CounterInt64SharedVariable
+	GetCounterInt64(key string, maxAllowedDrift int64) CounterInt64SharedVariable
```
aramalipoor, line 17:
```suggestion
	GetCounterInt64(key string, toleratedDownDrift int64) CounterInt64SharedVariable
```

To be more precise (only for downwards decreases)?

architecture/evm/evm_state_poller.go
```diff
@@ -95,6 +96,14 @@ func NewEvmStatePoller(
 		e.tracker.SetFinalizedBlockNumber(e.upstream.Config().Id, e.upstream.NetworkId(), value)
 	})
 
+	// TODO: replace with tracker func
+	lbs.OnDrift(func(currentVal, newVal int64) {
+		e.tracker.RecordBlockHeadDownDrift(e.upstream.Config().Id, e.upstream.NetworkId(), "latest", currentVal, newVal)
+	})
+	fbs.OnDrift(func(currentVal, newVal int64) {
```
aramalipoor, line 103:
```suggestion
	fbs.OnLargeDownDrift(func(currentVal, newVal int64) {
```
?

architecture/evm/evm_state_poller.go
```diff
@@ -95,6 +96,14 @@ func NewEvmStatePoller(
...
+		e.tracker.RecordBlockHeadDownDrift(e.upstream.Config().Id, e.upstream.NetworkId(), "finalized", currentVal, newVal)
```
aramalipoor, line 104:
```suggestion
		e.tracker.RecordBlockHeadLargeDownDrift(e.upstream.Config().Id, e.upstream.NetworkId(), "finalized", currentVal, newVal)
```
?

data/shared_state_variable.go
```diff
@@ -217,6 +248,12 @@ func (c *counterInt64) TryUpdateIfStale(ctx context.Context, staleness time.Dura
 		currentValue := c.value.Load()
 		if newValue > currentValue {
 			c.setValue(newValue)
+		} else if currentValue > newValue && (currentValue-newValue > c.ignoreDownDriftOf) {
+			c.setValue(newValue)
+
+			if c.driftCallback != nil {
+				c.driftCallback(currentValue, newValue)
+			}
```
aramalipoor, line 256: Looks like the logic of "if it's higher considering the drift then set value and call callback" is repeated few times in this func

Is it possible to simplify, in one place (using some vars), so it's easier to reason about?

architecture/evm/evm_state_poller.go
```diff
@@ -75,8 +75,9 @@ func NewEvmStatePoller(
 ) *EvmStatePoller {
 	lg := logger.With().Str("upstreamId", up.Config().Id).Str("component", "evmStatePoller").Logger()
 
-	lbs := sharedState.GetCounterInt64(fmt.Sprintf("latestBlock/%s", common.UniqueUpstreamKey(up)))
-	fbs := sharedState.GetCounterInt64(fmt.Sprintf("finalizedBlock/%s", common.UniqueUpstreamKey(up)))
+	// TODO: replace with integrity config
+	lbs := sharedState.GetCounterInt64(fmt.Sprintf("latestBlock/%s", common.UniqueUpstreamKey(up)), 1024)
```
aramalipoor, line 79: For now maybe let's put this in a const?
```suggestion
	lbs := sharedState.GetCounterInt64(fmt.Sprintf("latestBlock/%s", common.UniqueUpstreamKey(up)), DefaultToleratedBlockHeadDownDrift)
```

Also for the value (1024), do we need to be more strict or it's better to leave it as is. This is mainly from the same upstream (which means maximum 'normal' down drift we should expect is the evm poller interval / fastest block production on fastest chain)

Using too low of value can create false-positives hence creating flip-flop values for latest block, and using too high of value can disregard small re-orgs (e.g. if a chain is reorged in past 8-16 blocks)

architecture/evm/evm_state_poller.go
```diff
@@ -99,6 +102,14 @@ func NewEvmStatePoller(
 		e.tracker.SetFinalizedBlockNumber(e.upstream.Config().Id, e.upstream.NetworkId(), value)
 	})
 
+	// TODO: replace with tracker func
```
aramalipoor, line 105: Is this TODO needed? I see the tracker func already

health/tracker.go
```diff
@@ -57,6 +57,7 @@ type TrackedMetrics struct {
 	RequestsTotal          atomic.Int64     `json:"requestsTotal"`
 	BlockHeadLag           atomic.Int64     `json:"blockHeadLag"`
 	FinalizationLag        atomic.Int64     `json:"finalizationLag"`
+	BlockHeadDownDrift     atomic.Int64     `json:"blockHeadDownDrift"`
```
aramalipoor, line 60:
```suggestion
	BlockHeadLargeDownDrifts     atomic.Int64     `json:"blockHeadLargeDownDrifts"`
```

telemetry/metrics.go
```diff
@@ -393,4 +393,10 @@ var (
 		Name:      "cors_disallowed_origin_total",
 		Help:      "Total number of CORS requests from disallowed origins.",
 	}, []string{"project", "origin"})
+
+	MetricBlockHeadDownDrift = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "erpc",
+		Name:      "block_head_down_drift",
+		Help:      "Number of blocks drifted down by the most up-to-date upstream.",
```
aramalipoor, line 400:
```suggestion
	MetricBlockHeadLargeDownDrift = promauto.NewGaugeVec(prometheus.GaugeOpts{
		Namespace: "erpc",
		Name:      "block_large_head_down_drift",
		Help:      "Number of times block head drifted downwards by a large number vs previous latest blocks returned by the same upstream.",
```

data/shared_state_variable.go
```diff
@@ -35,17 +36,36 @@ func (v *baseSharedVariable) IsStale(staleness time.Duration) bool {
...
-	value    atomic.Int64
-	mu       sync.Mutex // still needed for complex operations
-	callback func(int64)
+	registry          *sharedStateRegistry
+	key               string
+	value             atomic.Int64
+	mu                sync.Mutex // still needed for complex operations
+	valueCallback     func(int64)
+	ignoreDownDriftOf int64
+	driftCallback     func(localVal, newVal int64)
```
aramalipoor, line 45:
```suggestion
	largeDriftCallback     func(localVal, newVal int64)
```

architecture/evm/evm_state_poller.go
```diff
@@ -19,6 +19,9 @@ import (
 
 const FullySyncedThreshold = 4
 
+// TODO: replace with integrity config
```
aramalipoor, line 22:
```suggestion
// TODO: find a clean way to pass integrity config to evm pollers for lazy-loaded networks (not statically configured in erpc.yaml), at the moment an evm state poller might be initiated "before" a network is physically created and configured (e.g. when a new network is lazy-loaded from a Repository Provider)
```

aramalipoor:
Regarding wording, maybe we could use "rollback" as that's more appropriate in EVM chain context...
* `ignoreRollbackOf`
* `OnLargeRollback`
* `block_head_large_rollback`

Wdyt?

aramalipoor:
Great work 🚀

PR #232, opened 2025-04-18
bug: "finalized" block tag is cached incorrectly under 'finalized' policy vs the correct 'realtime' policy

kasrakhosravi:


architecture/evm/block_ref.go
```diff
@@ -345,9 +356,11 @@ func parseCompositeBlockParam(param interface{}) (string, int64, error) {
 				}
 				blockNumber = bni
 			}
-		} else {
-			// Block tag ('latest', 'earliest', 'pending')
+		} else if isBlockTag(v) {
+			// Named block tag (latest, pending, safe, finalized, earliest)
 			blockRef = v
+		} else {
+			return "", 0, errors.New("unsupported block reference: " + v)
```
aramalipoor, line 363: We better be less strict here as for chains or tags that might be added in future.

Probably the best way to solve this issue is simply changing here: https://github.com/erpc/erpc/blob/1016e7d7b4b8ee0be0db64058aaa3ab9a93dc664/architecture/evm/json_rpc_cache.go#L579 to Realtime vs Finalized

aramalipoor: Although needs more unit tests and actual tests 👌

kasrakhosravi: yeah makes sense. thought this is an improvement but it is a bit strict and less ideal to throw an error here.

architecture/evm/json_rpc_cache.go
```diff
@@ -576,7 +576,7 @@ func (c *EvmJsonRpcCache) getFinalityState(ctx context.Context, req *common.Norm
 	if blockRef == "latest" || blockRef == "pending" {
 		finality = common.DataFinalityStateUnfinalized
 	} else if blockRef == "finalized" {
```
kasrakhosravi, line 578: @aramalipoor also good to check for `safe` block tag here as well?

architecture/evm/block_ref.go
aramalipoor: Let's have these unit test scenarios at least (add others if you find more edge-cases):

* User asks for "finalized" tag it must be cached with "realtime" finality flag
* User asks for hex number of finalized block exactly/immediately after we have polled that finalized block, it must be cached with "finalized" finality flag
* WHEN data is cached under realtime finality flag and is not expired yet THEN user requests with the hex number => we must hit the cache ideally
* WHEN data is cached under 'finalized' flag (via hex number) i.e. not realtime flag, THEN user requests with 'finalized' tag => we must NOT hit the cache

architecture/evm/json_rpc_cache.go
aramalipoor: Good point! Actually for all non-hex tags we should consider "realtime" not finalized or unfinalized?  Because on every block these tags _might_ be different.

aramalipoor: Let's also update the docs accordingly. That if a block reference extracted from the req/resp is one of the tags the finality will be considered 'realtime'

kasrakhosravi: yeah that was the logic I had in mind and wanted to have a `IsBlockTag` helper to distinguish them, but checking with "is not empty and not 0x preferred" is also a good option

kasrakhosravi: probably if the user override the `finality` of a method, then we should respect that, but by default we should consider all block tags as real-time

aramalipoor: > if the user override the `finality` of a method

Can you elaborate which override? Do you mean the cache policy config? If yes, then that's more of a targetting vs overriding... i.e. user indicates 'realtime' data to be stored in redis for example, there's no way for them to override "latest block tag must be considered finalized/unfinalized" as that's hard-coded here in our code.

kasrakhosravi: I meant the config regarding customising cacheable methods

```
database:
  evmJsonRpcCache:
     ...
     methods:
      <method_name>:
         finalized: bool 
        realtime: bool 
 ```

Comment (kasrakhosravi) [reply architecture/evm/json_rpc_cache.go]: I think so too

File: architecture/evm/json_rpc_cache.go [hunk 3]
```diff
@@ -573,10 +573,15 @@ func (c *EvmJsonRpcCache) getFinalityState(ctx context.Context, req *common.Norm
 
 	blockRef, blockNumber, _ := ExtractBlockReferenceFromRequest(ctx, req)
 
-	if blockRef == "latest" || blockRef == "pending" {
+	// Decide finality from the block reference string:
+	//   • "*" wildcard entry: unfinalized (matches default policies)
+	//   • tag  (first char not a digit): realtime
+	if blockRef == "*" {
```
aramalipoor, line 579: Good catch! If we don't maybe let's add unit tests for hash-based eth_getBlockByNumber?

erpc/evm_json_rpc_cache_test.go
```diff
@@ -523,12 +522,11 @@ func TestEvmJsonRpcCache_Set(t *testing.T) {
...
-				method:         "eth_getTransactionByHash",
-				params:         `["0x123"]`,
-				result:         `"result":{"hash":"0x123","blockNumber":null}`,
-				expectedCache:  true,
-				expectedPolicy: unfinalizedPolicy, // Should match unfinalized policy with 30s TTL
+				name:          "Pending Transaction Request",
+				method:        "eth_getTransactionByHash",
+				params:        `["0x123"]`,
+				result:        `"result":{"hash":"0x123","blockNumber":null}`,
+				expectedCache: false,
```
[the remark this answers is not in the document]
kasrakhosravi: yeah, the assertion `AssertNotCalled(t, "Set")` without correct number of arguments was a loos assertion and was not correctly checking the `expectedCache`

as you said when we have a pending transaction, we would cache it under `unfinalized` state based on our logic which is the correct behavior.

architecture/evm/json_rpc_cache.go
kasrakhosravi: I think you meant `eth_getBlockByHash`?

Because `eth_getBlockByNumber` does not seem to accept hash of the block as param

```
{
    "params": [
        "0x3f07a9c83155594c000642e7d60e8a8a00038d03e9849171a05ed0e2d47acbb3",
        false
    ],
    "id": 1,
    "jsonrpc": "2.0"
}
```

aramalipoor: Having eth_getBlockByHash for sure is also good to have. I actually meant hex-based block by number (vs tag-based).

kasrakhosravi: oh yeah I think for that we have unit-tests

aramalipoor: Thinking about this again will this be a breaking-change/different behavior than current logic?

e.g. if ref = * but block num is of a very old block, it must be considered FInalized

kasrakhosravi: hmm yeah based on the logic we have `*` means I couldn't figure out the block. 

I think the reason we cannot rely on the `resp != nil` to handle that flow is because in `Get` cache, we are not relying on response, so we might be detecting the wrong policy?

https://github.com/erpc/erpc/blob/1016e7d7b4b8ee0be0db64058aaa3ab9a93dc664/architecture/evm/json_rpc_cache.go#L104

aramalipoor: Yeah that makes sense. Ultimate test would be cases for "cache transaction trace" or "cache eth_getTransactionByHash" where there's no block number in req nor resp.

kasrakhosravi: am I correct that for both `trace_transaction` and `eth_getTransactionByHash`, we actually return `blockNumber` as part of response?

https://www.quicknode.com/docs/ethereum/trace_transaction
https://www.quicknode.com/docs/ethereum/eth_getTransactionByHash

aramalipoor: Ah you're right was thinking about "trace_replayTransaction" trace func where no blockNumber exists.

Re eth_getTransactionByHash it's an interesting case, we can return '*' (as in it's not necessary to invalidate on reorg) but since we'd have a blockNumber probably better to have that as blockRef.

kasrakhosravi: so probably safe to remove `ReqRefs: ArbitraryBlock` for both `eth_getTransactionReceipt` and `eth_getTransactionByHash` as both return `blockNumber` as part of their response?

https://github.com/erpc/erpc/blob/1016e7d7b4b8ee0be0db64058aaa3ab9a93dc664/common/defaults.go#L361

aramalipoor: I believe that's needed for cache hits (i.e. the * means to use the secondary index where blockNum is not needed to 'look up' the cache entry).

PR #236, opened 2025-04-23
fix: thundering herd on counter shared state variables value func calls

kasrakhosravi:


architecture/evm/evm_state_poller.go
```diff
@@ -27,6 +27,31 @@ const DefaultToleratedBlockHeadRollback = 1024
...
+	once   sync.Once
+}
+
+func newPollingMultiplexer() *pollingMultiplexer {
+	return &pollingMultiplexer{
+		done: make(chan struct{}),
+	}
+}
+
+func (m *pollingMultiplexer) Close(result interface{}, err error) {
```
kasrakhosravi, line 45: @aramalipoor I thought about re-using the multiplexer logic [here](https://github.com/erpc/erpc/blob/main/erpc/multiplexer.go), but the logic around multiplexing polls seemed different than network requests.

do you think having it as part of `state poller` makes sense? or should we think about having it more generic?

aramalipoor: They seem similar, only that "result" type is different. You could try a design with generics see if it's clean.

Having a generic Multiplexer (probably in `util` pkg) is good re testability and DRY.

aramalipoor: Even some of the usage parts could be abstracted away (`executeMultiplexedPoll` in here, or `handleMultiplexing` in Network)...

If it's not straightfoward, for now only having a generic-based Multiplexer object could be enough. Later we can refactor the usage in a separate PR.

common/multiplexer.go
```diff
@@ -0,0 +1,63 @@
...
+// It should only be called after Done() channel is closed.
+func (m *Multiplexer[T]) Result() (T, error) {
+	m.mu.RLock()
+	defer m.mu.RUnlock()
+	return m.result, m.err
+}
+
+// Key returns the key associated with this multiplexer.
+func (m *Multiplexer[T]) Key() string {
+	return m.key
```
aramalipoor, line 62: Any reason for not having `key` field just to be public (vs a getter func)?

docs/pages/operation/monitoring.mdx
```diff
@@ -167,4 +168,7 @@ avg(erpc_upstream_score_overall) by (network, upstream)
 
 # CORS issues by origin
 sum(rate(erpc_cors_disallowed_origin_total{}[5m])) by (project, origin)
-```
+
+# Multiplexed poll operations by finalized poll operation
+sum(rate(erpc_upstream_multiplexed_polls_total{operation="finalized"}[5m])) by (network, upstream)
```
aramalipoor, line 173: Not sure if this metric is valuable from user's perspective? It creates bunch of cardinality but probably wouldn't help investigating issues? As a erpc core dev though it'll be useful during development.

common/multiplexer.go
kasrakhosravi: hmm I thought the key here was meant to be immutable (only being set in the constructor?)

docs/pages/operation/monitoring.mdx
kasrakhosravi: hmm interesting. would the same apply to `erpc_network_multiplexed_request_total ` as well?

aramalipoor: That one is actually useful as the requests are done by the users and they would get benefit from knowing how many were multiplexed. This internal polling call should be as few as possible from user's PoV anyway, anything more than necessary would be our code logic flaw. In perfect world we don't even need this multiplexer because our locking mechanisms wouldn't even cause too many redundant polls.

common/multiplexer.go
aramalipoor: Yeah especially as an immutable no need to go via a func call vs direct var access?

data/postgresql.go
```diff
@@ -383,7 +383,7 @@ func (p *PostgreSQLConnector) Lock(ctx context.Context, key string, ttl time.Dur
 
 	if !acquired {
 		_ = tx.Rollback(ctx)
-		err := fmt.Errorf("failed to acquire lock: already locked")
+		err := common.NewErrLockAlreadyHeld(fmt.Errorf("advisory lock already held"))
```
aramalipoor, line 386: `advisory`?

PR #251, opened 2025-04-28
feat: support full URI for Redis connector

### LINKED ISSUE #248: bug: RedisConnectorConfig does not support username option
**Describe the bug**
RedisConnectorConfig has no way to set `username` to utilise a non-default user.

**To Reproduce**
N/A

**Expected behavior**
It's possible to configure a username.

**Screenshots**

**Relevant Logs:**
w/ the invalid configuration including `username`
```jsonl
{"level":"error","error":"failed to load configuration from ./erpc.yaml: yaml: unmarshal errors:\n  line 13: field username not found in type common.RedisConnectorConfig\n  line 24: field username not found in type common.RedisConnectorConfig","time":1745839823194,"message":"failed to load configuration"}
{"level":"error","time":1745839823194,"message":"failed to start erpc: failed to load configuration from ./erpc.yaml: yaml: unmarshal errors:\n  line 13: field username not found in type common.RedisConnectorConfig\n  line 24: field username not found in type common.RedisConnectorConfig"}
```

w/o specifying any `username`.
```jsonl
{"level":"warn","component":"sharedState","connector":"redis","task":"redis-connect/redis","error":"failed to connect to Redis: WRONGPASS invalid username-password pair or user is disabled.","time":1745839631992,"message":"initialization task failed"}

```

**Additional context**
Attempting to create a non-default redis user, with locked down access to certain keys and commands - we configured a user `restricted` on the target Redis instance.
After creation, attempted to configure the connector such as (not based on documentation, but a guess of what the config would look like:
```
      connectors:
        - id: short-term
          driver: redis
          redis:
            addr: <REDACTED>:6379
            username: restricted
            password: <REDACTED>
            connPoolSize: 10
            tls:
              enabled: true
```

Until switching to the restricted user, it was running perfectly fine with the `default` user - which is the default for the configured redis client.
https://github.com/erpc/erpc/blob/main/data/redis.go#L83

kasrakhosravi:
fixes #248

### DIFF
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -641,6 +641,16 @@ func (m *MemoryConnectorConfig) SetDefaults() error {
 }
 
 func (r *RedisConnectorConfig) SetDefaults() error {
+	if r.URI != "" && (r.Addr != "" ||
+		r.Username != "" ||
+		r.Password != "" ||
+		r.DB != 0) {
+
+		return fmt.Errorf(
+			"redis connector: provide either 'uri' or discrete fields ('addr', 'username', 'password', 'db'), not both",
+		)
+	}
+
 	// Set default values for timeouts and connection pool size
 	if r.ConnPoolSize == 0 {
 		r.ConnPoolSize = 8
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -641,13 +641,9 @@ func (m *MemoryConnectorConfig) SetDefaults() error {
 }
 
 func (r *RedisConnectorConfig) SetDefaults() error {
-	if r.URI != "" && (r.Addr != "" ||
-		r.Username != "" ||
-		r.Password != "" ||
-		r.DB != 0) {
-
+	if r.URI != "" && r.Addr != "" {
 		return fmt.Errorf(
-			"redis connector: provide either 'uri' or discrete fields ('addr', 'username', 'password', 'db'), not both",
+			"redis connector: provide either 'uri' or 'addr/username/password/db', not both",
 		)
 	}
 
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -560,10 +560,10 @@ func (c *SharedStateConfig) SetDefaults(defClusterKey string) error {
 		return err
 	}
 	if c.FallbackTimeout == 0 {
-		c.FallbackTimeout = Duration(500 * time.Millisecond)
+		c.FallbackTimeout = Duration(10 * time.Second)
 	}
 	if c.LockTtl == 0 {
-		c.LockTtl = Duration(10 * time.Second)
+		c.LockTtl = Duration(30 * time.Second)
 	}
 	return nil
 }
@@ -647,6 +647,11 @@ func (r *RedisConnectorConfig) SetDefaults() error {
 		)
 	}
 
+	// URI is provided directly
+	if r.URI != "" {
+		return nil
+	}
+
 	// Set default values for timeouts and connection pool size
 	if r.ConnPoolSize == 0 {
 		r.ConnPoolSize = 8
@@ -659,11 +664,6 @@ func (r *RedisConnectorConfig) SetDefaults() error {
 	}
 	if r.SetTimeout == 0 {
 		r.SetTimeout = Duration(2 * time.Second)
-	}
-
-	// URI is provided directly
-	if r.URI != "" {
-		return nil
 	}
 
 	// URI needs to be constructed from individual fields
@@ -701,6 +701,11 @@ func (r *RedisConnectorConfig) SetDefaults() error {
 		u.Path = "/" + strconv.Itoa(r.DB)
 
 		r.URI = u.String()
+
+		r.Addr = ""
+		r.Username = ""
+		r.Password = ""
+		r.DB = 0
 	}
 
 	return nil
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -2,8 +2,10 @@ package common
 
 import (
 	"fmt"
+	"net"
 	"net/url"
 	"slices"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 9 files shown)


data/redis.go
```diff
@@ -80,14 +80,35 @@ func (r *RedisConnector) Id() string {
...
+	if r.cfg.URI != "" {
+		r.logger.Debug().Str("uri", util.RedactEndpoint(r.cfg.URI)).Msg("attempting to connect to Redis using URI")
+		options, err = redis.ParseURL(r.cfg.URI)
+		if err != nil {
+			return fmt.Errorf("failed to parse Redis URI: %w", err)
+		}
+
+		options.DialTimeout = r.initTimeout
+		options.ReadTimeout = r.getTimeout
+		options.WriteTimeout = r.setTimeout
```
paymog, line 96: do you need to set all this or can you rely on the undelrying redis lib to set all this up using the URI?

docs/pages/config/database/drivers.mdx
```diff
@@ -141,11 +167,24 @@ export default createConfig({
...
+- When using URI, these parameters are included in the URI string:
+  - `redis://[username]:[password]@[host][:port][/database]`
+  - Database is optional and defaults to 0
+
+- These parameters must be configured separately, even when using URI:
+  - Connection pool size
+  - TLS configuration
+  - Timeout settings (initTimeout, getTimeout, setTimeout)
+
+- If both URI and individual parameters are provided, the URI takes precedence for connection details, but explicit timeouts and connection pool size from the config will still be used.
```
paymog, line 181: imo if a URI is provided, the other config should be ignored since everything should be configurable directly in the URI

data/redis.go
kasrakhosravi: seems like `redis-go` already has default options for all three [here](https://pkg.go.dev/github.com/redis/go-redis/v9#Options) when parsing the URI; but I don't think it is something you can setup as part of the URI?

we also have default values for them in erpc [here](https://github.com/erpc/erpc/blob/main/common/defaults.go#L655)

kasrakhosravi: I need to test if this is possible or not, but would this be more readable?

`redis://user:secret@cache.example.com:6379/0?pool_size=20&dial_timeout=4s&read_timeout=2s&write_timeout=2s`

common/validation.go
```diff
@@ -404,17 +404,27 @@ func (p *PostgreSQLConnectorConfig) Validate() error {
...
+
+	if c.InitTimeout == 0 {
 		return fmt.Errorf("database.*.connector.redis.initTimeout is required")
 	}
-	if p.GetTimeout == 0 {
+	if c.GetTimeout == 0 {
 		return fmt.Errorf("database.*.connector.redis.getTimeout is required")
 	}
-	if p.SetTimeout == 0 {
+	if c.SetTimeout == 0 {
```
aramalipoor, line 427: If timeouts are defined in URI then these are allowed to be 0 here.

Perhaps if URI is provided to not validate separate fields?

data/redis.go
aramalipoor: Now with the URI support probably you only want to do these only if `r.setTimeout > 0` and actually set...

docs/pages/config/database/drivers.mdx
```diff
@@ -76,18 +76,14 @@ database:
...
         driver: redis
         redis:
-          addr: YOUR_REDIS_ADDRESS_HERE
-          password: YOUR_REDIS_PASSWORD_HERE
-          db: XXX
-          connPoolSize: 8
+          # Connection URI (Required)
+          # Format: redis://[username]:[password]@[host][:port][/database][?dial_timeout=value1&read_timeout=value2&write_timeout=value3&pool_size=value4]
+          uri: "redis://username:password@host:port/db?dial_timeout=5s&read_timeout=1s&write_timeout=2s&pool_size=10"
           tls:
```
aramalipoor, line 82: How is TLS handled? Will `rediss://` take care of it? What would be effect of having both configured?

data/redis.go
```diff
@@ -80,32 +80,107 @@ func (r *RedisConnector) Id() string {
...
+
+		if options.TLSConfig == nil {
+			// URI was redis:// — take YAML config wholesale.
+			options.TLSConfig = tlsConfig
+			r.logger.Debug().Msg("enabled TLS via YAML configuration")
+		} else {
+			// URI was rediss:// — merge YAML extras onto the baseline.
+			if len(tlsConfig.Certificates) > 0 {
+				options.TLSConfig.Certificates = tlsConfig.Certificates
+				options.TLSConfig.InsecureSkipVerify = false
```
aramalipoor, line 152: Should we leave that to whatever was parsed from the URI? or maybe let the configured value to override?

In situations where people use self-issued certs for internal uses

kasrakhosravi: do you mean the `InsecureSkipVerify`?

data/redis.go
```diff
@@ -80,32 +80,107 @@ func (r *RedisConnector) Id() string {
...
+	if redisURI == "" {
+		// Construct from discrete fields (addr, username, password, db)
+		if strings.TrimSpace(r.cfg.Addr) == "" {
+			return fmt.Errorf("missing Redis connection information: either 'uri' or 'addr' must be supplied")
+		}
+
+		var userInfo string
+		if r.cfg.Username != "" || r.cfg.Password != "" {
+			userInfo = r.cfg.Username + ":" + r.cfg.Password + "@"
+		}
```
aramalipoor, line 96: Will this work for current users with:
```yaml
redis:
  addr: some-redis:6379
  password: some-secret
```

Without username? i.e. what was the default username already set when users don't provide to Redis.options? Probably we should use that in such case to not break current users

data/redis.go
aramalipoor: I see, so we're assuming if certificates are provided user for sure doesn't intent to skip verification. I think it makes sense.

docs/pages/config/database/drivers.mdx
```diff
@@ -141,7 +133,45 @@ export default createConfig({
...
-Example of Redis config with eviction policy:
+
+#### TLS options
+
+When your Redis endpoint already uses **rediss://** (for example, Railway or Upstash), TLS is negotiated automatically and you can omit the entire `tls:` block.  
+Add the `tls:` section only when you need **mutual‑TLS** (client certificate/key) or your server uses a **private CA**:
+
+```yaml
+redis:
+  uri: redis://user:pass@redis.internal:6380/0
```
aramalipoor, line 144:
```suggestion
  uri: rediss://user:pass@redis.internal:6380/0
```
Shouldn't this be `rediss://` ?

docs/pages/config/database/drivers.mdx
```diff
@@ -141,7 +133,45 @@ export default createConfig({
...
+redis://[[username]:[password]@][host][:port][/database][?dial_timeout=value1&read_timeout=value2&write_timeout=value3&pool_size=value4]
+```
+
+Examples:
+- `redis://localhost:6379/0` - Basic connection to localhost
+- `redis://user:pass@redis.example.com:6379/0` - Connection with authentication
+- `redis://:password@redis.example.com:6379/0` - Connection with password only (no username)
+- `redis://redis.example.com:6379` - Connection without database number (defaults to 0)
+- `redis://redis.example.com:6379/0?dial_timeout=5s&read_timeout=1s&write_timeout=2s&pool_size=10` - Connection with timeouts and pool size
+
```
aramalipoor, line 165: As a user I'd appreciate if a docs page has a real example like:
```suggestion
- `redis://global-shared-states-redis-master.redis.svc.cluster.local:6379` - Kubernetes service deployed via [Bitnami charts](https://github.com/bitnami/charts/tree/master/bitnami/redis)
```

data/redis.go
kasrakhosravi: yes, it will. we cover three scenarios:

* if no username is passed, the `default` username will be applied by default
* if no password is passed, there is a case that the password is set as empty ""
* if username is passed but not password, that is wrong and we cover it in the validation

common/defaults.go
```diff
@@ -662,6 +653,33 @@ func (r *RedisConnectorConfig) SetDefaults() error {
...
+	// Clean up Addr if it has scheme prefixes
+	r.Addr = strings.TrimPrefix(r.Addr, "rediss://")
+	r.Addr = strings.TrimPrefix(r.Addr, "redis://")
+
+	// Construct URI from discrete fields
+	if r.Addr != "" {
+		var userInfo string
+		if r.Username != "" || r.Password != "" {
+			userInfo = r.Username + ":" + r.Password + "@"
+		}
```
aramalipoor, line 672:
```suggestion
		if r.Password != "" {
			userInfo = r.Username + ":" + r.Password + "@"
		}
```
Probably this is what we need, for cases where username is not provided and looks like this URI is valid: `redis://:mypassword@host`

PR #260, opened 2025-05-05
fix: not treat "ignore methods" as ErrUpstreamExhausted

kasrakhosravi:


aramalipoor:
Closing for now as I'm not sure about logic or perf impact yet and the log is low prio

PR #263, opened 2025-05-06
feat: add support for superchain provider

kasrakhosravi:


thirdparty/superchain.go
```diff
@@ -0,0 +1,341 @@
...
+	ChainID              int64    `json:"chainId"`
+	RPC                  []string `json:"rpc"`
+	Explorers            []string `json:"explorers"`
+	SuperchainLevel      int      `json:"superchainLevel"`
+	GovernedByOptimism   bool     `json:"governedByOptimism"`
+	DataAvailabilityType string   `json:"dataAvailabilityType"`
+	Parent               struct {
+		Type  string `json:"type"`
+		Chain string `json:"chain"`
+	} `json:"parent"`
```
aramalipoor, line 112: If we're not using other fields no need to parse the, if we only use "rpc" probably let's just use map[string]?

PR #264, opened 2025-05-06
feat: add tenderly vendor

zekiblue:
- Adds tenderly vendor

### DIFF
--- a/docs/pages/config/projects/providers.mdx
+++ b/docs/pages/config/projects/providers.mdx
@@ -19,6 +19,7 @@ Providers make it easy to add well-known third-parties RPC endpoints quickly. He
 - [`etherspot`](#etherspot) Accepts etherspot.io rpc endpoint for account-abstraction (ERC-4337) support.
 - [`dwellir`](#dwellir) Accepts dwellir.com api key and automatically adds all their EVM chains.
 - [`conduit`](#conduit) Accepts conduit.xyz api key and automatically adds all their EVM chains.
+- [`tenderly`](#tenderly) Accepts tenderly.co api key and automatically adds all their EVM chains.
 
 <Callout type='info'>
   eRPC supports **any EVM-compatible** JSON-RPC endpoint when using [`evm` type](/config/projects/upstreams). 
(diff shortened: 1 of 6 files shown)


erpc.dist.yaml
```diff
@@ -93,6 +93,20 @@ projects:
             backoffMaxDelay: 10s
             backoffFactor: 0.3
             jitter: 500ms
+      - id: tenderly-chain-example-1
+        endpoint: https://mainnet.tenderly.co/xxxxxxxxxxxxxxxxxxxxxx
```
aramalipoor, line 97:
```suggestion
      - id: tenderly-example-1
        endpoint: tenderly://YOUR_TENDERLY_API_KEY
```

We could suggest the provider in this example

erpc.dist.yaml
```diff
@@ -93,6 +93,20 @@ projects:
...
+        rateLimitBudget: global
```
aramalipoor, line 98:
```suggestion
        rateLimitBudget: global-tenderly
```

So that user can define a rate limit based on their Tenderly plan. We'd need to add the rate limit budget at the bottom of this yaml as well.

thirdparty/tenderly.go
```diff
@@ -0,0 +1,138 @@
...
+	// Other errors can be properly handled by generic error handling
+	return nil
+}
+
+func (v *TenderlyVendor) OwnsUpstream(ups *common.UpstreamConfig) bool {
+	if strings.HasPrefix(ups.Endpoint, "tenderly://") || strings.HasPrefix(ups.Endpoint, "evm+tenderly://") {
+		return true
+	}
+
+	return strings.Contains(ups.Endpoint, ".gateway.tenderly.co")
```
aramalipoor, line 137:
```suggestion
	return strings.Contains(ups.Endpoint, ".tenderly.co")
```
Not sure if that's the only domain supported, maybe safer to include the top-level, wdyt?

aramalipoor:
Thanks @zekiblue for this nice work 💯  I'll personally use it as soon as it's merged :D

thirdparty/tenderly.go
zekiblue: for the nodes tenderly always uses gateway.tenderly.co. So this should be good.

zekiblue:
@aramalipoor @kasrakhosravi what else needs to happen in this pr?

aramalipoor (approved):

aramalipoor:
@zekiblue everything looks great. Just needed the conflict resolution which you already done. Merging soon 🙌

aramalipoor:
I think there's a small syntax typo in defaults.go that tests are failing

zekiblue:
> I think there's a small syntax typo in defaults.go that tests are failing

It should be resolved now

docs/pages/config/projects/providers.mdx
```diff
@@ -472,45 +473,6 @@ export default createConfig({
 </Tabs.Tab>
 </Tabs>
 
-#### `superchain`
```
aramalipoor, line 475: @zekiblue probably accidentally removed

docs/pages/config/projects/providers.mdx
```diff
@@ -654,8 +616,4 @@ providers:
       apiKey: xxxxx
       networksUrl: https://api.conduit.xyz/public/network/all # (optional) Endpoint to fetch all supported networks
       recheckInterval: 24h # (optional) How often to recheck the API for newly added networks (default: 24h)
-  - vendor: superchain
```
aramalipoor, line 657: @zekiblue also here probably accidentally removed

docs/pages/config/projects/providers.mdx
zekiblue: ops, yes reverting these back

PR #265, opened 2025-05-06
feat: implement consensus failsafe policy

wparr-circle:
Follows up on the earlier creation of the skeleton configuration for the consensus failsafe policy in https://github.com/erpc/erpc/commit/0ff1daad636c230e9755068e3ae5936adada45f4

This introduces the required changes to implement the consensus behaviour, where requests are forwarded to multiple upstreams via the round-robin-esque functionality in erpc/networks.go which assigns a unique upstream per execution. For each parent execution, we create n child executions determined by requiredParticipants.
Upon completion of these executions we can then achieve consensus if the number of matching response hashes exceed agreementThreshold, otherwise causing either a lowParticipantsBehavior where there were fewer responses (ie. if an execution attempted to use the same upstream as previously observed) or disputeBehavior if all upstreams responded but we never reached the threshold.

punishMisbehavior is partially implemented, to determine sit out period for a misbehaving node - but likely requires further thought looping in disputeThreshold and actually persisting something to shared state.

Likewise failureBehavior is implemented, although not clear on if failure as implemented here where any upstream responds an error should be counted towards failure as originally planned - or if there is another use case. Considering that much of the failure handling is already owned by lowParticipantsBehavior and disputeBehavior respectively.


consensus/executor.go
```diff
@@ -14,66 +21,445 @@ type executor[R any] struct {
...
+			return string(jr.Result)
+		}
+	}
+	return fmt.Sprintf("%+v", result)
+}
+
+// hashResult creates an MD5 hash of a string
+func hashResult(s string) string {
+	return fmt.Sprintf("%x", md5.Sum([]byte(s))) //#nosec G401 -- Permit usage of md5 for fast hashing, not used for security
+}
```
aramalipoor, line 49: We have a [req.CacheHash()](https://github.com/erpc/erpc/blob/7dff3a3bff1001fd0df0489a727cc3e8c2ad9e24/common/request.go#L424) which we can utilize here, which is suppose to create similar cache values for the identical requests (identical as in ignoring cases for hashes for example, or ignoring jsonrpc.id etc).

There might be some improvements needed for it to be adapted for consensus tracking:
- The current implementation is for "Request" method and params mainly, we could clone for Response
- When comparing the current logic orders the "keys" in embedded objects to ensure no matter the structure of json or key ordering sent by the user it ends up with a deterministic hash. Should be OK for response as well but having unit tests for complex responses is useful (maybe one for getLogs array-based response, and one for a nested trace response).
- When hashing requests memory usage is not a big deal as usually "params" is small for majority of json-rpc  requests, but for responses we might have much larger objects, therefore using hash might be less memory-efficient than a deep comparison utility func. Probably if the performance efficiency is like 100x and memory usage not too large (e.g. 10% of request size at peak) we can keep using hashing for comparison. A golang benchmark comparing both approaches can be deterministic in actual memory usage + ops/sec.

aramalipoor:
In terms of tests it'll be great to have few integration tests in [http_server_test.go](https://github.com/erpc/erpc/blob/main/erpc/http_server_test.go) which basically does a closed loop test for http server -> network -> upstream -> failsafe path, it doesn't need to be exhaustive but few happy/unhappy tests can give confidence that this policy actually behaves as expected (e.g. nothing is broken in integration point of Consensus Policy + Forward funcs)

consensus/executor.go
```diff
@@ -14,66 +21,445 @@ type executor[R any] struct {
...
+
+		// Create a buffered channel for the single result
+		resultChan := make(chan *failsafeCommon.PolicyResult[R], 1)
+
+		// Track responses for consensus checking
+		responses := make([]*execResult[R], 0, e.requiredParticipants)
+		responseMu := sync.Mutex{}
+
+		// Track used upstreams to ensure we don't use the same one twice
+		usedUpstreams := make(map[string]struct{})
```
aramalipoor, line 66: In terms of memory usage (think +10k RPS load) probably it'll be less to use struct as key:
```suggestion
		usedUpstreams := make(map[common.Upstream]struct{})
```
And something like:
```
// mark an upstream used
usedUpstreams[up] = struct{}{}

// check it
if _, seen := usedUpstreams[up]; seen {
    // we’ve already picked this one
}
```

consensus/executor.go
```diff
@@ -14,66 +21,445 @@ type executor[R any] struct {
...
+				// Get the upstream from the response
+				var upstream common.Upstream
+				result := innerFn(consensusExec)
+				if result == nil {
+					return
+				}
+
+				// Get upstream from response if available
+				if resp, ok := any(result.Result).(*common.NormalizedResponse); ok {
+					if ups := resp.Upstream(); ups != nil {
```
aramalipoor, line 86: For this logic we need to come up with a more deterministic strategy since `innerFn` can end up choosing any upstream randomly and if it's the same upstream it wastes 1 request that we'll ignore the response of. On high traffic this can increase redundant traffic accidentally.

One option (which feels a bit hacky) is to use Directives ([similar to this line](https://github.com/erpc/erpc/blob/main/architecture/evm/eth_getLogs.go#L556)) to say to ignore current upstreams. I feel it's better to think of a better alternative 🤔 

Also there are various edge-cases to consider:
1. If total number upstreams defined is lower than `requiredParticipants` which _could_ be user mistake (or some sort of issue for lazy-loaded upstreams that failed to initialize e.g. due to temp network issues), where should this be reported (during request's lifecycle, during config startup), what should happen the individual request?
2. If total number of selected upstreams (via [selection policy](https://docs.erpc.cloud/config/projects/selection-policies)) is less than `requiredParticipants`, what would be expected/reasonable behavior?

For these could say the behavior is what is defined for `lowParticipantsBehavior`

consensus/executor.go
aramalipoor: Reading the code around "mostCommonResultCount" I see re (3) probably the hash calc approach will have good advantages!

consensus/executor.go
wparr-circle: From testing I'm not sure if I would call the behaviour non-deterministic - as in the case that there are multiple upstreams available, the [loopCtx section of networks.go](https://github.com/erpc/erpc/blob/main/erpc/networks.go#L313) ensures that upstreams are selected in order.

Certainly, and I can't fingerprint how this is done - but if there is only a single upstream available it will repeat requests. I'm not happy with this behaviour, unfortunately it proved impossible to get the upstream via context (ie. you cannot alter context of the parent execution) etc. as it's only available on the response.

The behaviour can actually be seen in one of the tests here where the upstream is called 3 times, when realistically it should probably have failed fast as opposed to sending spurious requests.
See: https://github.com/erpc/erpc/pull/265/files#diff-673990bc24f791e8e8226bc56877d5df609d3ecd3f431446abf6c6715b862947R101

And to your point the expected behaviour in response to that scenario is what is defined in lowParticipantsBehavior - of course with the unfortunate behaviour it currently has at the minute where it actually tries an upstream 3 times even though it isn't going to matter.

Changing that to correct it such as we only send 1 request per upstream does sound ideal.

Maybe something similar to this section, where we could maintain a sync.Map and _if_ the tryForward is associated with the consensus policy skip the upstream if it was seen before for an execution of tryForward for a consensus policy.
I don't know if it's possible to actually determine that predicate however, since there was no way to mark custom attributes on the execution and we're locked in with the likes of attempts, hedges etc.
```
				if _, respondedEmptyBefore := emptyResponses.Load(u); respondedEmptyBefore {
					loopSpan.SetAttributes(
						attribute.Bool("skipped", true),
						attribute.String("skipped_reason", "upstream already responded empty"),
					)
					ulg.Debug().Msgf("upstream already responded empty no reason to retry, skipping")
					req.Unlock()
					loopSpan.End()
					continue
				}
```

consensus/executor.go
wparr-circle: Added a small change to key by the memory address of the upstream as opposed to the string id.

consensus/executor.go
wparr-circle: I've finished up taking care of this in https://github.com/erpc/erpc/pull/265/commits/8ae188358eb9f77825af83ace3238c9e231688aa

Alot of the logic from the CacheHash wasn't really reusable, but I moved the majority of logic for the hashing out of the consensus package towards the json rpc file in the common package. The response now exposes a `CanonicalHash` function which sorts the result prior to hashing.

If we don't want to maintain canonicalization ourselves, there are some libraries available that we could pull in - although many of these have few maintainers / or stale trunks.
- https://github.com/cyberphone/json-canonicalization
- https://github.com/ucarion/jcs
- https://github.com/gowebpki/jcs

PR #267, opened 2025-05-08
fix: correct locking for shared state counter variables

aramalipoor:


data/shared_state_variable.go
```diff
@@ -280,16 +184,60 @@ func (c *counterInt64) OnValue(cb func(int64)) {
...
 
+func (c *counterInt64) OnLargeRollback(cb func(currentVal, newVal int64)) {
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	c.largeRollbackCallback = cb
+}
+
 func (c *counterInt64) setValue(val int64) {
 	c.value.Store(val)
-	c.lastUpdated.Store(time.Now().UnixNano())
```
kasrakhosravi, line 285: is it safe to remove this one?

Issue #239, opened 2025-04-24
crash: Docker container exit code 139


valamidev:
**Describe the bug**

```
      "Id": "dad0ec03bcd39b89b6f9cecd09f57fa8ecc6efadc5c772ec44853bea66b0c3a4",
        "Created": "2025-04-24T12:22:37.255835303Z",
        "Path": "/root/erpc-server",
        "Args": [],
        "State": {
            "Status": "exited",
            "Running": false,
            "Paused": false,
            "Restarting": false,
            "OOMKilled": false,
            "Dead": false,
            "Pid": 0,
            "ExitCode": 139,
            "Error": "",
            "StartedAt": "2025-04-24T12:22:37.424844678Z",
            "FinishedAt": "2025-04-24T12:22:48.627481794Z"
        },
```

While running ERPC container with a single upstream it just exit without any log or exit code, sometimes it can run up to 5 minutes sometimes it crash in few second. 

Same happen for v0.0.35 and for v0.0.44.

aramalipoor:
Hey @valamidev this most probably related to ARM64 architecture #231 can you confirm? If it's arm64 (e.g. Mac M1..M4) then there's a limitation on the Sonic library dependency, best is to build the image only for amd64.

From 0.0.45 onwards the image will only have amd64, until we find a better efficient JSON parsing library to replace Sonic.

valamidev:
Yes, I did run on Mac M1.

aramalipoor:
I see, in that case other users have built an amd64 image and they could continue with no issues on performance or otherwise.

The main image `ghcr.io/erpc/erpc:main` is already amd64 only if you want to try. You can also clone / build locally it's pretty straightforward.

Let me know if you needed any help doing that 👌

Issue #231, opened 2025-04-17
bug: container restarts with 139 SIGSEGV on arm64 (M1/M2/M3/M4) architecture


insider89:
**Describe the bug**
We are running erpc lb on the k8s cluster. We see that application constantly restarted. In the logs I didn't see any error. The exit code is 139, which means Segmentation Fault (SIGSEGV).

**To Reproduce**
1. Use erpc version `0.0.43`

2. Use this erpc.yaml config:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: erpc
  name: erpc
  namespace: atk
spec:
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      app.kubernetes.io/instance: atk
      app.kubernetes.io/name: erpc
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      annotations:
        prometheus.io/port: "4001"
        prometheus.io/scrape: "true"
      creationTimestamp: null
      labels:
        app.kubernetes.io/component: erpc
    spec:
      containers:
        image: ghcr.io/erpc/erpc:0.0.43
        imagePullPolicy: IfNotPresent
        livenessProbe:
          failureThreshold: 3
          httpGet:
            path: /healthcheck
            port: http
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
        name: erpc
        ports:
        - containerPort: 4000
          name: http
          protocol: TCP
        - containerPort: 4001
          name: metrics
          protocol: TCP
        readinessProbe:
          failureThreshold: 3
          httpGet:
            path: /healthcheck
            port: http
            scheme: HTTP
          periodSeconds: 10
          successThreshold: 1
          timeoutSeconds: 1
        resources:
          limits:
            memory: 512Mi
          requests:
            cpu: 100m
            memory: 256Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /root/erpc.yaml
          name: config
          subPath: erpc.yaml
      dnsPolicy: ClusterFirst
      imagePullSecrets:
      - name: image-pull-secret-docker
      - name: image-pull-secret-ghcr
      - name: image-pull-secret-harbor
[...]

PR #268, opened 2025-05-08
fix: bump sonic to v1.13.2 to avoid SIGFAULTs on arm64 architecture

ribeirojose:
A bunch of issues re. aarch64 architecture have been addressed upstream in `sonic` since the `v1.12.3` release which is currently being used by `erpc`. Namely, probably most importantly to us the following:

https://github.com/bytedance/sonic/releases/tag/v1.13.2
https://github.com/bytedance/sonic/releases/tag/v1.12.9

Bumping it addresses #239 and #231 and potentially unblocks releases of ARM images which erpc stopped building as of #233. 

We attempted a new build on our side for ARM and have it running in production in a K8s cluster successfully. Only issue we see right now is memory-related but that seems unrelated to this update.

Tests pass running with `make test`, but honestly I wasn't sure whether I needed to do anything else to make a contribution to the repo apart from what's in https://github.com/erpc/erpc/blob/main/CONTRIBUTING.md. Happy to do anything else that would still be needed. I'd love to add a test scenario that would've been able to catch this but I'm not honestly too versed in go or in erpc's codebase to feel ready to do that 😅 so happy to get assistance if you think it's a must.

Thanks for the awesome project!


aramalipoor:
Thanks @ribeirojose for looking into this, `arm64` is one of those things I personally really like to get done. I can get into testing/investigating next week. Need to test this under some real workflow + reviewing their code changes to see if they actually addresses the use of Unsafe functions.

ribeirojose:
> Thanks @ribeirojose for looking into this, `arm64` is one of those things I personally really like to get done. I can get into testing/investigating next week. Need to test this under some real workflow + reviewing their code changes to see if they actually addresses the use of Unsafe functions.

Thanks for jumping on this, @aramalipoor! Let me know if there’s anything I can do to help with testing or share details from our workload—happy to pitch in wherever’s useful.

aramalipoor:
If you've already tried this in your own env and don't see any random panics that's already a huge good sign. If you've got observations around this in your envs please do share as it'll be super helpful 🙌

aramalipoor:
> Only issue we see right now is memory-related but that seems unrelated to this update.

Also can you elaborate on this, is it a leak, or something else? Relevant:  https://docs.erpc.cloud/operation/production#memory-usage

ribeirojose:
@allangarciasilva is at the forefront of this so tagging you here. Can you give more context about this? Also if you have the capacity feel free to share more details about our setup. 

FWIW @aramalipoor we're pre-launch and just getting started with our production environment so workload is not huge yet. In any case my first goal is to slash some RPC calls for dev, then using erpc to centralize all our RPC dependencies across back/frontend.

> > Only issue we see right now is memory-related but that seems unrelated to this update.
> 
> Also can you elaborate on this, is it a leak, or something else? Relevant: https://docs.erpc.cloud/operation/production#memory-usage

allangarciasilva:
> @allangarciasilva is at the forefront of this so tagging you here. Can you give more context about this? Also if you have the capacity feel free to share more details about our setup.

Thanks for tagging me, @ribeirojose.

Just wanted to chime in with our observations. We've been running our erpc container in a Kubernetes cluster on AWS, specifically on T4g instances which are arm64-based.

Initially, we were using an image based on the `erpc_linux_arm64` binary from release [0.0.45](https://github.com/erpc/erpc/releases/tag/0.0.45) (compiled with Sonic v1.12.3). With that, the container would consistently crash with exit code 139, similar to the issues reported in #231 and #239. We observed that the container would typically not last more than 30-40 seconds before crashing.

We are now using a custom Dockerfile that is essentially based on the 0.0.45 release, with the key difference being the bump of the Sonic library to v1.13.2, as included in @ribeirojose's commit. So far, we've seen the application appear to run correctly and it hasn't shown any panics or seg faults, even after several hours of execution.

For reference, the following Dockerfile was used to build this image and has proven functional in our tests:

```dockerfile
ARG BUILDPLATFORM
FROM --platform=${BUILDPLATFORM} golang:1.23-alpine AS erpc-builder
RUN apk add --no-cache git

ARG ERPC_VERSION=0.0.45
RUN git clone https://github.com/erpc/erpc.git -b ${ERPC_VERSION} --depth 1 /erpc
WORKDIR /erpc

# Update the Sonic library so we can fix the support for ARM 64
RUN go get github.com/bytedance/sonic@v1.13.2 
RUN go mod download

# Build config
ARG TARGETOS
ARG TARGETARCH
ENV CGO_ENABLED=0 \
    GOOS=${TARGETOS} \
    GOARCH=${TARGETARCH}

# Build the final binary
RUN LDFLAGS="-w -s -X common.ErpcVersion=${ERPC_VERSION} -X common.ErpcCommitSha=$(git rev-parse --short HEAD)"; \
    go build -v -ldflags="$LDFLAGS" -a -installsuffix cgo -o /build/erpc ./cmd/erpc/main.go

# Alias for the node-24 image
FROM node:24-alpine3.20 AS node-base

# Install dependencies
FROM node-base AS node-deps

COPY ./package.json /package.json
RUN cd / && npm install

# Runtime image
FROM node-base

COPY --from=erpc-builder /build/erpc /erpc
COPY --from=node-deps /node_modules /node_modules
COPY ./erpc.ts /erpc.ts

ENTRYPOINT ["/erpc"]
CMD ["--config", "/erpc.ts"]
```

Hope this helps! Let me know if any more details about our setup would be useful.

aramalipoor:
@allangarciasilva this is great news! Thanks for confirming the new version of Sonic is running smoothly. I have higher confidence we can bring back the arm64 builds 🥇

aramalipoor:
Thanks for the contribution and testing folks 🚀

PR #320, opened 2025-06-03
fix: excluding syncing and selection policy excluded upstreams in EvmHighestLatestBlockNumber

kasrakhosravi:


upstream/registry.go
```diff
@@ -138,6 +138,13 @@ func (u *UpstreamsRegistry) PrepareUpstreamsForNetwork(ctx context.Context, netw
 	for _, p := range allProviders {
 		// Capture loop variable in local scope
 		provider := p
+
+		taskName := fmt.Sprintf("provider/%s/network/%s", provider.Id(), networkId)
+		if _, loaded := u.initializer.GetTask(taskName); loaded {
+			u.logger.Debug().Str("taskName", taskName).Msg("provider bootstrap task already scheduled, skipping.")
+			continue
+		}
+
```
aramalipoor: This shouldn't be needed because initializer logic for tasks is already idempotent, same task name will not create new tasks.

is it not behaving as you expect at the moment?

erpc/networks.go
```diff
@@ -110,6 +135,22 @@ func (n *Network) EvmHighestFinalizedBlockNumber(ctx context.Context) int64 {
...
+
+		// Check if the node is syncing - skip syncing nodes as their block numbers may be unreliable
+		if u.EvmSyncingState() == common.EvmSyncingStateSyncing {
+			n.logger.Debug().Str("upstreamId", u.Id()).Msg("skipping syncing upstream for highest finalized block calculation")
+			continue
+		}
+
+		// Check if upstream is excluded by selection policy
+		if n.selectionPolicyEvaluator != nil {
+			// We use "eth_getBlockByNumber" as it's a common method that would be used to get finalized block
```
aramalipoor, line 138: Smart!

upstream/registry.go
kasrakhosravi: aha..no it was mostly as a prevention but came to my mind that this prevention should already happen at the initialzier level

erpc/networks.go
```diff
@@ -42,6 +42,15 @@ type Network struct {
...
 	// Initialize policy evaluator if configured
 	if n.cfg.SelectionPolicy != nil {
+
+		initCtx, cancelInit := context.WithTimeout(ctx, 30*time.Second)
+		defer cancelInit()
+
+		err := n.upstreamsRegistry.GetInitializer().WaitForTasks(initCtx)
+		if err != nil {
+			n.logger.Warn().Err(err).Msg("failed to wait for initial upstreams")
+		}
```
aramalipoor: This is probably fail if 1 upstream is broken (which happens a lot) and that might halt the whole network initi and users seeing "no upstreams found for network evm:1"

Can you trace back the call chain to see if this timeout is more than 30s and the initializer never succeeds what will users see?

The ideal scenario is requests "wait" for this Bootstrap() to finish, and worst case end up in timeout response. Which means if bootstrapping finishes in 500ms-1s then from perspective of first-ever request for this network user will just see a 1s add. latency, but NOT the "no upstreams found for evm:1"

aramalipoor: Another thing to consider here probably we want this behavior:
1. A low timeout second for wait (like 1-3s)
2. It silently ignores only if "at least 1 upstream is active" (currently it always silently ignores)
3. Fail the bootstrap if after the timeout still 0 upstreams are active for this network

We need to see the effect of (3) in adverse situations, will it create too many redundant attempts of Bootstrap if things are constantly failing, or will it nicely only have 1 bootstrap attempt at any given time even if many requests are bombarding. This is the situation on fresh deployments, and we have 500/1k incoming RPS, we don't want to overload the bootstrapping logic, only 1 attempt at any given time and all requests to be waiting for that 1 attempt to finish/fail.

util/initializer.go
```diff
@@ -384,6 +384,14 @@ func (i *Initializer) Errors() error {
 	return errors.Join(errs...)
 }
 
+func (i *Initializer) GetTask(name string) (*BootstrapTask, bool) {
+	task, loaded := i.tasks.Load(name)
+	if !loaded {
+		return nil, false
+	}
+	return task.(*BootstrapTask), true
+}
```
aramalipoor: Probably not needed anymore

upstream/registry.go
```diff
@@ -153,16 +153,40 @@ func (u *UpstreamsRegistry) PrepareUpstreamsForNetwork(ctx context.Context, netw
...
+		case <-timeout.C:
+			// Timeout expired - check if we have any active upstreams
+			u.upstreamsMu.RLock()
+			upstreams := u.networkUpstreams[networkId]
+			activeCount := 0
+			for _, ups := range upstreams {
+				if !u.metricsTracker.IsCordoned(ups, "*") {
+					activeCount++
+				}
+			}
```
aramalipoor: There should better ways to assert being active/healthy. The cordoning is only if selection policy is active and it had a chance to do a evaluation at least once.

The fact of upstream being in "u.networkUpstreams" means it's already initialized. But how many are necessary to return successfully is debatable.

---

P.S. The ticket.C section is already taking care of asserting active upstreams and <-ctx.Done(): already takes care of timeout (i.e. if it takes long), we don't need this <-timeout.C case afaict.

PR #328, opened 2025-06-06
test: param marshalling

mikelsr:
Tests for #325, which did not solve the issue.


aramalipoor:
@mikelsr looks like #330 is working, I tested it with the pimlico methods you mentioned, let me know if it works for you as well.

PR #331, opened 2025-06-10
feat: cache empty response based on config policy

kasrakhosravi:
skip caching empty responses when it is allowed by the user via config

```
ignore: (default) Ignore empty responses and do not cache them
allow: Allow caching empty responses as well
only: Only cache empty responses, e.g. if you want to give different TTL
```


architecture/evm/json_rpc_cache.go
```diff
@@ -290,6 +290,7 @@ func (c *EvmJsonRpcCache) Get(ctx context.Context, req *common.NormalizedRequest
 	return common.NewNormalizedResponse().
 		WithRequest(req).
 		WithFromCache(true).
+		WithEmptyBehavior(policy.EmptyState()).
```
aramalipoor: Do we really need to pass around the policy empty state via response? Can we just handle the requirement within Get() func itself?

e.g. if it says "empty response must be ignored" we just return a NotFound error...

aramalipoor: Separately, do we need to change anything in Set() func to respect this policy flag?

architecture/evm/json_rpc_cache.go
```diff
@@ -264,6 +264,39 @@ func (c *EvmJsonRpcCache) Get(ctx context.Context, req *common.NormalizedRequest
 		return nil, nil
 	}
 
+	resp := common.NewNormalizedResponse().
+		WithRequest(req).
+		WithFromCache(true).
+		WithJsonRpcResponse(jrr)
+
+	if resp.IsResultEmptyish() {
+		switch policy.EmptyState() {
```
aramalipoor: Probably we should add IsResultEmptyish() on jrr object itself, rather than create a temp resp, on each request under heavy traffic that could be too much memory allocations.

kasrakhosravi: good catch

architecture/evm/json_rpc_cache.go
kasrakhosravi: seems like we already got it covered here? 

https://github.com/erpc/erpc/blob/d3a6d82d0763c2dd5fa62457cab709bae90980e3/architecture/evm/json_rpc_cache.go#L418

aramalipoor (approved):

PR #332, opened 2025-06-11
fix: set "intrinsic gas too high" error as ErrEndpointExecutionException in error normalizer

kasrakhosravi:


architecture/evm/error_normalizer.go
```diff
@@ -227,6 +227,18 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 			)
 		}
 
+		if strings.Contains(strings.ToLower(msg), "intrinsic gas too high") {
```
aramalipoor: Can we put it among the list in previous "if"?

PR #334, opened 2025-06-11
fix: treating "ChainException" and "invalid block range" drpc errors as ErrEndpointMissingData

kasrakhosravi:


thirdparty/drpc.go
```diff
@@ -171,6 +171,19 @@ func (v *DrpcVendor) GetVendorSpecificErrorIfAny(req *common.NormalizedRequest,
 				),
 			)
 		}
+
+		if strings.Contains(msg, "ChainException") ||
```
aramalipoor: This feels too broad, should we also include `code=40000`?

PR #336, opened 2025-06-11
chore: make eth_sendRawTransaction idempotent

wparr-circle:
Fixes https://github.com/orgs/erpc/projects/6?pane=issue&itemId=98141625

We recently observed that whenever a failsafe duplicated execution is executed for non-idempotent methods such as eth_getBlockByNumber, we would observe errors.
This change introduces some post response handling for eth_getBlockByNumber which allows handling of errors which may not be the result of a user such as `already known` (transaction already in the mempool), or `nonce too low` (transaction already executed) for a short period of time provided transaction hash derived from signed payload can be exact matched to recent on-chain transaction.

### DIFF
--- a/architecture/evm/eth_sendRawTransaction.go
+++ b/architecture/evm/eth_sendRawTransaction.go
@@ -1,0 +1,100 @@
+package evm
+
+import (
+	"context"
+	"encoding/hex"
+	"fmt"
+	"strings"
+
+	"github.com/erpc/erpc/common"
+	"github.com/ethereum/go-ethereum/core/types"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/trace"
+)
+
+func upstreamPostForward_eth_sendRawTransaction(ctx context.Context, n common.Network, u common.Upstream, rq *common.NormalizedRequest, rs *common.NormalizedResponse, re error, _ bool) (*common.NormalizedResponse, error) {
+	ctx, span := common.StartDetailSpan(ctx, "Upstream.PostForwardHook.eth_sendRawTransaction", trace.WithAttributes(
+		attribute.String("request.id", fmt.Sprintf("%v", rq.ID())),
+		attribute.String("network.id", n.Id()),
+		attribute.String("upstream.id", u.Id()),
+	))
+	defer span.End()
+
+	if re != nil {
+		err, ok := re.(common.StandardError)
+		if !ok {
+			return rs, re
+		}
+
+		deepestMessage := err.DeepestMessage()
+
+		switch deepestMessage {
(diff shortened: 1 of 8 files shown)


architecture/evm/eth_sendRawTransaction.go
```diff
@@ -0,0 +1,100 @@
...
+			}
+
+			txnHash := "0x" + hex.EncodeToString(txn.Hash().Bytes())
+
+			jrr, err := common.NewJsonRpcResponse(rq.ID(), txnHash, nil)
+			if err != nil {
+				return rs, err
+			}
+
+			if deepestMessage == "nonce too low" {
```
aramalipoor, line 46: I suggest we move these error detections to [error_normalizer.go](https://github.com/erpc/erpc/blob/main/architecture/evm/error_normalizer.go) so it's consistent with the rest of the flow

architecture/evm/eth_sendRawTransaction.go
```diff
@@ -0,0 +1,100 @@
...
+				}
+
+				// todo(wparr-circle): could we short circuit this by checking the timestamp of the pre-signed transaction to avoid
+				// sending a eth_sendRawTransaction request on every nonce too low error? Probably not...
+				_, err := evmUpstream.EvmGetTransactionByHash(ctx, txnHash)
+				if err != nil {
+					break
+				}
+
+				// todo(wparr-circle): we should check if the transaction is older than t seconds ago, and if so, return the original error...
```
aramalipoor, line 59: Looks like if we compare v,r,s (from `tx.RawSignatureValues()`) then it's 100% the same tx 🤔 

Here's some gpt suggestions:
```golang
package evm

import (
	"bytes"
	"math/big"

	"github.com/ethereum/go-ethereum/common"
	"github.com/ethereum/go-ethereum/core/types"
)

// IsExactTxMatchSigNonce does the minimal field‑level check.
// Assumes EIP‑2 rules (low‑s) are enforced network‑side.
func IsExactTxMatchSigNonce(a, b *types.Transaction) bool {
	if a.Nonce() != b.Nonce() {
		return false
	}
	ra, sa, va := a.RawSignatureValues()
	rb, sb, vb := b.RawSignatureValues()
	return ra.Cmp(rb) == 0 && sa.Cmp(sb) == 0 && va.Cmp(vb) == 0
}
```

and usage like:
```golang
rawTx := new(types.Transaction)
_ = rawTx.UnmarshalBinary(rawSignedBytes)   // your local tx

// remoteTx := decode(jsonRpcResponse)      // tx fetched via eth_getTransactionByHash

if IsExactTxMatchHash(rawTx, remoteTx) {
	fmt.Println("✔ Perfect match (hash)")
} else {
	fmt.Println("✘ Mismatch")
}
```

PR #338, opened 2025-06-13
fix: robust selection of upstreams in consensus policy

aramalipoor:


erpc/networks_consensus_test.go
```diff
@@ -419,12 +549,187 @@ func TestNetwork_Consensus(t *testing.T) {
 				assert.NoError(t, err)
 				assert.NotNil(t, actualJrr)
 
-				assert.Equal(t, string(expectedJrr.Result), string(actualJrr.Result))
+				if actualJrr != nil {
```
wparr-circle: This could potentially skip a test evaluation if the expected exists.

Maybe add a case that checks that if expected is set, we requireJrr not nil to proceed otherwise if actualJrr is set we require an expected.

PR #354, opened 2025-06-25
chore: make eth_sendRawTransaction idempotent 

kasrakhosravi:


architecture/evm/eth_sendRawTransaction.go
```diff
@@ -0,0 +1,109 @@
...
+		if _, isDuplicateNonce := err.(*common.ErrEndpointDuplicateNonce); isDuplicateNonce {
+			txn, err := getTransactionFromRequest(ctx, rq)
+			if err != nil {
+				return rs, err
+			}
+
+			txnHash := "0x" + hex.EncodeToString(txn.Hash().Bytes())
+
+			// For nonce too low, verify the transaction exists on-chain and is exactly the same
+			if strings.Contains(deepestMessage, "nonce too low") {
```
aramalipoor: Is this check really needed? If so then it means we should have 2 different error types to avoid any sort of string checks in our domain logic (we should isolated text-based error detection to error_normalizer only)

PR #388, opened 2025-07-17
feat: extend matchers to support params and finality in failsafe and upstream filtering

kasrakhosravi:
> [!NOTE]
> **Medium Risk**
> Introduces a new matcher-based configuration schema and rewires cache/failsafe selection logic (including legacy conversion), which can change runtime behavior for caching and resilience policies if configs are interpreted differently. Also modifies Docker cross-build settings, which could affect multi-arch image builds.
> 
> **Overview**
> **Adds a matcher object model (`matchers`) for cache policies and failsafe policies**, supporting network/method/params/finality constraints plus `include`/`exclude` actions, with validation that normalizes rule sets (including auto-inserting catch-all rules in some cases).
> 
> **Reworks EVM JSON-RPC cache policy selection and empty-result handling** to carry the *specific matcher that matched* through GET/SET (`PolicyWithMatcher`) so empty caching behavior is applied per matcher, and updates cache policy matching logic (including legacy field conversion) accordingly.
> 
> **Changes failsafe defaulting semantics** so user-defined failsafe policies *replace* defaults rather than being merged by method/finality, adds deep-copy fixes for matcher pointers, and updates tests/docs to reflect the new schema and behavior. Separately, updates the `Dockerfile` to use `--platform=$BUILDPLATFORM` and `TARGETOS/TARGETARCH` for reliable multi-arch builds, and extends `.gitignore`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1d4d4be1e7254f9054dbe1305488ee7690e858c8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


docs/pages/config/matcher.mdx
```diff
@@ -1,52 +1,631 @@
...
+- **Performance**: Rules are evaluated once and cached per request
+- **Flexibility**: Supports both declarative (YAML) and programmatic (JavaScript) rules
+- **Priority System**: Rules are evaluated in priority order
 
-// Match only numeric values (e.g. ignore block tags)
-0x*
+## Configuration
+
+Routing rules can be configured at the network level:
+
```
aramalipoor: It's better to move these docs to upstreams page, because this Matchers page is for the matchers syntax which is used in other places as well.

Good to to only have only have general syntax definition in this page, wdyt?

erpc/networks.go
```diff
@@ -227,41 +228,27 @@ func (n *Network) EvmLeaderUpstream(ctx context.Context) common.Upstream {
...
 	}
 
-	// Then, try to find a match for finality only
 	for _, fe := range n.failsafeExecutors {
-		if fe.method == "*" && len(fe.finalities) > 0 {
-			if slices.Contains(fe.finalities, finality) {
+		if fe.config != nil && len(fe.config.Matchers) > 0 {
+			matcher := matchers.NewConfigMatcher(fe.config.Matchers)
+			result := matcher.MatchRequest(networkId, method, params, finality)
+			if result.Matched && result.Action == common.MatcherInclude {
```
aramalipoor: For evaluation probably we can have a simpler external interface?
```
if matcher.Match(fe.config.Matchers, req) {
  // ...
}
```

Then we can avoid the req handling above as well?

matchers/matcher.go
```diff
@@ -0,0 +1,218 @@
...
+			return common.MatchResult{Matched: false, Action: common.MatcherInclude}
+		}
+	}
+
+	// Fallback to exclude for safety
+	return common.MatchResult{Matched: false, Action: common.MatcherExclude}
+}
+
+// MatchForCache evaluates if a response should be cached
+func (m *ConfigMatcher) MatchForCache(networkId, method string, params []interface{}, finality common.DataFinalityState, isEmptyish bool) common.MatchResult {
```
aramalipoor: Any reason for this? 

We can have `matchers.Match(cfg, req, resp)` exposed from the package, very simple and works in every place

matchers/matcher.go
```diff
@@ -0,0 +1,218 @@
...
+type ConfigMatcher struct {
+	configs []*common.MatcherConfig
+}
+
+// NewConfigMatcher creates a new matcher with the given configs
+func NewConfigMatcher(configs []*common.MatcherConfig) *ConfigMatcher {
+	return &ConfigMatcher{
+		configs: configs,
+	}
+}
```
aramalipoor: What would be the benefit of creating a separate object and have an instance-based design?

erpc/networks.go
aramalipoor: FYI since this func is running on every single request, it must be absolute minimum cpu/memory usage (e.g. creating a new instance of an object happens 5k/sec), I was thinking of precompiling the matcher rules, but don't think really is possible, the match logic we have seems to be simplest possible.

matchers/matcher.go
kasrakhosravi: totally. if we go with a simple `matchers.Match`, no need for this anymore

matchers/matcher.go
```diff
@@ -0,0 +1,185 @@
...
+
+// Match provides a simplified interface for request/response matching
+func Match(configs []*common.MatcherConfig, req *common.NormalizedRequest, resp *common.NormalizedResponse) bool {
+	if len(configs) == 0 {
+		return false
+	}
+
+	// Extract required fields from request
+	method, _ := req.Method()
+	finality := req.Finality(nil)
```
aramalipoor: This must be real context, nil probably just panics in prod, `Match(ctx context.Context, ...`

common/data.go
```diff
@@ -58,6 +58,27 @@ func (f *DataFinalityState) UnmarshalYAML(unmarshal func(interface{}) error) err
 	return fmt.Errorf("invalid data finality state: %s", s)
 }
 
+// DataFinalityStateSlice provides custom YAML unmarshaling for []DataFinalityState
+type DataFinalityStateSlice []DataFinalityState
+
+func (f *DataFinalityStateSlice) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
aramalipoor: Is this necessary? The DataFinalityState already has an UnmarshalYAML, isn't it working when accepted as array?

common/config.go
```diff
@@ -705,9 +705,33 @@ func (c *EvmUpstreamConfig) Copy() *EvmUpstreamConfig {
 	return copied
 }
 
+type MatcherConfig struct {
+	Network  string                 `yaml:"network,omitempty" json:"network"`
+	Method   string                 `yaml:"method,omitempty" json:"method"`
+	Params   []interface{}          `yaml:"params,omitempty" json:"params"`
+	Finality DataFinalityStateSlice `yaml:"finality,omitempty" json:"finality" tstype:"DataFinalityState[]"`
```
aramalipoor:
```suggestion
	Finality []DataFinalityState `yaml:"finality,omitempty" json:"finality" tstype:"DataFinalityState[]"`
```
Any downsides to something like this?

matchers/matcher.go
```diff
@@ -0,0 +1,185 @@
...
+	var params []interface{}
+	if jrpcReq, err := req.JsonRpcRequest(); err == nil && jrpcReq != nil {
+		params = jrpcReq.Params
+	}
+
+	// Determine if response is empty
+	isEmptyish := false
+	if resp != nil {
+		isEmptyish = resp.IsObjectNull() || resp.IsResultEmptyish()
+	}
```
aramalipoor, line 178:
```suggestion
	isEmptyish := resp != nil && !resp.IsObjectNull() && !resp.IsResultEmptyish(ctx)
```
Also this way it is `true` when resp doesn't even exist or is null.

Ideally anywhere we have IsResultEmptyish we should pass the ctx, it gives us correct tracing when debugging, it's optional because as far I can tell it wouldn't break prod without it, just won't give us correct traces.

common/defaults.go
```diff
@@ -1397,24 +1447,46 @@ func (u *UpstreamConfig) SetDefaults(defaults *UpstreamConfig) error {
...
+										methodMatch, _ = WildcardMatch(dfsMatcher.Method, fsMatcher.Method)
+									}
+									finalityMatch = matchFinalities(dfsMatcher.Finality, fsMatcher.Finality)
+									if methodMatch && finalityMatch {
+										break
+									}
+								}
+								if methodMatch && finalityMatch {
+									break
+								}
```
aramalipoor: Should we also apply other matching criteria vs just method/finality? 

I'm not even sure if this while defaulting logic is correct and intuitive, in terms of what the users would expect when defining failsafe policies in the upstreamDefaults + upstreams themselves.

The current logic probably has some unexpected behaviors, as in mixing certain defaults even though user doesn't expect due to criteria difference.

If there's no clean way of doing this "defauling" logic can be that it overrides the whole policies?

common/defaults.go
```diff
@@ -33,6 +33,51 @@ func matchFinalities(finalities1, finalities2 []DataFinalityState) bool {
...
+	hasUpstreamPolicies := fs.Retry != nil || fs.Hedge != nil || fs.CircuitBreaker != nil ||
+		fs.Timeout != nil || fs.Consensus != nil
+	hasDefaultPolicies := dfs.Retry != nil || dfs.Hedge != nil || dfs.CircuitBreaker != nil ||
+		dfs.Timeout != nil || dfs.Consensus != nil
+
+	// If upstream has policies and default has different policies, don't match
+	// This prevents mixing policies from different sources
+	if hasUpstreamPolicies && hasDefaultPolicies {
+		return false
+	}
```
aramalipoor: Wouldn't this be always true (and return)? Because there's always some policies defined...

If we want to ignore defaults we have to do one step before applying defaults (something like "if any policies are defined do not apply the defaults")

I'm not sure what is the correct defaulting algorithm for these it needs some thinking...
Another way to solve it is to extend the `methodMatch && finalityMatch` to consider other fields as well, i.e. params, empty, network...

common/defaults.go
```diff
@@ -1329,8 +1374,15 @@ func (p *ProviderConfig) SetDefaults(upsDefaults *UpstreamConfig) error {
 		p.UpstreamIdTemplate = "<PROVIDER>-<NETWORK>"
 	}
 	if p.Overrides != nil {
+		// Do not inherit Failsafe defaults into overrides; overrides should replace them.
```
aramalipoor: What does this mean exactly?

common/config.go
```diff
@@ -744,6 +771,23 @@ func (c *FailsafeConfig) Copy() *FailsafeConfig {
...
+		copied.Matchers = make([]*MatcherConfig, len(c.Matchers))
+		for i, matcher := range c.Matchers {
+			if matcher != nil {
+				matcherCopy := &MatcherConfig{}
+				*matcherCopy = *matcher
+				// Deep copy the Params slice
+				if matcher.Params != nil {
+					matcherCopy.Params = make([]interface{}, len(matcher.Params))
+					copy(matcherCopy.Params, matcher.Params)
+				}
```
aramalipoor, line 1136: Do we need similar copy for finality array? Or it'll work cuz of primitives?

kasrakhosravi: great catch..yeah I think regardless of primitive, better to copy it as a slice

data/cache_policy.go
```diff
@@ -160,92 +184,26 @@ func (p *CachePolicy) MatchesSizeLimits(size int) bool {
...
-			if err != nil {
-				return false, err
-			}
-			if !match {
-				return false, nil
+	// For new matcher configs, find the first INCLUDE matcher (skip auto-added excludes)
+	if len(p.config.Matchers) > 0 {
+		for _, matcher := range p.config.Matchers {
+			if matcher != nil && matcher.Action == common.MatcherInclude {
+				return matcher.Empty
```
aramalipoor: Two important aspects for this:
* We should not at need to handle legacy fields here, they must be fully migrated to new approach, and the new approach should be able to behave as we intend, if not possible we'd need a design update.
* The empty state behavior should be used from the exact matcher that matched the request initially, this current code is just covering 1 edge-case which is when we have exclude on default matcher + have different empty on the next matcher.

kasrakhosravi: @aramalipoor I don't think we are that far away from having this in the core as a feature in the core. can you have another fresh look at it after we release the current version with rate limiter and block timestamp distance improvements?

aramalipoor: We need to test and benchmark this carefully to ensure it won't impact prod as it's basically run on every single request and some tests are needed with a load of like 8k-10k rps with a normal compute instance.

data/cache_policy.go
```diff
@@ -60,192 +102,120 @@ func (p *CachePolicy) MarshalJSON() ([]byte, error) {
...
+			match, err := common.WildcardMatch(matcher.Method, method)
+			if err != nil || !match {
+				continue
 			}
 		}
-		return true, nil
-	case []interface{}:
-		// For arrays, match each element
-		paramArray, ok := param.([]interface{})
-		if !ok || len(p) != len(paramArray) {
```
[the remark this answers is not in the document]
kasrakhosravi: False positive — the asymmetry is architectural, not a defect.

`MatchesForGetWithMatcher` deliberately omits the `empty:` check because at GET-policy-selection time we don't have the cached value yet, so emptiness is unknowable. The filter happens post-fetch in `EvmJsonRpcCache.Get` around `json_rpc_cache.go:336`, where the `if jrr.IsResultEmptyish() { switch policyWithMatcher.EmptyState() ... }` block applies the matcher's `Empty` constraint once the value is in hand.

End-to-end coverage already exists at `evm_json_rpc_cache_test.go:1893+` (the "empty behavior set to ignore/allow/only" table-driven subtests). 

Documented the architectural intent in `MatchesForGetWithMatcher`'s docstring and added a focused regression test (`TestCachePolicy_MatchersIntegration` → `MatchesForGet ignores empty constraint by design (filter happens post-fetch)`) so a future refactor "tightening" the GET path can't accidentally reject all GETs for matchers with `empty:only`. Both shipped in commit 1d4b523.

PR #406, opened 2025-08-11
feat: chainId config verification + network bootstrap improvements

### LINKED ISSUE #301: bug: configuration allows for mixing chains/networks
**Describe the bug**
We recently had a small issue where whilst defining eRPC configuration, we accidentally mixed up ARB and AVAX - resulting a configuration similar to:
```yaml
      - id: avax-fuji
        upstreams:
          - endpoint: https://avalanche-fuji-c-chain-rpc.publicnode.com
          - endpoint: https://arbitrum-sepolia-rpc.publicnode.com
        upstreamDefaults:
          group: "default"
          evm:
            chainId: 43113
```

This ended up producing issues with the metrics `erpc_upstream_{block, finalization}_head_lag`, where the arbitrum upstream was considered the _most_ up to date upstream, despite not actually being an avalanche fuji upstream, which proceeded to be scored slightly higher allowing more traffic to weighted towards the abitrum upstream.

**To Reproduce**
1. Use erpc version (e.g `0.0.40`) 0.0.45 was used, should also affect 0.0.46

2. Use this erpc.yaml config:
```yaml
logLevel: debug
projects:
  - id: main
     upstreams:       
       - id: avax-fuji
          upstreams:
            - endpoint: https://avalanche-fuji-c-chain-rpc.publicnode.com
            - endpoint: https://arbitrum-sepolia-rpc.publicnode.com
          upstreamDefaults:
            group: "default"
            evm:
              chainId: 43113
# ...
```

3. Make this request:
```bash
curl --location 'http://localhost:4000/main/evm/43113' \
--header 'Content-Type: application/json' \
--data '{
    "params": [
        "latest",
        false
    ],
    "id": 9199,
    "jsonrpc": "2.0"
}'
```

**Expected behavior**
Requests should be proxied to avax only, as the only upstream which matches chainId 43113

**Screenshots**

**Relevant Logs:**
```jsonl
// ...
```

**Additional context**

kasrakhosravi:
fixes #301


upstream/upstream.go
```diff
@@ -922,28 +922,31 @@ func (u *Upstream) detectFeatures(ctx context.Context) error {
...
-			if err != nil {
-				return common.NewErrUpstreamClientInitialization(
-					&common.BaseError{
-						Code:  "ErrUpstreamChainIdDetectionFailed",
-						Cause: err,
-					},
-					u,
-				)
+		// Always detect the actual chainId from the endpoint to avoid cross-network misconfiguration.
+		// If user provided a chainId and it mismatches, prefer the detected value to ensure correct routing/metrics.
```
aramalipoor: I think we should actually explicitly fail here, so user does a double check why their expected chain ID doesn't match... if they want auto-detection they can just leave it as is.

aramalipoor:
@kasrakhosravi it would be great to have such check also in "validate" sub-command.

Btw for the failing tests you can rebase/merge with latest main branch after #405 is merged

upstream/upstream.go
```diff
@@ -1047,6 +1047,41 @@ func (u *Upstream) detectFeatures(ctx context.Context) error {
 					u,
 				)
 			}
+		} else if common.IsChainIdChecksEnabled(ctx) {
```
aramalipoor: This is bit of a hack... It's better if we just have a separate isolated path for Validate() command.

It can be a simple loop that calls Upstream.EvmGetChainId(ctx) for them and checks. We can keep this core app logic clean and it assumes if a node gets here it's already validated.

telemetry/metrics.go
```diff
@@ -409,6 +410,14 @@ var (
...
 )
 
+var telemetryInitOnce sync.Once
+
+func init() {
+	telemetryInitOnce.Do(func() {
+		_ = SetHistogramBuckets("")
+	})
+}
+
```
aramalipoor: This is risky (in terms of ordering), where exactly is it needed? We can make run only in that flow. If it's for validation path we can run it there.

erpc/init.go
```diff
@@ -77,6 +77,11 @@ func Init(
 		return err
 	}
 
+	// Bootstrap core before starting servers so routes are ready
+	if err := erpcInstance.Bootstrap(appCtx); err != nil {
+		return err
+	}
```
aramalipoor: This is also a significant change (looks good to me though) and it might impact our zero-downtime rollouts in prod...

aramalipoor: Let's keep it but not release until we're sure it doesn't impact rollouts (I can test after this PR is merged)

telemetry/metrics.go
kasrakhosravi: yeah I agree. this is needed for the validation flow. will run it there

cmd/erpc/main.go
```diff
@@ -81,7 +84,54 @@ func main() {
...
+			if err := erpc.AnalyseConfig(cfg, logger); err != nil {
+				return err
+			}
+
+			// Ensure telemetry histograms are initialized for validation path
+			var bucketStr string
+			if cfg.Metrics != nil && cfg.Metrics.HistogramBuckets != "" {
+				bucketStr = cfg.Metrics.HistogramBuckets
+			}
+			_ = telemetry.SetHistogramBuckets(bucketStr)
```
aramalipoor: Shouldn't we return this error instead of silently ignoring? This is basically validating if buckets are correct syntax?

PR #444, opened 2025-09-12
fix(thirdparty/dwellir): add Hyperliquid (999), XDC (50), IoTeX (4689)

eliasfaltin:
- Adds chain 999: api-hyperliquid-mainnet-evm.
- Adds chain 50: api-xdc-mainnet.
- Adds chain 4689: api-iotex-mainnet.


thirdparty/dwellir.go
```diff
@@ -19,6 +19,7 @@ var dwellirNetworkSubdomains = map[int64]string{
 	44:        "api-darwiniacrab.n",
 	46:        "api-darwinia.n",
 	40:        "api-xdc-mainnet.n",
+	50:        "api-xdc-mainnet.n",
```
aramalipoor: Why they seem to have same domain?

PR #463, opened 2025-10-06
feat: add NetworkLatestBlockTimestampLag metric to show distance of current latest block vs current time 

kasrakhosravi:


architecture/evm/evm_state_poller.go
```diff
@@ -688,6 +698,66 @@ func (e *EvmStatePoller) fetchBlock(ctx context.Context, blockTag string) (int64
 	return blockNum, nil
 }
 
+// fetchBlockWithTimestamp fetches both block number and timestamp for a given block tag.
+// This is used specifically for latest block polling to track timestamp distance metrics.
+func (e *EvmStatePoller) fetchBlockWithTimestamp(ctx context.Context, blockTag string) (int64, int64, error) {
```
aramalipoor: Why not just extend the fetchBlock to expose timestamp? This seems like redundant bloat, any specific reason?

architecture/evm/evm_state_poller.go
```diff
@@ -412,8 +412,18 @@ func (e *EvmStatePoller) PollLatestBlockNumber(ctx context.Context) (int64, erro
 		e.latestBlockFailureCount = 0
 		e.stateMu.Unlock()
 
+		// Update network-level block timestamp distance metric if we have a valid timestamp
+		if blockTimestamp > 0 {
+			e.tracker.SetLatestBlockTimestampForNetwork(
```
aramalipoor: I think we should ensure that this is updated only if latest block on network-level is indeed updated, basically both to be atomic.

We don't want a case where timestamp is updated but the latest block number is still not.

aramalipoor: Haven't looked into it deeply but probably best place to ensure atomic update is in SetLatestBlock of the tracker

architecture/evm/evm_state_poller.go
kasrakhosravi: def. I actually did it like that first, but since fetchBlock seemed a bit sensitive and I only needed timestamp in a specific scenario for latest block fetching, I recreated that function. now this is part of the `fetchBlock`

architecture/evm/evm_state_poller.go
```diff
@@ -682,10 +703,24 @@ func (e *EvmStatePoller) fetchBlock(ctx context.Context, blockTag string) (int64
...
+	}
+
+	// Extract timestamp - this is a hex string representing Unix timestamp
+	timestampStr, err := jrr.PeekStringByPath(ctx, "timestamp")
+	var blockTimestamp int64
+	if err == nil && timestampStr != "" {
+		// Force-copy the small string to avoid any potential reference to backing buffers
+		timestampStr = string(append([]byte(nil), timestampStr...))
+		blockTimestamp, err = common.HexToInt64(timestampStr)
+		if err != nil {
```
aramalipoor: There's a chance providers return timestamp as number (int), not a string, not fully sure if this path already covers it. A unit test for this would be useful.

architecture/evm/evm_state_poller.go
```diff
@@ -412,8 +421,14 @@ func (e *EvmStatePoller) PollLatestBlockNumber(ctx context.Context) (int64, erro
 		e.latestBlockFailureCount = 0
 		e.stateMu.Unlock()
 
+		// Store timestamp atomically for OnValue callback to use
+		if blockTimestamp > 0 {
+			e.latestBlockTimestamp.Store(blockTimestamp)
+		}
```
aramalipoor: Do we need to remove this to avoid breaking atomicity? The SetLatestBlockNumber is already taking care of it atomically?

kasrakhosravi: The OnValue callback only receives the block number as a parameter, but SetLatestBlockNumber needs both the block number and timestamp. The atomic store here provides the timestamp for the callback

```
lbs.OnValue(func(value int64) {
    timestamp := e.latestBlockTimestamp.Load()  // Needs this atomic value
    e.tracker.SetLatestBlockNumber(e.upstream, value, timestamp)
})
```

architecture/evm/evm_state_poller.go
kasrakhosravi: ah didn't expect the timestamp to be an int here. added a unit test for it and also covered it in the logic [here](https://github.com/erpc/erpc/pull/463/files#diff-ca5825fe2addf77a207fed9f0c8effd4711fa72df181df953b5c3569daa3bb50R719)

architecture/evm/evm_state_poller.go
aramalipoor: This looks like a hack. We need to dig deeper here, from my current understanding "OnValue" is triggered when remote replicas sent an update for latest block. If that's the case then we need a different approach altogether, or somehow accommodate that fact.

In worst case we can make the replica syncing to carry metadata for each counter (like this timestamp), but ideally a simpler robust solution can be done.

kasrakhosravi: I see the problem you mentioned. basically the remote replicas can send block updates and due to race condition or parallel requests, we might be using the wrong local timestamp that we are storing

architecture/evm/evm_state_poller.go
```diff
@@ -26,6 +28,62 @@ const FullySyncedThreshold = 4
 // (e.g. when a new network is lazy-loaded from a Repository Provider)
 const DefaultToleratedBlockHeadRollback = 1024
 
+// Block and timestamp encoding constants
+// We use 26 bits for block number (supports up to 67M blocks)
+// and 37 bits for timestamp (supports timestamps until year 6325)
+// Total: 63 bits (leaving sign bit alone to avoid overflow)
+// This allows both values to be atomically synced across replicas
+const (
+	blockNumberBits       = 26
```
kasrakhosravi: is this correct?

architecture/evm/evm_state_poller.go
aramalipoor: re these:
image

1. A variation of this is to only emit this new metrics on the primary replica (not the secondary synced nodes), which means the path from fetch latest -> set block number  -> updates timestamp... but the replicas (which probably receive via OnValue) do not emit that metric (maybe a -1 passed as timestamp indicates to ignore, which is the hacky solution)... the downside of this path is the metric emitted from replicas will be stale, and user needs to know when setting alarms to properly write the promql.
2. Is most complex, but I faced before that we needed some other metadata along with counters... we also need to gracefully handle existing data when users upgrade from a previous version to this version (e.g. store values as json, and discard/covert if not json).
3. That's not gonna help because it'll still be not atomic as required.

architecture/evm/evm_state_poller.go
kasrakhosravi: what is this?

architecture/evm/evm_state_poller.go
kasrakhosravi: 1 and 2 - good points. from the reasonings I heard from you is worth to go for option 2 and take on the extra complexity?

3- a variation of 3 could be a single encoder value which instead of tracking block number and timestamp separately, we pack both into one number.

```
Replica A: Fetches block 1000, timestamp 1699000000
           Packs both: encoded = (1000 << 37) | 1699000000
                       encoded = 137438954399000000
           Syncs: Sends this single number 

Replica B: Receives 137438954399000000
           Unpacks: block = 137438954399000000 >> 37 = 1000
                    timestamp = 137438954399000000 & 0x1FFFFFFFFF = 1699000000
           Both values correct
```

aramalipoor: Option (3) is too fragile and risky. Between 1 and 2 not sure, it depends on the size of complexity, as we're using these counters in many places, we should also consider the scale under which these are updated (in some instances we have 400+ chains and majority with <1s debounce)

architecture/evm/evm_state_poller.go
```diff
@@ -116,7 +121,14 @@ func NewEvmStatePoller(
...
 	lbs.OnValue(func(value int64) {
-		e.tracker.SetLatestBlockNumber(e.upstream, value)
+		// Check if this update is from our local fetch or a remote replica
+		timestamp := int64(0)
+		if value == e.latestLocalBlockNumber.Load() {
+			// This is our own update, use the actual timestamp
+			timestamp = e.latestBlockTimestamp.Load()
+		}
+		// For remote updates, timestamp remains 0 to avoid using stale/incorrect timestamps
+		e.tracker.SetLatestBlockNumber(e.upstream, value, timestamp)
```
aramalipoor: I don't think this will be accurate, there's a chance remote value is also equal to our value. 

If we want to go the route of "only emitting metric from primary" then here we should also pass 0 intentionally.

architecture/evm/evm_state_poller.go
```diff
@@ -412,8 +430,17 @@ func (e *EvmStatePoller) PollLatestBlockNumber(ctx context.Context) (int64, erro
 		e.latestBlockFailureCount = 0
 		e.stateMu.Unlock()
 
+		// Store both the block number and timestamp for this local fetch
+		// The OnValue callback will check if the update matches this block number
+		// and use the correct timestamp accordingly
+		e.latestLocalBlockNumber.Store(blockNum)
```
aramalipoor: This seems like an extra redundancy, we already have the block number (as the counter).

architecture/evm/evm_state_poller.go
```diff
@@ -412,8 +430,17 @@ func (e *EvmStatePoller) PollLatestBlockNumber(ctx context.Context) (int64, erro
...
+		if blockTimestamp > 0 {
+			e.latestBlockTimestamp.Store(blockTimestamp)
+		}
```
aramalipoor: Do we need to update this under the stateMu lock? What are the consequences if do/don't?

architecture/evm/evm_state_poller.go
kasrakhosravi: you are completely right. 0 has to be passed intentionally here

architecture/evm/evm_state_poller.go
kasrakhosravi: I am thinking that we don't even need to store this as part of the `latestBlockTimestamp` as it should be already covered in SetLatestBlockNumber. checking the full code and flow now to make sure there are not leftover logic from previous attempt

architecture/evm/evm_state_poller.go
```diff
@@ -412,8 +420,13 @@ func (e *EvmStatePoller) PollLatestBlockNumber(ctx context.Context) (int64, erro
 		e.latestBlockFailureCount = 0
 		e.stateMu.Unlock()
 
+		// Directly update tracker with the correct timestamp for this locally-fetched block
+		// This happens BEFORE the OnValue callback is triggered, ensuring only the fetching node emits the metric
+		e.tracker.SetLatestBlockNumber(e.upstream, blockNum, blockTimestamp, "evm_state_poller")
```
aramalipoor: Why is the evm_state_poller origin label pass in here?

aramalipoor: It's always the same value + it has the same as default. Any specific reason for doing in such way?

kasrakhosravi: def redundant. fixed

health/tracker.go
```diff
@@ -750,6 +751,23 @@ func (t *Tracker) SetLatestBlockNumber(upstream common.Upstream, blockNumber int
 		g := t.getLatestBlockGauge(t.projectId, "*", netLabel, "*")
 		g.Set(float64(blockNumber))
 		needsGlobalUpdate = true
+
+		// Atomically update timestamp when network-level block number is updated
+		if blockTimestamp > 0 {
+			oldTimestamp := ntwMeta.evmLatestBlockTimestamp.Load()
+			if blockTimestamp > oldTimestamp {
```
aramalipoor: Why do we have this check?

health/tracker.go
```diff
@@ -750,6 +751,23 @@ func (t *Tracker) SetLatestBlockNumber(upstream common.Upstream, blockNumber int
...
+				ntwMeta.evmLatestBlockTimestamp.Store(blockTimestamp)
+
+				// Calculate and record distance metric from EVM state poller
+				currentTime := time.Now().Unix()
+				distance := currentTime - blockTimestamp
+				telemetry.MetricNetworkLatestBlockTimestampDistance.WithLabelValues(
+					t.projectId,
+					netLabel,
+					"evm_state_poller",
+				).Set(float64(distance))
```
aramalipoor: Will negative values work correctly here? I assume in cases where a node is wrongly in future it should correctly show negative to proper inform the user

health/tracker.go
kasrakhosravi: redundant check because of `if blockNumber > oldNtwVal` check above it. fixed

health/tracker.go
kasrakhosravi: we are covering this here (and I already observed it in the metrics as well when testing locally). `distance` can have negative value and will be properly stored via `float64(distance)`and prometheus's gauge metric

PR #477, opened 2025-10-28
feat: add enforceNonNullTaggedBlocks configuration for eth_getBlockByNumber with block tags (pending, latest, etc.)

kasrakhosravi:


architecture/evm/eth_getBlockByNumber.go
```diff
@@ -254,18 +254,41 @@ func enforceHighestBlock(ctx context.Context, network common.Network, nq *common
...
+			isTag = bnp != "" && !strings.HasPrefix(bnp, "0x")
+		}
+	}
+
+	// Only enforce non-null for tagged blocks, not numeric blocks
+	if !isTag {
+		return nr, nil
+	}
+
 	if nr == nil || nr.IsObjectNull() || nr.IsResultEmptyish() {
```
aramalipoor: I think it's cheaper if we first check "emptiness" because parsing the request and reading params might not be necessary

docs/pages/config/projects/networks.mdx
```diff
@@ -403,6 +403,54 @@ export default createConfig({
...
+- **When `false`**: Null responses for tagged blocks are returned as-is without conversion to errors
+- **Scope**: Only applies to block tags, not numeric block numbers (hex values starting with `0x`)
+- **Use case**: Useful for pending blocks that don't exist yet, or when you want to handle null tagged block responses at the application level
+
+**Example:**
+```yaml
+evm:
+  integrity:
+    enforceNonNullTaggedBlocks: false  # Allow null responses for tagged blocks
+```
```
aramalipoor: Let's keep the docs section super compact and simple, only 1 codeblock with all the integrity options and just a comment on top of each for what it means.

Also make sure the config "parents" are properly preserved so the user knows exactly where this will sit, from top-level projects down to these configs.

docs/pages/config/projects/networks.mdx
```diff
@@ -430,7 +478,9 @@ projects:
           # Validate requested block range and ensure upstream has data for both ends.
           # Enabled by default; set to false to skip availability checks.
           integrity:
-            enforceGetLogsBlockRange: true
+            enforceHighestBlock: true              # Default: true
+            enforceGetLogsBlockRange: true         # Default: true
+            enforceNonNullTaggedBlocks: true       # Default: true - Convert null tagged block responses to errors
```
aramalipoor: Why are these added here? this section is about getLogs only

architecture/evm/eth_getBlockByNumber.go
```diff
@@ -254,30 +254,54 @@ func enforceHighestBlock(ctx context.Context, network common.Network, nq *common
...
-			),
-			nr.Upstream(),
-		)
 	}
-	return nr, nil
+
+	// For tagged blocks, check if enforcement is EXPLICITLY disabled in config
+	if isTag {
+		ncfg := network.Config()
+		if ncfg != nil &&
```
kasrakhosravi, line 280: a bit of defensive check here. i saw somewhere in this file also we did a similar pattern

architecture/evm/eth_getBlockByNumber_test.go
```diff
@@ -0,0 +1,246 @@
...
+	"github.com/erpc/erpc/common"
+	"github.com/erpc/erpc/util"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+)
+
+// testNetwork is a simple test implementation of common.Network interface for this file
+type testNetwork struct {
+	cfg *common.NetworkConfig
+}
```
aramalipoor, line 16: Let's avoid creating new mocks. Probably we should add this to AGENTS.md these llms try to create new mocks every time, I think we have now multiple network mocks lol

Probably easier to test from erpc/network_xxxx_test.go so that it uses real objects everywhere, and only mock actual http calls via gock.

PR #572, opened 2025-11-13
feat: add finality filter to ScoreMultiplierConfig

kasrakhosravi:


upstream/registry_test.go
```diff
@@ -777,6 +779,231 @@ func TestUpstreamsRegistry_Multiplier(t *testing.T) {
...
+						blockLag:    0.9,  // Significantly lagging
+						respLatency: 0.1,  // Faster
+					},
+				},
+			},
+			expectedPercents: []struct{ min, max float64 }{
+				{0.75, 0.90}, // Well-synced node should dominate for unfinalized
+				{0.10, 0.25}, // Lagging node penalized heavily despite better metrics elsewhere
+			},
+			description: "Unfinalized requests should heavily prioritize block synchronization",
```
aramalipoor: Few test ideas:
- Unfinalized or Realtime requests prefer lowest block lag (not latency)
- Finalized requests prefer fastest response (lag is ok)

aramalipoor:
For the default scores (where users rely on erpc's common best practices) should we do something like:
- For realtime/unfinalized block lag gets higher multipler than latency
- For unknown/finalized block lag gets lower multipler than latency

common/defaults.go
```diff
@@ -2130,7 +2130,40 @@ func (r *RateLimitAutoTuneConfig) SetDefaults() error {
 }
 
 func (r *RoutingConfig) SetDefaults() error {
-	if r.ScoreMultipliers != nil {
+	if r.ScoreMultipliers == nil || len(r.ScoreMultipliers) == 0 {
```
kasrakhosravi, line 2133: * if the user has not set any `ScoreMultipliers`, we have this pre-defined for them

* but what if the user has `ScoreMultipliers` set and also a finality of `real-time` but not the `BlockHeadLag`; in that case do we want to have the per finality score for it or the generic ones here?

https://github.com/erpc/erpc/pull/572#diff-120763f2d6d7a51b93b2561f5d964260a804fb8a86ed1a5074aa3f8b5a0dfa93R2180

aramalipoor: I think if user has defined anything we shouldn't apply any defaults, that'd be too magic/suprising

kasrakhosravi: totally agree

common/defaults.go
```diff
@@ -2130,7 +2130,40 @@ func (r *RateLimitAutoTuneConfig) SetDefaults() error {
...
+				ErrorRate:       util.Float64Ptr(4.0),
+				RespLatency:     util.Float64Ptr(6.0),
+				TotalRequests:   util.Float64Ptr(1.0),
+				ThrottledRate:   util.Float64Ptr(3.0),
+				BlockHeadLag:    util.Float64Ptr(8.0),
+				FinalizationLag: util.Float64Ptr(2.0),
+				Misbehaviors:    util.Float64Ptr(5.0),
+				Overall:         util.Float64Ptr(1.0),
+			},
+			// For finalized/unknown: prioritize latency (block lag doesn't matter)
```
aramalipoor, line 2150: Maybe let's include:
> Even though "unknown" might include requests for tx hashes or block hashes of tip-of-chain range, they'll be retried due to retryEmpty logic if an upstream doesn't have the data. This is not ideal for numeric getBlockByNumber calls.

kasrakhosravi: adding it as a comment here or improving the logic based on this?

aramalipoor: I mean more like a comment, people (and agents lol) know why "unknown" is added here not above

PR #657, opened 2025-12-17
feat: add tag filtering support for QuickNode endpoints

a-monteiro:
## Motivation

QuickNode allows organizing endpoints using tags (e.g., "production", "staging", "us-east"). Previously, eRPC would fetch ALL endpoints for an API key. This adds filtering support to selectively use specific endpoints by tag IDs or labels.

## Usage

```yaml
upstreams:
  - id: quicknode-prod
    endpoint: quicknode://API_KEY?tagLabels=production,us-east
```

## Changes

- Add query parameter parsing for `tagIds` and `tagLabels` in URL format
- Add `QuicknodeFilterParams` struct for type-safe filter handling
- Format filters as comma-separated values per QuickNode Console API spec
- Add tests for URL parsing and filter extraction

## Reference

QuickNode Console API: https://www.quicknode.com/docs/console-api/endpoints/v0-endpoints

### DIFF
--- a/common/defaults.go
+++ b/common/defaults.go
@@ -1238,9 +1238,56 @@ func buildProviderSettings(vendorName string, endpoint *url.URL) (VendorSettings
 			"apiKey": endpoint.Host,
 		}, nil
 	case "quicknode", "evm+quicknode":
-		return VendorSettings{
+		settings := VendorSettings{
 			"apiKey": endpoint.Host,
-		}, nil
+		}
+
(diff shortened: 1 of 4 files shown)


aramalipoor:
Nice one @a-monteiro thanks for adding it

a-monteiro:
> Nice one @a-monteiro thanks for adding it

Thanks for the quick turn around, appreciate it. Will send a couple more your way.

PR #693, opened 2026-01-07
feat: include response body in error logs

kasrakhosravi:
> [!NOTE]
> Improves observability by logging the structured error response built for failures and enabling structured logging for JSON-RPC errors.
> 
> - Add `MarshalZerologObject` to `ErrJsonRpcExceptionExternal` and `HttpJsonRpcErrorResponse` for structured zerolog output
> - Refactor error handling: build error response first (`buildErrorResponseBody`), then log with both `request` and `response`; return the built response from `processErrorBody`
> - Update log lines to include `.Object("response", ...)` for client-side, cancellation, deadline, and generic error paths
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit bd1782b00d0761601317243bb561f972bbb3045c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/erpc/http_server.go
+++ b/erpc/http_server.go
@@ -1172,7 +1172,7 @@ func buildErrorResponseBody(nq *common.NormalizedRequest, err, origErr error, in
 		return serr
 	}
 
-	return &common.BaseError{
+	return common.BaseError{
 		Code:    "ErrUnknown",
 		Message: "unexpected server error",
 		Cause:   err,


erpc/http_server.go
```diff
@@ -1070,23 +1075,28 @@ func processErrorBody(logger *zerolog.Logger, startedAt *time.Time, nq *common.N
 			common.ErrCodeJsonRpcRequestUnmarshal,
 			common.ErrCodeProjectNotFound,
 		) {
-			logger.Debug().Err(err).Object("request", nq).Msgf("forward request errored with client-side exception")
+			logger.Debug().Err(err).Object("request", nq).Interface("response", resp).Msgf("forward request errored with client-side exception")
```
aramalipoor: We should probably use Object() and a carefully written marshal func for this to ensure proper bytes are written.

Also the memory usage is important to stay same, i.e. in all response types (getLogs multi-writer case, jsonRpcResponse objects, or raw body bytes, etc) it will cleanly write the bytes without impacting the actual response.

erpc/http_server.go
```diff
@@ -1054,8 +1054,31 @@ type HttpJsonRpcErrorResponse struct {
...
 
+func (r *HttpJsonRpcErrorResponse) MarshalZerologObject(e *zerolog.Event) {
+	if r == nil {
+		return
+	}
+	e.Str("jsonrpc", r.Jsonrpc)
+	e.Interface("id", r.Id)
+
+	if r.Error != nil {
+		if errBytes, err := common.SonicCfg.Marshal(r.Error); err == nil {
```
aramalipoor: Is there a way to avoid this Marshal?

erpc/http_server.go
```diff
@@ -1054,8 +1054,31 @@ type HttpJsonRpcErrorResponse struct {
...
+			if common.IsSemiValidJson(errBytes) {
```
aramalipoor: If we had to keep the above Marshal then why would we need this?

erpc/http_server.go
kasrakhosravi: the `r.Error` is a `map[string]interface{}` and needs to be converted to json bytes (either implicitly or explicitly)

I think we can't do that without using Interface{}. maybe `Interface` is fine here because the error object is not large (prob few bytes?) and we don't gain a ton of performance benefit by using sonic?

sth like this

```
func (r *HttpJsonRpcErrorResponse) MarshalZerologObject(e *zerolog.Event) {
    if r == nil {
        return
    }
    e.Str("jsonrpc", r.Jsonrpc)
    e.Interface("id", r.Id)
    if r.Error != nil {
        e.Interface("error", r.Error)
    }
}
```

erpc/http_server.go
kasrakhosravi: good point. redundant here in case of direct or indirect marshaling. prob only needed if you are not sure wether it is properly marshaled?

erpc/http_server.go
```diff
@@ -1070,23 +1085,42 @@ func processErrorBody(logger *zerolog.Logger, startedAt *time.Time, nq *common.N
...
+	switch v := resp.(type) {
+	case *HttpJsonRpcErrorResponse:
+		return e.Object("response", v)
+	case *common.BaseError:
+		return e.Object("response", v)
+	case common.StandardError:
+		return e.Object("response", v)
+	default:
+		// Fallback for unexpected types - uses reflection but shouldn't happen
+		return e.Interface("response", v)
```
aramalipoor: This fallback shouldn't happen normally, and probably we won't know the consequences of Interface() on an unknown resp. Any reason to keep it?

erpc/http_server.go
```diff
@@ -1070,23 +1085,42 @@ func processErrorBody(logger *zerolog.Logger, startedAt *time.Time, nq *common.N
 			common.ErrCodeJsonRpcRequestUnmarshal,
 			common.ErrCodeProjectNotFound,
 		) {
-			logger.Debug().Err(err).Object("request", nq).Msgf("forward request errored with client-side exception")
+			addResponseToLog(logger.Debug().Err(err).Object("request", nq), resp).Msgf("forward request errored with client-side exception")
```
aramalipoor: Can't we do this using the native zerolog Object() and chaining syntax, to keep things consistent and rely on zerolog optimizations?

erpc/http_server.go
kasrakhosravi: yeah probably we don't have any other error types. better be removed.

erpc/http_server.go
```diff
@@ -1139,13 +1159,13 @@ func processErrorBody(logger *zerolog.Logger, startedAt *time.Time, nq *common.N
...
-	if _, ok := err.(*common.BaseError); ok {
-		return err
+	if berr, ok := err.(*common.BaseError); ok {
+		return berr
 	} else if serr, ok := err.(common.StandardError); ok {
 		return serr
 	}
 
-	return common.BaseError{
+	return &common.BaseError{
```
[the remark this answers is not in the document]
kasrakhosravi: this is prob not a risk because it should only affects the fallback "ErrUnknown" case which shouldn't happen

erpc/http_server.go
```diff
@@ -1054,8 +1054,23 @@ type HttpJsonRpcErrorResponse struct {
...
 }
 
+func (r *HttpJsonRpcErrorResponse) MarshalZerologObject(e *zerolog.Event) {
+	if r == nil {
+		return
+	}
+	e.Str("jsonrpc", r.Jsonrpc)
+	e.Interface("id", r.Id)
+	if r.Error != nil {
+		e.Interface("error", r.Error)
```
aramalipoor: We should also use Object() here, for strict marshalling

erpc/http_server.go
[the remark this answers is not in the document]
aramalipoor: Any reason to return pointer type here? I feel it should be properly done to avoid hiding errors (seems affecting batch flow).

erpc/http_server.go
kasrakhosravi: makes sense

erpc/http_server.go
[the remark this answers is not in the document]
kasrakhosravi: the main reason was `common.BaseError{}` doesn't implement MarshalZerologObject (only pointer does)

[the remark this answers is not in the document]
kasrakhosravi: I might be mis-reading the batch flow but I think cracking to the default flow was already not good..with this we would just return a proper error (if it happens)

common/errors.go
```diff
@@ -2217,6 +2217,17 @@ func (e *ErrJsonRpcExceptionExternal) GetCause() error {
...
 }
 
+func (e *ErrJsonRpcExceptionExternal) MarshalZerologObject(ev *zerolog.Event) {
+	if e == nil {
+		return
+	}
+	ev.Int("code", e.Code)
+	ev.Str("message", e.Message)
+	if e.Data != nil {
+		ev.Interface("data", e.Data)
```
kasrakhosravi, line 2227: seems like we have to use the `Interface` here because usptreams can return string, object, null; so not possible to do strict object marshaling

erpc/http_server.go
```diff
@@ -1054,12 +1054,34 @@ type HttpJsonRpcErrorResponse struct {
...
+
+	// Build the response first, then log with it
+	resp := buildErrorResponseBody(nq, err, origErr, includeErrorDetails)
+
+	// Log the error with the response
 	if !common.IsNull(err) {
 		if nq != nil {
 			nq.RLock()
 		}
+		respMarshaler, _ := resp.(zerolog.LogObjectMarshaler)
```
aramalipoor, line 1084: Why do we need this casting? 🤔 Is it because because of response interface?

kasrakhosravi: exactly. `buildErrorResponseBody` returns `interface{}`, but `zerolog.Object()` requires `zerolog.LogObjectMarshaler`. The cast bridges that gap.

PR #704, opened 2026-01-13
feat: add ignoreNetworks to blocklist upstream for chains

msf:
Allows blocking a given upstream for specific networks via config. The exclusion is permanent and done on startup at registration time.

This feature adds a `ignoreNetworks` to the upstream configuration like so:
```
  upstreams:
    - id: upstream-a
      endpoint: https://a.example.com
      # Works for all chains, no restrictions

    - id: upstream-b
      endpoint: https://b.example.com
      ignoreNetworks:
        - "evm:12345"   # Never use B for 12345
```

Why: while using eRPC to loadbalance across multiple upstream providers we've identified bad data on a given chain coming from a specific provider.

Right now we do this in an imperfect way that causes some errors (due to requests being sent to a bad upstream provider) on startup time, like so:
```
networks:
  - architecture: evm
    evm:
      chainId: 12345  # not the real chainId
    selectionPolicy:
      resampleExcluded: false # don't resample the excluded upstream, we want it OUT
      evalFunction: |
        (upstreams) => {
          // Filter out XYZ due to blockTimestamp type mismatch (returns hex string instead of number)
          return upstreams.filter(u => u.config.group !== 'upstream-b');
        }
 ```


msf:
> @msf is this upstream a custom (non-http) like "alchemy://"?
> 

Yes, in reality, we're running with 4 non-http custom upstreams, plus some normal http ones and 70 networks.
The non-http, all of them have dozens of chains configured on their respective web-consoles.
We want to disable a specific chain on one of these upstreams.

I guess I didn't pay enough attention to the difference/overlap between upstreams and providers... I'm surprised I missed the fact that I should use providers and not upstreams. (Why these two different concepts?)


> If so the correct way to do this is to use "providers": https://docs.erpc.cloud/config/projects/providers#advanced-config
> 
> In eRPC each upstream in reality only works with 1 network, so it doesn't make sense to have a ignore network config, because it is equivalent of just commenting that upstream out (as in disabling/removing it). But I understand why this is confusing, the "alchemy://xxx" upstream is a syntax sugar, just to make it easier vs using "providers:" therefore not able to ignore a network in that model.
> 
> The "providers" should work here 🙂 Let me know if this helps in your use-case

Trying it today

PR #719, opened 2026-01-21
feat: make http client timeouts configurable

0x666c6f:
## Summary
- Add configurable timeout settings to `JsonRpcUpstreamConfig` and `ProxyPoolConfig`
- Previously hardcoded HTTP transport timeouts (30s ResponseHeaderTimeout, 60s client timeout) now configurable
- Defaults preserved for backward compatibility

## Problem
HTTP client timeouts were hardcoded in `clients/http_json_rpc_client.go` and `clients/proxy_pool_registry.go`:
- `ResponseHeaderTimeout: 30s` 
- `Timeout: 60s`

This caused issues with slow upstreams where requests would timeout at the transport layer (30s) before the configurable failsafe timeout (e.g., 120s) could apply.

## Solution
Added the following configurable fields to both `JsonRpcUpstreamConfig` and `ProxyPoolConfig`:
- `timeout` - Total request timeout (default: 60s)
- `responseHeaderTimeout` - Time to wait for response headers (default: 30s)
- `tlsHandshakeTimeout` - TLS handshake timeout (default: 10s)
- `idleConnTimeout` - Idle connection timeout (default: 90s)
- `expectContinueTimeout` - Expect-continue timeout (default: 1s)

## Example Configuration
```yaml
upstreams:
  - id: slow-upstream
    endpoint: https://rpc.example.com
    jsonRpc:
      timeout: "120s"
      responseHeaderTimeout: "90s"
```

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -793,6 +793,29 @@ type JsonRpcUpstreamConfig struct {
 	EnableGzip    *bool             `yaml:"enableGzip,omitempty" json:"enableGzip"`
 	Headers       map[string]string `yaml:"headers,omitempty" json:"headers"`
 	ProxyPool     string            `yaml:"proxyPool,omitempty" json:"proxyPool"`
+
+	// HTTP client timeout settings (optional, with sensible defaults)
+	// Timeout is the total time limit for a request including connection, headers, and body.
+	// Default: 60s
+	Timeout Duration `yaml:"timeout,omitempty" json:"timeout" tstype:"Duration"`
+
+	// ResponseHeaderTimeout specifies the time to wait for a server's response headers
+	// after fully writing the request. This does not include the time to read the response body.
+	// Default: 30s
+	ResponseHeaderTimeout Duration `yaml:"responseHeaderTimeout,omitempty" json:"responseHeaderTimeout" tstype:"Duration"`
+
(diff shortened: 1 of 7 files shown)


common/config.go
```diff
@@ -786,13 +786,149 @@ func (c *RateLimitAutoTuneConfig) Copy() *RateLimitAutoTuneConfig {
...
+		}
+	}
+	return ResolvedHTTPClientTimeouts{
+		Timeout:               t.Timeout.WithDefault(DefaultHTTPClientTimeout),
+		ResponseHeaderTimeout: t.ResponseHeaderTimeout.WithDefault(DefaultResponseHeaderTimeout),
+		TLSHandshakeTimeout:   t.TLSHandshakeTimeout.WithDefault(DefaultTLSHandshakeTimeout),
+		IdleConnTimeout:       t.IdleConnTimeout.WithDefault(DefaultIdleConnTimeout),
+		ExpectContinueTimeout: t.ExpectContinueTimeout.WithDefault(DefaultExpectContinueTimeout),
+	}
+}
```
aramalipoor, line 851: @0x666c6f can we move these to the defaults.go file which is responsible for all the defaulting logic/values?

PR #740, opened 2026-02-05
feat: add client-side load balancing for gRPC upstreams

hawkaa:
Adds opt-in client-side load balancing for gRPC connections. This is useful when connecting to Kubernetes headless services where you want requests distributed across multiple pods instead of sticking to one.

Uses gRPC's built-in DNS resolver and round_robin policy by default.

Example config:
```yaml
upstreams:
  - id: my-upstream
    endpoint: grpc://my-service.svc.cluster.local:50051
    grpc:
      loadBalancing:
        policy: round_robin
```

---

> [!NOTE]
> **Medium Risk**
> Changes connection targeting and request distribution semantics for all gRPC upstreams, which can affect routing, caching affinity, and failure behavior in production.
> 
> **Overview**
> Enables client-side load balancing for gRPC upstream connections by switching the dial target to `dns:///...` so gRPC resolves multiple A records (e.g., Kubernetes headless services).
> 
> Updates the default gRPC service config to use the `round_robin` load balancer while keeping existing retry and `waitForReady` behavior, so RPCs can be distributed across resolved endpoints.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 5a1a05c121ea6089fd3452d51d75e0c6e8c4bd8d. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -276,12 +276,11 @@ type ConnectorConfig struct {
 }
 
 type GrpcConnectorConfig struct {
-	Bootstrap     string                   `yaml:"bootstrap,omitempty" json:"bootstrap"`
-	Servers       []string                 `yaml:"servers,omitempty" json:"servers"`
-	Headers       map[string]string        `yaml:"headers,omitempty" json:"headers"`
-	GetTimeout Duration `yaml:"getTimeout,omitempty" json:"getTimeout" tstype:"Duration"`
+	Bootstrap  string            `yaml:"bootstrap,omitempty" json:"bootstrap"`
+	Servers    []string          `yaml:"servers,omitempty" json:"servers"`
+	Headers    map[string]string `yaml:"headers,omitempty" json:"headers"`
+	GetTimeout Duration          `yaml:"getTimeout,omitempty" json:"getTimeout" tstype:"Duration"`
 }
-
 
 type MemoryConnectorConfig struct {
 	MaxItems     int    `yaml:"maxItems" json:"maxItems"`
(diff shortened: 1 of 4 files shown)


common/config.go
```diff
@@ -276,10 +276,44 @@ type ConnectorConfig struct {
...
+	GetTimeout    Duration                 `yaml:"getTimeout,omitempty" json:"getTimeout" tstype:"Duration"`
+	LoadBalancing *GrpcLoadBalancingConfig `yaml:"loadBalancing,omitempty" json:"loadBalancing"`
+}
+
+// GrpcLoadBalancingConfig configures client-side load balancing for gRPC connections.
+// When enabled, gRPC will use its DNS resolver to discover all pod IPs (useful for
+// Kubernetes headless services) and distribute RPCs across them.
+type GrpcLoadBalancingConfig struct {
+	// Policy specifies the load balancing policy: "round_robin" (default) or "pick_first"
+	Policy string `yaml:"policy,omitempty" json:"policy"`
```
kasrakhosravi: if we are doing rotation (round robin or pick first) and there is a bad pod, will there be issues? do we need health-check?

kasrakhosravi: also do we need validation on the policy type? (e.g if the user types rund_robin)

hawkaa: >if we are doing rotation (round robin or pick first) and there is a bad pod, will there be issues? do we need health-check?
I'm actually not sure. The current gRPC implementation works very well and self heals if the pod stops responding. That is my experience. So I don't think this will be worse.

> also do we need validation on the policy type? (e.g if the user types rund_robin)
Yes, we should. I assume there is a pattern i should follow here.

PR #751, opened 2026-02-10
fix: distinguish pending vs mined txs for eth_getTransactionReceipt null responses

kasrakhosravi:
## Summary

**Problem:** `eth_getTransactionReceipt` returning null for pending transactions is treated as "missing data" (`ErrEndpointMissingData`) and retried across all upstreams, even though null is a valid response for pending/unknown txs. On Ethereum mainnet alone, this generates ~150-320K failed requests/day (~750K wasted upstream calls/day) for only ~2-3.5K unique tx hashes that get polled repeatedly.

**Root cause:** Three separate code paths react to empty responses, but only `HandleUpstreamPostForward` (EVM hooks) respects the `MarkEmptyAsErrorMethods` list. `MarkUpstreamCompleted` and failsafe retry unconditionally treat empty responses as errors, bypassing the configuration.

**Fix:**

1. **Dedicated post-forward hook** (`eth_getTransactionReceipt.go`): When a receipt is null, performs a lightweight side-call to `eth_getTransactionByHash` on the same upstream to determine if the tx is pending (null is valid, accept it) or mined-but-unindexed (retry on next upstream). This fires on <0.5% of total receipt traffic and actually reduces total upstream calls by preventing blind retries.

2. **`emptyAccepted` flag** on `NormalizedResponse`: Allows the hook to signal downstream code paths that a null result was intentionally accepted.

3. **Guards** in `MarkUpstreamCompleted` and failsafe retry: Respect the `emptyAccepted` flag to prevent overriding the hook's decision.

4. **Feature flag** (`receiptPendingCheck`): The hook is gated behind `EvmNetworkConfig.ReceiptPendingCheck` (default: disabled). Set to `true` in network config to enable. This allows safe rollout and testing before making it default.

## Config

```yaml
evm:
  receiptPendingCheck: true  # Enable the pending-check hook
```

## Test plan

- [x] Unit tests for pending, mined, unknown, and side-call failure scenarios
- [x] Integration tests for retry behavior with the hook enabled
- [x] Verified locally with real Ethereum mainnet RPCs
- [ ] Enable on mainnet/base via deployment config and monitor error rates

---

> [!NOTE]
> **Medium Risk**
> Changes core empty-result handling in request tracking and failsafe retry, which can alter retry/selection behavior across methods; gated behind `receiptPendingCheck`, but mistakes could reduce necessary retries or mask upstream indexing issues.
> 
> **Overview**
> Adds an **optional** EVM post-forward validation for `eth_getTransactionReceipt` that, when the receipt result is `null`, side-calls `eth_getTransactionByHash` on the same upstream to classify the tx as *pending/unknown* (accept empty) vs *mined* (raise `ErrEndpointMissingData` to trigger retry).
> 
> Introduces an `emptyAccepted` flag on `NormalizedResponse` plus a new network config `evm.receiptPendingCheck` (default off) so downstream selection (`MarkUpstreamCompleted`) and network failsafe retry logic do **not** treat intentionally-accepted empty results as missing data. Updates unit/integration tests and the fake upstream to support method-specific forwarding in these scenarios.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 899457ce8d7d6c5bb2182bb794a157bf6be416ac. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/architecture/evm/eth_getTransactionReceipt.go
+++ b/architecture/evm/eth_getTransactionReceipt.go
@@ -22,6 +22,11 @@ func upstreamPostForward_eth_getTransactionReceipt(
 		return rs, re
 	}
 
+	// Feature flag: skip the side-call unless explicitly enabled
+	if nCfg := n.Config(); nCfg == nil || nCfg.Evm == nil || nCfg.Evm.ReceiptPendingCheck == nil || !*nCfg.Evm.ReceiptPendingCheck {
+		return rs, re
+	}
+
 	if rq != nil {
 		if rd := rq.Directives(); rd != nil && !rd.RetryEmpty {
 			return rs, re
--- a/architecture/evm/hooks_test.go
+++ b/architecture/evm/hooks_test.go
@@ -7,6 +7,7 @@ import (
 	"testing"
 
 	"github.com/erpc/erpc/common"
+	"github.com/erpc/erpc/util"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
 )
@@ -201,6 +202,7 @@ func TestUpstreamPostForward_UnexpectedEmpty_EmptyConfig(t *testing.T) {
 
 func TestUpstreamPostForward_TransactionReceipt_PendingTx(t *testing.T) {
 	network := newTestNetworkWithMarkEmptyMethods(common.DefaultMarkEmptyAsErrorMethods)
+	network.cfg.Evm.ReceiptPendingCheck = util.BoolPtr(true)
 
 	fakeUp := common.NewFakeUpstream("rpc1").(*common.FakeUpstream)
 	fakeUp.ForwardFunc = func(ctx context.Context, nq *common.NormalizedRequest, _ bool) (*common.NormalizedResponse, error) {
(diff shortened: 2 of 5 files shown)


architecture/evm/eth_getTransactionReceipt.go
```diff
@@ -0,0 +1,91 @@
...
+		return rs, re
+	}
+
+	// Feature flag: skip the side-call unless explicitly enabled
+	if nCfg := n.Config(); nCfg == nil || nCfg.Evm == nil || nCfg.Evm.ReceiptPendingCheck == nil || !*nCfg.Evm.ReceiptPendingCheck {
+		return rs, re
+	}
+
+	if rq != nil {
+		if rd := rq.Directives(); rd != nil && !rd.RetryEmpty {
```
kasrakhosravi, line 31: @aramalipoor does this make sense?

aramalipoor: It depends on the behavior we expect here, what is the end goal? 🤔

PR #774, opened 2026-02-19
fix: make ErrNetworkNotSupported retryable instead of fatal

kasrakhosravi:
## Summary

- Remove the `NewTaskFatal()` wrapper from `ErrNetworkNotSupported` in `PrepareUpstreamsForNetwork` (two call sites in `upstream/registry.go`)
- Networks that fail initialization are now `TaskFailed` (retryable) instead of `TaskFatal` (permanent death)
- The auto-retry loop will re-attempt with exponential backoff (3s–130s, 1.5x factor), allowing recovery when upstreams come back online

## Problem

When all upstreams for a chain were temporarily unavailable during eRPC startup, `PrepareUpstreamsForNetwork` marked the network as `TaskFatal`. This was a **permanent terminal state** with no recovery path — the network was dead until a pod restart.

## Fix

Remove `NewTaskFatal()` wrapper → error becomes `TaskFailed` → eligible for auto-retry → networks self-heal when upstreams recover.

## Test plan

- [x] `TestNetworksBootstrap_SlowProviderUpstreams_InitializeThenServe` — passes
- [x] `TestNetworksBootstrap_UnsupportedNetwork_FatalFast` — passes (still returns `ErrNetworkNotSupported`)
- [x] `TestNetworksBootstrap_ProviderInitializing_503Retry` — passes
- [x] `go build ./...` — clean

---

> [!NOTE]
> **Medium Risk**
> Changes error classification for network initialization (fatal vs retryable), which can alter startup/steady-state recovery behavior and client-visible errors, but is localized and covered by updated tests.
> 
> **Overview**
> Network bootstrap no longer marks `ErrNetworkNotSupported` as *fatal*: `PrepareUpstreamsForNetwork` now returns the raw error so the auto-retry loop can re-attempt initialization when upstreams recover.
> 
> HTTP transport error mapping is updated so `ErrNetworkNotSupported` is treated like other missing resources (404), and the related server test expectation is adjusted from `ErrNetworkNotFound` to `ErrNetworkNotSupported`. `.gitignore` is also updated to ignore `temp/**`, `erpc-bin`, and `erpc.yml`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit e94d800b46f1ce9719370ca903e7c2860a0f590c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/erpc/http_server.go
+++ b/erpc/http_server.go
@@ -1028,7 +1028,7 @@ func determineResponseStatusCode(res interface{}) int {
 	case common.HasErrorCode(err, common.ErrCodeAuthUnauthorized, common.ErrCodeEndpointUnauthorized):
 		return http.StatusUnauthorized
 	// 404 Not Found - resource not found
-	case common.HasErrorCode(err, common.ErrCodeProjectNotFound, common.ErrCodeNetworkNotFound):
+	case common.HasErrorCode(err, common.ErrCodeProjectNotFound, common.ErrCodeNetworkNotFound, common.ErrCodeNetworkNotSupported):
 		return http.StatusNotFound
 	// 413 Request Entity Too Large
 	case common.HasErrorCode(err, common.ErrCodeEndpointRequestTooLarge):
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -20,4 +20,5 @@ node_modules
 .next
 .env
 bin/**
-temp/**
\ No newline at end of file
+temp/**erpc-bin
```
aramalipoor: Shouldn't this be on 2 lines?

PR #788, opened 2026-02-25
feat: forward headers, server-level `allowMethods` and `ignoreMethods` config

sistemd:
- Add a `forwardHeaders` server-level config to determine which headers are forwarded from requests to backends.
- Add `allowMethods` and `ignoreMethods` server-level config to filter which methods can be handled by the server, to ultimately reduce logging noise.

### DIFF
--- a/erpc/http_server.go
+++ b/erpc/http_server.go
@@ -443,9 +443,19 @@ func (s *HttpServer) createRequestHandler() http.Handler {
 				}
 
 				if project != nil {
-					for _, key := range project.Config.ForwardHeaders {
-						if value := headers.Get(key); value != "" {
-							nq.ForwardHeaders.Add(key, value)
+					for _, matchKey := range project.Config.ForwardHeaders {
+						for key, values := range headers {
+							matches, err := common.WildcardMatch(matchKey, key)
+							if err != nil {
+								responses[index] = processErrorBody(&lg, &startedAt, nq, err, &common.TRUE)
+								common.EndRequestSpan(requestCtx, nil, responses[index])
+								return
+							}
+							if matches {
+								for _, value := range values {
+									nq.ForwardHeaders.Add(matchKey, value)
+								}
+							}
 						}
 					}
 				}


aramalipoor:
@sistemd thanks for the PR, what do you think about having this on Project-level? To keep the server component/config focused on transport concepts and logic, and for example "frontend" projects might allow diff methods vs "indexer" project.

erpc/http_server.go
```diff
@@ -440,9 +442,67 @@ func (s *HttpServer) createRequestHandler() http.Handler {
 					return
 				}
 
+				for _, key := range s.serverCfg.ForwardHeaders {
+					if value := headers.Get(key); value != "" {
+						nq.ForwardHeaders.Add(key, value)
+					}
```
aramalipoor: With regards to memory management, we might be better off passing the headers map on the request object? I think other usages of header exists this can help both.

And then we do the matching when constructing the upstream request.

erpc/http_server.go
```diff
@@ -440,9 +442,67 @@ func (s *HttpServer) createRequestHandler() http.Handler {
 					return
 				}
 
+				for _, key := range s.serverCfg.ForwardHeaders {
+					if value := headers.Get(key); value != "" {
```
aramalipoor: For key matching how about we use WildCard matching? One could add "x-amz-*" for better DX.

sistemd:
> @sistemd thanks for the PR, what do you think about having this on Project-level? To keep the server component/config focused on transport concepts and logic, and for example "frontend" projects might allow diff methods vs "indexer" project.

https://github.com/erpc/erpc/pull/788/commits/d2c91db3b9397654bc5d4bca7ab53e509edf0cda

erpc/http_server.go
sistemd: https://github.com/erpc/erpc/pull/788/commits/cdb9922785237a097501cdd193486e0b1c8ebb90

erpc/http_server.go
sistemd: Sorry, I don't understand what you would like me to do here or how it improves things. Why does it matter where the matching happens?

PR #789, opened 2026-02-25
enh: reduce `user` cardinality for IP CIDR authorization strategies

TPXP:
On private networks with many clients, allowing large CIDR ranges can lead to many IPs being used as metric labels, causing significant cardinality in metrics like `erpc_network_evm_block_range_requested_total`, `erpc_network_failed_request_total`, `erpc_network_multiplexed_request_total`, `erpc_network_request_duration_seconds_bucket`.

This PR adds a configuration option `ipCidrLabelMode` on network auth strategies that changes the `user` label to the IP cidr instead of the IP address when set to `cidr`. Default behaviour is unchanged

📖  We may want to update docs, I'm not sure which page to update

### DIFF
--- a/auth/strategy_network.go
+++ b/auth/strategy_network.go
@@ -12,6 +12,7 @@ type NetworkStrategy struct {
 	cfg          *common.NetworkStrategyConfig
 	allowedIPs   []*net.IP
 	allowedCIDRs []*net.IPNet
+	ipAsUser     bool
 }
 
 var _ AuthStrategy = &NetworkStrategy{}
@@ -19,6 +20,7 @@ var _ AuthStrategy = &NetworkStrategy{}
 func NewNetworkStrategy(cfg *common.NetworkStrategyConfig) (*NetworkStrategy, error) {
 	s := &NetworkStrategy{
 		cfg: cfg,
+		ipAsUser: cfg.IPAsUser,
 	}
 
 	// Parse and store allowed IPs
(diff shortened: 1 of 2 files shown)


common/config.go
```diff
@@ -1748,13 +1748,20 @@ type SiweStrategyConfig struct {
...
+)
+
 type NetworkStrategyConfig struct {
 	AllowedIPs     []string `yaml:"allowedIPs" json:"allowedIPs"`
 	AllowedCIDRs   []string `yaml:"allowedCIDRs" json:"allowedCIDRs"`
 	AllowLocalhost bool     `yaml:"allowLocalhost" json:"allowLocalhost"`
 	TrustedProxies []string `yaml:"trustedProxies" json:"trustedProxies"`
 	// RateLimitBudget, if set, is applied to the authenticated user (client IP)
 	RateLimitBudget string `yaml:"rateLimitBudget,omitempty" json:"rateLimitBudget,omitempty"`
+	IPCidrLabelMode IPCidrLabelMode `yaml:"ipCidrLabelMode,omitempty" json:"ipCidrLabelMode,omitempty"`
```
aramalipoor:
```suggestion
	IPAsUser bool `yaml:"ipAsUser,omitempty" json:"ipAsUser,omitempty"`
```
How about this? False by default and if anyone wants they can enable it

auth/strategy_network.go
```diff
@@ -9,16 +9,18 @@ import (
 )
 
 type NetworkStrategy struct {
-	cfg          *common.NetworkStrategyConfig
-	allowedIPs   []*net.IP
-	allowedCIDRs []*net.IPNet
+	cfg                 *common.NetworkStrategyConfig
+	allowedIPs          []*net.IP
+	allowedCIDRs        []*net.IPNet
+	IPCidrLabelOnlyCidr bool
```
aramalipoor:
```suggestion
	ipAsUser bool
```
Besides the name comment, it should be defined as private field

common/config.go
TPXP: Sure, that works too :)

auth/strategy_network.go
TPXP: Oh, good catch - forgot about this

PR #790, opened 2026-02-25
feat: dynamic block time EMA for debounce

kasrakhosravi:
## Summary

Replace hardcoded `KnownBlockTimes` map with dynamic block time learning via EMA on `block.timestamp`. The EMA drives the `statePollerDebounce` interval so each chain auto-tunes its debounce based on actual block production rate.

## What changed

- **EMA block time estimation** (`health/tracker.go`): Feeds `block.timestamp` differences (not `time.Now()`) into an EMA, normalized by block gap. For fast chains where consecutive blocks share the same integer-second timestamp, samples are skipped until the timestamp advances — blockGap normalization recovers sub-second precision. Sanity bounds (10ms–120s) with EMA reset on suppression to prevent delayed spikes.
- **Dynamic debounce** (`evm_state_poller.go`): `resolveDebounce` returns `blockTime × 0.7` (configurable via `dynamicBlockTimeDebounceMultiplier`). Falls back to `fallbackStatePollerDebounce` (5s) then 1s floor. Zero/negative multiplier rejected.
- **Poll timer unchanged**: Uses configured `statePollerInterval` (fixed ticker). Dynamic block time only controls debounce, not the timer.
- **Syncing check timeout**: Capped at 5s to prevent blocking the poll cycle.
- **Healthcheck**: `blockTimeMs` exposed per network in verbose and networks mode.
- **Config**: `statePollerDebounce` default changed from 5s to dynamic. New `dynamicBlockTimeDebounceMultiplier` (default 0.7) and `fallbackStatePollerDebounce` (default 5s) options.
- **Cleanup**: Removed `KnownBlockTimes`, `block_times.go`, `inferDebounceIntervalFromBlockTime`.

## Metrics

- `erpc_network_dynamic_block_time_milliseconds` — EMA-estimated block time per network

## Test plan

- 18 unit tests for EMA (convergence, fast chains, chain halts, clock skew, block gap normalization)
- Deployed to goldsky-dev and prod-us-west-2 (mainnet + arbitrum) with dynamic debounce

### DIFF
--- a/architecture/evm/evm_state_poller.go
+++ b/architecture/evm/evm_state_poller.go
@@ -366,7 +366,7 @@ func (e *EvmStatePoller) resolveDebounce(cfg *common.EvmNetworkConfig) time.Dura
 		// debounce expires before the next block is expected. This prefers freshness
 		// over saving RPC calls — EMA smooths long-term, multiplier covers the tail.
 		mult := common.DefaultDynamicBlockTimeDebounceMultiplier
-		if cfg != nil && cfg.DynamicBlockTimeDebounceMultiplier != nil {
+		if cfg != nil && cfg.DynamicBlockTimeDebounceMultiplier != nil && *cfg.DynamicBlockTimeDebounceMultiplier > 0 {
 			mult = *cfg.DynamicBlockTimeDebounceMultiplier
 		}
 		return time.Duration(float64(blockTime) * mult)


erpc/projects_registry.go
```diff
@@ -90,6 +90,7 @@ func (r *ProjectsRegistry) RegisterProject(prjCfg *common.ProjectConfig) (*Prepa
 
 	wsDuration := prjCfg.ScoreMetricsWindowSize.Duration()
 	metricsTracker := health.NewTracker(&lg, prjCfg.Id, wsDuration)
+	metricsTracker.SetKnownEvmBlockTimes(evm.KnownBlockTimes)
```
aramalipoor: Why "evm.KnownBlockTimes" is not used directly in metrics tracker?

kasrakhosravi: I started with that first actually, but seems like we are already importing `health ` in `architecture/evm`; so would cause a circular dependency?

kasrakhosravi: https://github.com/erpc/erpc/blob/0bad6f82bf05557601500d2a032795fd0ac4b53d/architecture/evm/evm_state_poller.go#L13

aramalipoor: Ah I see, yeah that's annoying. We need to find some time to do a clean up in terms of boundaries of these 3 areas (evm, erpc, health, upstream)

kasrakhosravi:
> I don't see where is this new data used? The "infer" function initially is just on startup...
> 
> This component is useful in places like "retry" delay, or for example dynamic debounce (debounceInterval) in evm state poller, etc.
> 
> Also do we have a "min samples" concept to e.g. only provide the blockTime if we have enough samples 5 or 10 etc.

1- for this PR, I tried to only focus on `debounce interval`. it is used in `evm_state_poller` here

https://github.com/erpc/erpc/pull/790/changes#diff-ca5825fe2addf77a207fed9f0c8effd4711fa72df181df953b5c3569daa3bb50R158

2- I actually started with a configurable minSamples, but ended up dropping it:

* The first sample could be noisy and replace the known default, but the odds are slim — and we already reject anything below 50ms or above 120s, so truly bad values never enter the EMA.
* The known block times already cover cold start, so there's a reasonable fallback in place until samples arrive.
* Exposing minSamples as a config felt risky — a high value could delay the EMA from kicking in and leave the poller on stale defaults for too long.
* Overall, fewer knobs to tune and less code to reason about.

@aramalipoor wdyt?

architecture/evm/evm_state_poller.go
```diff
@@ -1434,15 +1432,10 @@ func (e *EvmStatePoller) binarySearchEarliest(ctx context.Context, probe common.
 	return l, nil
 }
 
-func (e *EvmStatePoller) inferDebounceIntervalFromBlockTime(chainId int64) {
-	if d, ok := KnownBlockTimes[chainId]; ok {
-		// Anything lower than 1 second has a chance of causing a thundering herd (e.g. a high RPS for a method like getLogs).
-		// If users truly want to have a smaller value they can directly set the debounce interval
-		// either on network config or upstream config.
```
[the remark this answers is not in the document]
kasrakhosravi: @aramalipoor this one I am abit worried about; but bringing back the logic of ignoring values below `1s` kind of defeats the purpose of our dynamic value?

kasrakhosravi:
> > debounce interval. it is used in evm_state_poller here
> 
> 1. This is the bootstrap, and this "infer" func only runs initially (i.e. when we don't have any samples + won't update debounce as new samples come in), or maybe I'm missing something?
> 2. The min sample is less about anomalies, and it's more about being closer to reality (e.g. on Polygon ~2.3s could be avg. but some blocks might be slower or faster, so over 10 blocks we have more accurate value). By the way on the usage (e.g. on retry delays for unavailable block, or evm poller debounce, etc) we might need to have a multipler, because you sometimes need half the block time or double the block time depending on your goal. Not sure what's the best way to achieve this 🤔
> 
> Re known block times, I'm leaning towards completely removing it, because then we'll have 2 places for block times, and architecturally the known values could hide some weakness in our ema calculation, where we don't see anything bad for ethereum/arbitrum, but for other chains it might have problems.
> 
> Re "leave the poller on stale defaults for too long" imo we need to make our design so that it is not at risk in such situation, it is OK to be less efficient though, e.g. too many redundant requests, but not risk of data integrity if values are not coming in, basically we need to design for the scenario where for a chain "known block time" doesn't exist + "enough ema samples have not arrived yet", still the service is running with 0 integrity issues, just bit less efficient.
> 
> What do you think?

1- excellent point. totally forgot the previous flow was based on a static value and not dynamic. performacne wise it should be very cheap to read the debounce value on new polls

2- def. enough samples make it more reliable. I am thinking wether we should hard-code this or allow users to configure?

3- wrt "multiplier", a naive approach i am thinking is for block time calculations to remain pure and for the use-cases like retry, we use the multiplier directly in their own logic.

4- agree with dropping known block times

5- we had a `5s` default for debounce value. I decreased that to `0`; so dynamic value actually can kick in. in case there is no dynamic block time yet (0), we have a pragmatic 1s debounce which i think makes it pretty saafe

architecture/evm/evm_state_poller.go
```diff
@@ -364,18 +350,35 @@ func (e *EvmStatePoller) Poll(ctx context.Context) error {
...
 
+// resolveDebounce returns the debounce interval for poll methods.
+//
+//	user config → EMA block time → network FallbackStatePollerDebounce → 1s hard floor
+func (e *EvmStatePoller) resolveDebounce(cfg *common.EvmNetworkConfig) time.Duration {
+	if dbi := e.debounceInterval; dbi != 0 {
+		return dbi
+	}
+	if dbi := e.tracker.GetNetworkBlockTime(e.upstream.NetworkId()); dbi != 0 {
+		return dbi
```
aramalipoor: Should we do a blockTime / 2? e.g. for eth mainnet with 12 second block time, a debounce of 6s (so if block is updated sooner than 12s we catch it).

Not sure if this is best solution... I think ideally the debounce is "time of last block in MS + blocktime - 1ms" or something like that. Is that easy to do with current flows?

health/tracker.go
```diff
@@ -801,6 +811,87 @@ func (t *Tracker) SetLatestBlockNumberForNetwork(network string, blockNumber int
 	ntwMeta.evmLatestBlockNumber.Store(blockNumber)
 }
 
+// ------------------------------------
+// Dynamic Block Time (EMA)
+// ------------------------------------
+
+const blockTimeEMAAlpha = 0.1
+const blockTimeMinSamples = 10 // minimum samples before the EMA is considered stable
```
aramalipoor: Maybe a 4 block is good enough min sample? for eth mainnet 10 * 12 seconds is like 2 minutes with 1s debounce, not worst thing but still a bit inefficient.

monitoring/Dockerfile
```diff
@@ -1,7 +1,7 @@
 FROM prom/prometheus:v3.9.1@sha256:1f0f50f06acaceb0f5670d2c8a658a599affe7b0d8e78b898c1035653849a702 as prometheus
 FROM grafana/grafana:12.3.3@sha256:9e1e77ade304069aee3196e9a4f210830e96e80ce9a2640891eccc324b152faf as grafana
 
-FROM ubuntu:22.10@sha256:e322f4808315c387868a9135beeb11435b5b83130a8599fd7d0014452c34f489
+FROM ubuntu:24.04
```
aramalipoor: I think dependabot complains here unless it's tagged with exact image hash

telemetry/metrics.go
```diff
@@ -88,6 +88,12 @@ var (
 		Help:      "Distance in seconds between the latest block timestamp and current time for a network.",
 	}, []string{"project", "network", "origin"})
 
+	MetricNetworkDynamicBlockTime = promauto.NewGaugeVec(prometheus.GaugeOpts{
+		Namespace: "erpc",
+		Name:      "network_dynamic_block_time_seconds",
```
aramalipoor: Perhaps to no lose granularity (both in metrics and when calculating values off of the dynamic block time) we should store/save/emit this value in milliseconds (not just here in metics but also when we store it or compare to block timestamp values).

On fast chains like Arbitrum we need to be aware of such granularity, e.g. a rounding down/up would mean 4 blocks behind/ahead.

architecture/evm/evm_state_poller.go
kasrakhosravi: /2 prob is not gonna work as you mentioned: if the halved debounce triggers a poll before the block arrives, TryUpdateIfStale resets its internal clock on every fetch (even if nothing changed), which can actually push the useful poll further out.

not sure if. correct, but I added a `shouldSkipTick()` that does `now < lastBlockOnChainTimestamp + blockTime`, skips the tick if a new block can't have arrived yet

telemetry/metrics.go
kasrakhosravi: makes total sense

architecture/evm/evm_state_poller.go
kasrakhosravi: https://github.com/erpc/erpc/pull/790/changes#diff-ca5825fe2addf77a207fed9f0c8effd4711fa72df181df953b5c3569daa3bb50R360

health/tracker.go
```diff
@@ -48,6 +48,14 @@ type NetworkMetadata struct {
...
 	evmLatestBlockTimestamp atomic.Int64
 	evmFinalizedBlockNumber atomic.Int64
+
+	// Dynamic block time EMA. Callers should use GetNetworkBlockTime which
+	// only returns the EMA once enough samples have been collected.
+	evmBlockTime             atomic.Int64 // current EMA value in nanoseconds
+	evmBlockTimeSamples      atomic.Int64 // number of samples folded into the EMA
+	evmBlockTimeMu           sync.Mutex   // protects ref fields below + EMA read-modify-write
+	evmBlockTimeRefBlock     int64        // block number from last valid (block, timestamp) pair
+	evmBlockTimeRefTimestamp int64        // timestamp from that same block
```
aramalipoor: Should we ensure if this and other block timestamp variables are millisecond-precision for fast chains like arb/monad/hyper?

As far I remember block.timestamp would give us seconds-precision, which would make it tricky for us!

kasrakhosravi: I think `block.timestamp` (source of evmBlockTimeRefTimestamp) is seconds, even on Arb/Monad — storing as ms would just be *1000, no actual gain right?. 

The EMA itself is already nanosecond internally, and per-block normalization handles the rest

architecture/evm/evm_state_poller.go
aramalipoor: Love the implementation, this is most efficient and MUCH better than fixed debounce! It's a real debounce, especially now with blockTime knowledge. Just need to ensure granularity is good enough.

health/tracker.go
aramalipoor: Yeah I think for this we need to find a solid solution otherwise in best-case we'll round the debounce to 1s effectively, which means the prev fixed values would be a better alternative 🤔 

Do these chains provide a millisecond field? A smart thing we can do is to gather 8 samples, and do an algo where if 4 blocks are in the same second we do 1s / 4 -> 250ms 😄

architecture/evm/evm_state_poller.go
```diff
@@ -171,6 +168,9 @@ func (e *EvmStatePoller) Bootstrap(ctx context.Context) error {
 				e.logger.Debug().Msg("shutting down evm state poller due to app context interruption")
 				return
 			case <-ticker.C:
+				if e.shouldSkipTick() {
+					continue
+				}
```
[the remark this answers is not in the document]
aramalipoor: Probably we can ignore finalized block here, or if necessary I think in real scenarios it can tolerate to be advanced same as latest block for now, not ideal but acceptable imo.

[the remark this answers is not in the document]
kasrakhosravi: hmm yeah I agree

health/tracker.go
kasrakhosravi: I wanna make sure I understand your comment correctly

* so right now we are doing exactly the formula you mentioned (timeDiff / blockDiff ) right? https://github.com/erpc/erpc/pull/790/changes#diff-ff58fa315bd2c355bfb72662561d04179bf4745f23f18f4099c9778d24e9ba92R859

* maybe you meant the 1s default we had for EMA values in resolve debounce which basically breaks fast chains. this seems like a better approach? https://github.com/erpc/erpc/pull/790/changes#diff-ca5825fe2addf77a207fed9f0c8effd4711fa72df181df953b5c3569daa3bb50R388

aramalipoor: 1. I think that formula is doing a different thing, it's basically a robustness if we missed few blocks on polling. But what i'm proposing above is to have "all" samples in the memory as array, and then see how many blocks are in the same seconds... so none of the blocks in the middle must be missed (an edge case that must be covered)... This current formula will give us 1s if prev store block is N and current stored block is N+1 (no gap) because timeDiff would be 0 in such scenario (both have same second-granularity... although there's a if that skips calc entirely if blocks are in the same second which is worse)
2. We can make the default lowest block time we've seen in the wild, or just 100ms, which means we'll be safe and almost all chains are slower than 100ms so we'd never serve stale data, BUT might send too many unnecessary requests, until we have enough samples.

architecture/evm/evm_state_poller.go
```diff
@@ -364,18 +353,59 @@ func (e *EvmStatePoller) Poll(ctx context.Context) error {
...
+	lastDetectedMs := e.tracker.GetLastNewBlockDetectedAt(e.upstream.NetworkId())
+	if lastDetectedMs <= 0 {
+		return defaultInterval
+	}
+	nextExpected := time.UnixMilli(lastDetectedMs).Add(blockTime)
+	delay := time.Until(nextExpected)
+	if delay < defaultInterval {
+		return defaultInterval
+	}
+	return delay
```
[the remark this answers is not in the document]
aramalipoor: @kasrakhosravi does this make sense?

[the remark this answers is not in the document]
kasrakhosravi: it does. the guard was backwards — defaultInterval (30s) should be the cap, not the floor

architecture/evm/evm_state_poller.go
```diff
@@ -364,18 +353,69 @@ func (e *EvmStatePoller) Poll(ctx context.Context) error {
...
+	// for low integrity lag; the per-upstream debounce prevents thundering herd.
+	if delay < 50*time.Millisecond {
+		return 50 * time.Millisecond
+	}
+	// Cap: never poll slower than the configured interval
+	if delay > defaultInterval {
+		return defaultInterval
+	}
+	return delay
+}
```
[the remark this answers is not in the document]
aramalipoor: I was worried about something along the same line, maybe an additional % buffer or something like that can help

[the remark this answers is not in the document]
kasrakhosravi: I think it is a valid concern; specially if the delay is long or halt

replaced the 50ms floor with gradual backoff: blockTime/10 + overdue/10, capped at blockTime/2. wdyt?

e.g mainnet goes from ~240 polls to ~8 over a full block period of lateness.

PR #799, opened 2026-03-07
feat(svm): multi-chain Solana support

kasrakhosravi:
## Summary

Adds **multi-chain SVM (Solana) support** to eRPC as a first-class network architecture alongside EVM. Solana and SVM-compatible forks (Fogo, Eclipse, …) run side by side, addressed as `svm:<cluster>` or `svm:<chain>:<cluster>`. EVM behavior is unchanged; the request pipeline is generalized behind a per-architecture handler seam.

This branch is the implementation from **[andreclaro/erpc#5](https://github.com/andreclaro/erpc/pull/5)** by @andreclaro, **reconciled onto current `main`**, then put through a full review pass whose findings are fixed here (see [Review hardening round](#review-hardening-round)). Design docs: [`DESIGN-MULTI-CHAIN-SOLANA_v0.5.md`](https://github.com/andreclaro/erpc/blob/solana-rebase/DESIGN-MULTI-CHAIN-SOLANA_v0.5.md).

## How it works

- **`ArchitectureHandler` seam** — per-architecture pre/post-forward hooks so the generic pipeline never switches on architecture inline; EVM logic moves behind the same interface.
- **`architecture/svm`** — per-upstream slot poller (`getSlot` processed+finalized, `getHealth`, `getMaxShredInsertSlot`), commitment-aware finality, a dedicated JSON-RPC cache, an error normalizer, and genesis-hash cluster validation.
- **Commitment-aware caching** — default-commitment injection happens at the **project pre-forward, before the cache read**, so GET/SET key on identical params (fixes a permanent cache miss). Only reads pinned to an immutable point (`getBlock(slot)`, `getTransaction(sig)`) at a finalized commitment are cached as finalized; moving-head state reads are not, whatever their commitment. A shape-aware per-method options table injects the commitment where each method actually parses it.
- **Multi-chain identity** — `util.SvmNetworkId(chain, cluster)`; nested per-(chain,cluster) genesis table; fail-closed genesis validation for known clusters, `checkGenesisHash` opt-in for forks. Routing handles `svm:<chain>:<cluster>` across URL path, body `networkId`, and aliases.
- **Write-path safety** — `sendTransaction`, `sendRawTransaction` and `requestAirdrop` are never auto-dispatched to a second upstream; write-path commitment is normalized to the correct per-method field (`preflightCommitment` vs `commitment`).
- **Consensus slot-lag pre-filter** — for SVM networks with an active consensus policy on a finalized-commitment request, upstreams trailing the pool leader by more than `maxFinalizedSlotLag` are pruned before consensus runs.

## URL form

```
https://<host>/<project>/svm/mainnet-beta
https://<host>/<project>/svm/fogo:mainnet      # svm:<chain>:<cluster>
```

```bash
curl https://<host>/main/svm/mainnet-beta -X POST -H 'content-type: application/json' \
  -d '{"jsonrpc":"2.0","id":1,"method":"getBalance","params":["<pubkey>",{"commitment":"finalized"}]}'
```

## Review hardening round

A full-persona review of this branch (correctness, reliability, API contract, security, performance, testing, maintainability, standards, adversarial) produced 11 P1-class candidates. Each was re-validated by an independent second reviewer against the exact head before being accepted; **two were rejected or demoted on that gate** and are noted at the bottom. Eight P1 and nine P2 defects are fixed here.

Several of these were only reachable because the branch had never run anything but `getSlot`/`getBlock` in anger. They are Solana-semantics bugs, not porting slips.

### P1 — data correctness and wire contract

| # | Defect | Why it mattered |
|---|---|---|
| 1 | SVM cache aliased case-differing base58 keys | `req.CacheHash` lowercases every string param — correct EVM hex normalization, catastrophic for case-sensitive base58. Two distinct valid accounts shared one cache entry and could be served **each other's data**. SVM now derives its own case-preserving, type- and length-delimited key. |
| 2 | `commitment: finalized` state reads cached forever | Solana `finalized` is the latest **rooted** slot, a head that moves every ~400ms — not an immutability guarantee. Combined with `DataFinalityStateFinalized` being the zero value a minimal policy matches, an unset TTL meaning no-expiry, and a `*` slot partition, `getBalance` was pinned at its first observed value permanently. |
| 3 | `getBlockHeight` rewritten to a slot number | Block height trails the slot by every skipped slot (tens of millions on mainnet-beta). The response was corrected up to the slot tip, so `getBlockHeight` vs `lastValidBlockHeight` made every transaction look permanently expired. |
| 4 | `minContextSlot` treated as a history bound | It is a node-freshness floor — the minimum bank slot at which a request may be *evaluated*. The guard rejected valid `limit:1` queries with `-32602` and told callers to paginate, which cannot satisfy it. Guard and its config key deleted. |
| 5 | TypeScript package could not express an SVM config | Generated interfaces existed but the hand-maintained public unions were still `"evm"`-only and the SVM types were never exported from the package root, so `createConfig({ architecture: "svm" })` did not compile. |
| 6 | Native Solana error codes and `error.data` destroyed | `-32002 SendTransactionPreflightFailure` became `-32600` with its `RpcSimulateTransactionResult` dropped, breaking `@solana/web3.js` `SendTransactionError` and exact-code dispatch. |
| 7 | Shred-insert lag inverted; health never reached routing | Lag was computed `processed - watermark` and clamped at zero, so the silent-stale detector could never fire — and `IsHealthy()` had no production consumer at all. |
| 8 | Mixed missing-data causes produced a random wire code | `-32004` (not yet available) vs `-32009` (skipped forever) was decided by `sync.Map` iteration order, flipping clients between "retry" and "this block will never exist". |

### P2

Confirmed `getSlot` no longer upgraded to a processed slot · `requestAirdrop` joins the non-idempotent write guard · poller `Bootstrap` made idempotent (genesis-validation retries leaked a ticker goroutine per attempt, forever) · SVM config validated at startup instead of failing asynchronously · `enforceBlockAvailability: false` and `maxFinalizedSlotLag: 0` now survive defaults merging, and `upstreamDefaults.svm` is merged at all · commitment injection respects each method's real parameter shape (`getLeaderSchedule` takes its config at index 0) and clamps to a commitment the method accepts · HTTP 401/403 with a JSON-RPC error body classify as auth failures · the `getBlock` availability guard fails **open** on unknown shred data and is overflow-safe · `context.slot` harvesting no longer scans multi-megabyte `getBlock` payloads · slot counters feed the health tracker through `OnValue` (so rejected values are not reported and cross-instance updates are) and register large-rollback alarms.

### Five defects found while fixing others

- The exhausted-cause **dominance scan kept the _last_ cause** of the dominant bucket, not the first as its own comment claimed. Invisible while ordering was random; would have selected exactly the wrong variant once ordering became deterministic.
- The network **sweep ignored `retryableTowardNetwork=false`**, so an authoritative terminal answer still burned every remaining upstream.
- A **data race on `SvmStatePoller.Enabled`**, surfaced by a new `-race` test. Concurrent `Bootstrap` is reachable in production — `upstream/registry.go` converges concurrent attempts with `LoadOrStore` and each one calls `Bootstrap`. The field was copied from the EVM poller without copying the EVM reader, so it was write-only dead state; deleted rather than made atomic.
- **A startup panic on any Alchemy Solana endpoint.** `thirdparty/alchemy.go:310` dereferenced `upstream.Evm.ChainId` in a debug log unconditionally. `Evm` is nil for `type: svm`, and the branch that populates it only runs when the endpoint must be *derived* — so a user-supplied `https://solana-mainnet.g.alchemy.com/v2/KEY` skipped it and eRPC `SIGSEGV`'d at boot, at any log level (zerolog evaluates arguments eagerly). Alchemy sells Solana endpoints, so this was reachable by a first-time SVM user following the docs. Verified both directions: panic without the guard, clean `validate` with it across alchemy, quicknode, helius, drpc and chainstack SVM upstreams. Every other vendor's `Evm` deref sits inside the endpoint-derivation branch an SVM upstream never enters.
- **The extractor's bare-HTTP-failure branch was unreachable.** It gated on `jr.Error == nil`, but the parse layer never yields a nil error for an unparseable body — it synthesizes `-32700`. So a plaintext or HTML `429` from a CDN in front of a provider fell through to the parse-error case and became a hard non-retryable *client* error: no failover, no capacity signal, and the caller saw a parse error instead of a rate limit. Measured end to end through the real HTTP path. `-32700` is also no longer grouped with caller-side errors — eRPC serializes the outbound request itself, so a parse error always means the upstream sent something unparseable, which is an upstream fault and must stay retryable.

### Shared-path changes (affect EVM too, deliberately)

1. **Deterministic cause ordering** in `NewErrUpstreamsExhausted` — an identical multiset of causes now always yields the same client-visible code. Ordering is retryable-first, then a stable key. Architecture-neutral: `common/` contains no Solana codes; SVM expresses "authoritative skip" through the same `retryableTowardNetwork` flag the retry path already reads.
2. **The sweep honors `retryableTowardNetwork=false`** as a stop condition. Two EVM errors already carried that flag and were being ignored: OP-Stack "sender is over rate limit" (whose own comment says retrying elsewhere is futile) and `ErrEndpointNonceException`. No error shape changes — only fewer wasted round-trips.

### Rejected on the validation gate

- **Go exported-constructor compatibility.** Real break, but rejected as a merge blocker: the project is v0 and routinely changes exported constructors without shims.
- **Poller requests contaminating the circuit breaker.** Rejected as PR-specific: `IsInternal` has no setter anywhere in the codebase, so EVM's pre-existing pollers behave identically. A codebase-wide follow-up, not an SVM defect.

## Production evidence

The Goldsky Solana devnet ETL has run its full production workload through this branch since the 2026-07-24 cutover from direct QuickNode — producer and consumer both on eRPC.

| Signal | Result |
|---|---|
| Forward progress | +165,235 slots in 16h48m = **2.733 slots/s**, matching devnet block production |
| Freshness | block-lag **6.2–7.7s**, flat, no drift |
| Consumer | 40,409 log lines, **0 errors, 0 restarts** over 17.5h |
| `getBlock` through eRPC | p50 **120ms**, p95 **284ms**, **0** sampled errors |
| Provider distribution | Helius, QuickNode and Alchemy all serving |
| Availability guard | **0** rejections/hour (was 82/hour before the guard-margin fix) |

This exercises `getSlot` and `getBlock` continuously and hard. It does **not** exercise the SVM cache, `getBlockHeight`, `getSignaturesForAddress`, `requestAirdrop`, or the TypeScript config path — which is precisely why the review above was necessary and why most of these defects survived a working staging deployment.

## Documentation and config surface

`docs/pages/**` previously contained **zero** SVM references while this PR added a whole network architecture. Adds an SVM cache page, an SVM reference section (commitment semantics, slot tracking), SVM coverage in the networks/upstreams pages, and the SVM error contract in the error reference — all registered in `_meta.js`. Also corrects EVM-universal claims that are wrong for Solana, notably `monitoring.mdx`'s blanket assertion that finalized catch-up is always wasted.

Both root design documents now match the shipped code. Each carries a corrections preamble, and the bodies were brought in line rather than just annotated — the `minContextSlot` reading, the `getSlot`/`getBlockHeight` conflation, the cache-finality tables, and the write-method set. The example configs are now verified to actually load; the main one had four blockers beyond the dropped key.

## Testing

`make build`, `make test-fast` and the TypeScript type suite pass against current `main` merged into this branch. `go vet ./...` is clean apart from one pre-existing warning in `erpc/query_executor_test.go`, which this PR never touches and which is present at the base.

This round adds **44 test functions**, deliberately weighted toward the levels the original branch never exercised: network-level failover and cache behavior through the real HTTP path, poller lifecycle, and concurrency under `-race`. Several defects above were found *by* writing them, and two proposed tests were rejected as unwritable-without-a-production-fix rather than being weakened to pass.

One disclosure, with the exact tally rather than a reassuring summary. `TestHttpServer_Gzip_E2E_StreamedJsonRpc` failed in **1 of 3** full `make test-fast` runs on this branch. It is green in isolation, green at `-count 2`, and green across two full `TestHttpServer_` shard runs; `main` merged to this branch's base ran the full suite green once. The test arrived in `0cc20a21` — the very `main` commit merged here — and this branch touches nothing in the gzip, `http_server`, `clients`, or EVM cache path (every `common/defaults.go` hunk is SVM-scoped, and the one that looks shared is comment-only). The failing assertion is that the second identical request is a cache `HIT`, which is ristretto-eviction sensitive under full parallel load. Assessed as pre-existing flakiness rather than a regression, but flagged rather than buried — if a reviewer disagrees, the deciding experiment is repeated full-suite runs on `main` alone.

## Known follow-ups

- Track a `confirmed` slot tip in the poller so confirmed `getSlot` can be floored against its own commitment rather than passed through uncorrected.
- Promote fully-in-the-past `getBlocks(start, end)` ranges to finalized caching; needs a comparison against the poller's finalized slot, which makes finality depend on mutable poller state.
- Make pollers set `IsInternal` (codebase-wide, EVM and SVM equally) and fix the docs that already claim they do.
- WebSocket subscriptions, fork genesis-table auto-discovery, and commitment-downgrade detection remain out of scope (see the design doc's Gaps & Limitations).

Co-authored with @andreclaro.

---

> [!NOTE]
> **High Risk**
> Large new architecture path affects caching, finality, failover, and wire error contracts; several correctness fixes (base58 cache keys, moving-head finality, write non-retry) show how wrong behavior could reach production. Shared retry/cause-ordering changes also touch EVM exhaustion handling.
> 
> **Overview**
> Introduces **SVM as a peer architecture to EVM** by registering `architecture/svm` on a generic `ArchitectureHandler` seam (EVM gets a thin `handler.go` wrapper). The core pipeline dispatches project/network/upstream hooks, finality, cache DAL, state pollers, and JSON-RPC error extraction per architecture instead of hard-coded `evm` imports.
> 
> **Solana-specific behavior** includes a slot poller (processed/finalized slots, health, shred-insert lag), shape-aware **commitment injection before the cache read** (fixes GET/SET key mismatch), per-method finality (only slot-pinned `getBlock`/`getTransaction` at `finalized` are immutable; moving-head reads stay realtime), case-preserving cache keys, genesis-hash validation for `svm:<cluster>` and `svm:<chain>:<cluster>`, write guards for `sendTransaction`/`sendRawTransaction`/`requestAirdrop`, and an error normalizer that **preserves native Solana codes and `error.data`** while classifying retry/auth/capacity for routing.
> 
> **EVM touchpoints** in this diff are mostly mechanical: block-tip access goes through `common.EvmHighestLatestBlockNumber` / `EvmHighestFinalizedBlockNumber` helpers, and the EVM extractor **no-ops on SVM upstreams** so composite extraction order does not swallow Solana semantics.
> 
> **Docs** add superseded and v0.5 SVM design records (with inline corrections vs shipped code) and extend **README** with Solana quick start, network ID forms, and SVM caching/commitment notes.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4072b8ce8f10f1bd28f4d23c73a2946e8702c88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jeongseup:
I hope to use it for solana RPC

kasrakhosravi:
## Shadow A/B: EVM backward-compatibility under real prod traffic

Ran this branch against live mirrored production EVM traffic on `erpc-shadow` and compared it to `main`, changing **only the image**.

### Method

Two arms, one byte-identical config (`sha256:2b7709f8…`), ~1,400 rps of real mirrored prod traffic across base / hyperevm / arbitrum-one / polygon / mainnet / bsc:

| Arm | Image | Commit |
|---|---|---|
| control | `shadow-test:control-e8a490af` | `e8a490af` — main at this PR's base |
| test | `shadow-test:test-cafc63fe` | `cafc63fe` — this PR's head |

Both built from source with the same toolchain. The existing `flairsdk/erpc:pr-799-6` was **not** used: it predates `erpc/http_server.go` (error serializer — EVM responses flow through it) and `thirdparty/alchemy.go`, so it would have validated the wrong code.

Oracle is eRPC's own `erpc_shadow_response_{identical,mismatch}_total` — prod compares its served response against the shadow's.

### The aggregate number is misleading

Raw aggregate mismatch was 11.328% (main) vs 12.995% (799), but that is a **method-mix artifact**, not behaviour. `eth_getLogs` volume alone shifted 128,374 → 49,634 between windows; head-race methods mismatch ~28% and deterministic ones ~0.02%, so mix changes swing the aggregate on their own.

Splitting by whether a method's result is deterministic for a given input:

```
                  main                        pr-799                      delta
DETERMINISTIC     0.0244%  (65/266,082)       0.0224%  (44/196,678)      -0.0021 pp
head-race        28.6267%  (49,953)          28.3882%  (46,987)          -0.2385 pp
```

Per deterministic method — identical on both arms:

| method | main | pr-799 | n (799) |
|---|---:|---:|---:|
| `eth_call` | 0.00% | 0.00% | 77,989 |
| `eth_getLogs` | 0.00% | 0.00% | 49,634 |
| `debug_traceTransaction` | 0.00% | 0.00% | 49,727 |
| `debug_traceBlockByNumber` | 0.01% | 0.00% | 14,419 |
| `trace_filter` | 1.41% | 0.98% | 4,479 |
| `eth_getCode` / `trace_block` | 0.00% | 0.00% | 430 |

**pr-799 is equal or better on every measured axis.** No regression.

### Why the head-race numbers are noise

`eth_blockNumber` mismatches **75.9% under main against main itself** — prod and shadow sample the tip microseconds apart. To size natural variance I compared two consecutive 15-minute windows within the control arm (no code change at all): per-chain swings up to **2.0 pp**, overall 0.34 pp. Per-chain deltas of +0.13 to +1.3 pp sit inside that. hyperevm (+4.3) and mainnet (−4.0) exceed it but move in opposite directions and resolve entirely into head-race methods.

### Infra (clean 30-min windows)

| metric | main | pr-799 | delta |
|---|---:|---:|---:|
| CPU cores | 3.741 | 3.213 | −14.1% |
| memory GiB | 1.033 | 0.965 | −6.6% |
| heap GiB | 0.457 | 0.833 | **+82.1%** |
| goroutines | 4,499 | 5,506 | **+22.4%** |
| GC / s | 1.849 | 1.712 | −7.4% |
| CPU throttled | 0.030 | 0.000 | −100% |

CPU and memory improved. Heap and goroutine count are up materially in relative terms though small in absolute (0.83 GiB against a 14 GiB `GOMEMLIMIT`); worth a look but not a merge blocker at this scale.

### Config compatibility

pr-799 booted on a config containing no SVM networks with **0 errors and 0 decode complaints**, and served all six EVM chains normally — the existing-deployment upgrade path. eRPC strict-decodes with `KnownFields(true)`, so this exercises it directly.

### Caveats

- One 30-min window per arm, sequential rather than concurrent, so slow drift is not fully excluded. The within-control variance measurement is what bounds this.
- SVM traffic was deliberately not exercised here — SVM is new surface and cannot regress existing behaviour. This tested only the EVM pipeline refactor (`upstream/upstream.go` −290 lines into `evm_upstream_ops.go`, `networks.go` +214, `network_executor.go` +72).
- All oracle errors observed were pod-restart artifacts, confined to two 2-minute windows at each deploy; control's restart produced more (5,564) than the test's (3,323). 28 of 31 windows were exactly zero.

### Verdict

**EVM backward compatibility holds.** On 196,678 graded deterministic-method comparisons, pr-799 diverges from main by −0.0021 pp — better, and far inside the ±2.0 pp noise floor.

Prod was never modified during the experiment; its mirror config was left untouched as a control (CPU −1.4%, p95 −2.7% across the run). `erpc-shadow` has been restored to #948's image and config.

kasrakhosravi:
### Correction to the infra table above

The heap and goroutine figures in my previous comment were **single-instant gauge samples**, which is invalid methodology against `GOGC=200` — heap oscillates by design between GC cycles, so one sample says little. Re-measured as time-averages over the same two 30-minute windows:

| metric | main | pr-799 | delta |
|---|---:|---:|---:|
| heap_inuse avg GiB | 0.634 | 0.605 | −4.6% |
| heap_inuse max GiB | 0.858 | 0.913 | +6.4% |
| heap_alloc avg GiB | 0.517 | 0.488 | −5.6% |
| RSS avg GiB | 0.815 | 0.807 | −0.9% |
| RSS max GiB | 1.215 | 1.268 | +4.4% |
| alloc rate MiB/s | 813.6 | 738.2 | −9.3% |
| GC cycles/s | 1.846 | 1.710 | −7.3% |
| GC pause p99 ms | 9.155 | 6.207 | −32.2% |
| CPU cores avg | 3.750 | 3.213 | −14.3% |
| **goroutines avg** | **4,152** | **4,981** | **+20.0%** |
| goroutines max | 6,290 | 6,549 | +4.1% |

**The reported "+82.1% heap" was an artifact and is withdrawn** — average heap is slightly *lower* on this branch, allocation rate is down 9.3%, and GC pause p99 improves 32%.

The one genuine resource difference is **steady-state goroutines: +20% (~830)**. Peak is only +4.1%, so this is a higher persistent baseline rather than a leak within the window. It is unexplained; the plausible cause is additional per-upstream workers/pollers introduced by the architecture-handler split, which would scale with upstream count — shadow ran 64 upstreams, and a prod tenant with substantially more would see proportionally more. Cheap to confirm with a goroutine profile before merge.

Absolute cost is small (~830 goroutines ≈ single-digit MiB of stacks), but it should be understood rather than assumed benign.

kasrakhosravi:
## Sustained shadow run: 3h50m of mirrored EVM traffic

Follow-up to the earlier 30-minute A/B. This branch ran for **3h50m** on a shadow instance fed a sampled mirror of real production EVM traffic (~1.4–2.0k rps across 6 EVM chains), using eRPC's own `shadow: { enabled: true }` upstream mechanism — so the oracle is `erpc_shadow_response_{identical,mismatch}_total`, comparing the served response against this branch's response on the same request.

TL;DR: **EVM correctness looks clean on 9.4M graded comparisons. There is one unexplained resource signal — goroutine growth — that I could not fully attribute, described at the bottom.**

### Correctness

Splitting methods by whether the result is deterministic for a given input (a mismatch there is a genuine bug) versus head-race (result legitimately depends on *when* you ask):

```
DETERMINISTIC   9,433,291 graded    411 mismatch   0.0044%
head-race       6,741,130 graded  1,064,587        15.79%
```

| method | graded | mismatch | rate |
|---|---:|---:|---:|
| `eth_getLogs` | 8,000,027 | 11 | 0.0001% |
| `eth_call` | 900,343 | **0** | 0.0000% |
| `debug_traceTransaction` | 364,143 | **0** | 0.0000% |
| `debug_traceBlockByNumber` | 108,746 | 1 | 0.0009% |
| `trace_filter` | 55,616 | 399 | 0.7174% |
| `eth_getCode`, `trace_block` | 4,416 | **0** | 0.0000% |

A `main` build at this PR's merge base, run earlier on the same config and same mirror, scored **0.0244%** on the deterministic set (n=266,082). This branch is below that floor, so no regression is visible. `trace_filter` was 1.41% on `main` and 0.72% here.

For scale on the head-race side: `eth_blockNumber` mismatches ~68%, and it does so on `main` against `main` too — two instances sampling the chain tip microseconds apart legitimately disagree. That class of number says nothing about correctness, which is exactly why the split matters.

### Reliability

```
requests       21,346,427
failed                 31    (0.00015%)
```

The only oracle errors (122) were in the two 2-minute windows around pod restarts, where the mirroring side briefly can't reach the shadow. Steady-state windows were zero.

### Latency

Per-chain p95/p99 were equal or better than the `main` control on 4 of 6 chains, and the two that looked worse resolved on inspection:

- Aggregate p95 rises were **method-mix artifacts** — one chain's cheap `eth_getLogs` share halved while `eth_call` and receipts tripled, so the aggregate moved with nothing getting slower.
- `eth_getBlockByHash` looked like a 5.7x regression in a 20-minute window. Over the full 3h30m it was **0.301s vs 0.729s before and 0.281s after** — i.e. the spot check caught a transient, and the sustained number is fine.

### The one open item: goroutine growth

Over the run, with load *decreasing*:

```
               first 30m    last 30m    slope/hr
goroutines         4,670       8,481      +1,210
heap GiB           0.699       0.877      +0.054
rss GiB            0.925       1.177      +0.074
cpu cores          3.762       3.888      -0.055
rps                1,844       1,425         -98
```

Goroutines grew 1.88x. The curve is flat for the first ~1.5h then climbs steadily. Compared against an instance running `main` on the same clock against the same upstreams:

```
elapsed     main    this branch
  0.0h     1.00x          1.00x
  2.3h     1.91x          1.44x
  3.7h     0.94x          1.88x
```

`main` spikes under load and **returns to baseline**; this run **ratchets and does not recover**. CPU stayed flat and GC rate fell, so it isn't load.

**Caveat — I could not cleanly attribute this, and it may not be the branch.** To run any image other than the one previously deployed there, I had to replace the shadow's config (it carried experiment-only fields from another branch that a different build strict-rejects via `KnownFields(true)`). That replacement also **restored `consensus` on `eth_getLogs` / `eth_getBlockReceipts`**, and `eth_getLogs` was 8M of the 21M requests. Consensus fans out concurrently, so it plausibly accounts for a higher goroutine baseline and growth. The previously-deployed build on the same instance also grew, at roughly half the rate.

Isolating it needs `main` + the *identical* config for a comparable duration — my `main` control used that config but only ran 30 minutes, too short to show a slope. Happy to run that if useful.

### What this run does not cover

- **Shared state was disabled** on this instance, while a typical multi-replica deployment uses it — and this PR touches `upstream/registry.go`, `networks_registry.go` and the state pollers. Largest untested area.
- Single replica; no multi-pod coordination.
- 6 EVM chains, read-only mirror. No write path (`eth_sendRawTransaction`), and shadow responses are discarded, so client-visible failure modes aren't exercised.
- No SVM networks were configured — deliberately. SVM is new surface and can't regress existing behaviour; this run targeted the EVM pipeline refactor (`upstream/upstream.go` → `evm_upstream_ops.go`, `networks.go`, `network_executor.go`).

### Methodology notes that may be useful to others

1. Aggregate mismatch rate is misleading — method mix dominates it. Split deterministic from head-race before drawing conclusions.
2. Don't sample `go_memstats_heap_inuse_bytes` as an instant gauge with `GOGC=200`; I reported +82% heap from a point sample, and the time-average was −4.6%.
3. A same-clock control on the same upstreams is essential — provider latency moves both instances together and otherwise reads as a regression.
4. For leak detection, mean-reversion vs ratchet is a better signal than slope; both instances had rising segments, only one failed to recover.
5. Short windows lie. A 20-minute sample showed a 5.7x latency regression that a 3h30m window showed as an improvement.

kasrakhosravi:
Triaged all three. **All three are real**; one is bigger than reported and one is half wrong.

### `getConfirmedBlock` missing from the finality tables — fixed in `749f8883`

Correct, and it turned out to be **three** tables, not two. `handler.go` dispatches `case "getBlock", "getConfirmedBlock":` to the same guard, so the alias is routed as a block read, but it was absent from:

| Table | Consequence |
|---|---|
| `slotPinnedMethods` | `GetFinality` fell to step 4 → Realtime, so a finalized historical read via the old name could never match an immutable cache policy |
| `commitmentOptionsIndex` | no commitment injection; and since `resolveCommitment` is the shared seam, this *also* silently downgraded classification |
| `atLeastConfirmedMethods` | **not in your report** — the clamp never fired, so a network defaulting to `processed` injected an unclamped commitment that agave rejects `-32602`. The alias was broken outright on such a network while `getBlock` beside it worked |

The third surfaced when the parity tests went looking for the same class of gap.

**One correction:** you also flagged `getConfirmedTransaction`. That name appears **nowhere in this repo** — there is no alias to keep in sync, so I did not add it. If it's ever introduced it needs all three entries.

16 subtests now assert `getBlock`/`getConfirmedBlock` parity across finality, injection and clamping, each with one shared `want` across both names so a divergence is inexpressible rather than merely detected. All three entries mutation-verified.

### Pre-forward guard bypasses the retry — confirmed, not fixed here

Verified: `erpc/networks.go:1170` does `return nil, err` the moment `HandleNetworkPreForward` reports `handled=true`, before any failsafe executor. So the `ErrEndpointMissingData` the guard raises is terminal and `shouldRetryWithReason` never sees it.

Not fixing it in this pass. The guard wants a retryable condition, but pre-forward is architecturally outside the retry boundary, so the fix is a design choice — move the guard inside the executor, or give pre-forward a way to signal "retryable" — not a typo. It deserves its own change with its own tests rather than being bolted onto a cache PR.

### Poller `Bootstrap` skips the initial poll — confirmed, not fixed here

Verified against EVM:

```text
EVM  Bootstrap, loop already running -> return e.Poll(ctx)
SVM  Bootstrap, loop already running -> return nil
SVM  Bootstrap, fresh                -> spawns goroutine, returns nil; Poll only on the first tick
```

So slot tip, shred lag and health stay unset for up to one debounce interval after bootstrap. Real divergence. Same reasoning as above — it changes startup ordering on the SVM hot path and belongs in its own change.

Both are tracked; happy to take them in this PR if you'd rather not split.

kasrakhosravi:
`f4072b8c` adds `jsonParsed` serving to the BDS cache path — the encoding that turned out to be **100% of the real `getBlock` traffic** (the Solana ETL walks mainnet with `encoding: jsonParsed`; the guard added earlier correctly refused it, so the cache served nothing).

Three-repo change, this is the last leg:

1. [manifesto#16](https://github.com/blockchain-data-standards/manifesto/pull/16) — contract (`GetBlockRequest.includeParsed`, `CompiledInstruction.parsed`) + the mechanical renderer (`BlockToJsonRpcParsed`), golden-tested against a live node. Found and fixed two pre-existing shared-path fidelity bugs live under `encoding: json` (returnData null-vs-omit, uiAmount zero-vs-null).
2. [prism#440](https://github.com/goldsky-io/prism/pull/440) — the reader runs **Agave's own parsers** (`solana-transaction-status`) on request and attaches the result per instruction. Draft until #16 merges (submodule pin).
3. This branch — the guard narrows instead of rejecting: `jsonParsed` is served with `includeParsed=true` and spliced; `base58`/`base64`/`base64+zstd` still fall through.

The `replace` directive pinning the manifesto fork commit comes out when #16 merges — same dance as the descriptor fix.

Worth restating the operational caveats that jsonParsed does NOT fix: the reader's coverage gap at the tip (writer backfilling) and the 4–5× full-block latency vs live providers. This makes the cache *eligible* for the ETL's traffic; ranking and coverage decide how often it actually serves.

kasrakhosravi:
CI green after two unrelated fixes — recording both so the next person doesn't re-diagnose them.

**1. Merge-only compile break (real, fixed).** This branch adds the `*svm.SvmJsonRpcCache` parameter to `NewERPC`; `main` concurrently added a caller in `erpc/http_server_test.go` that predates it. Neither branch is broken alone — only the merge is, which is why CI (which builds the PR merge commit) failed while both sides were green. Merged `main` in and added the missing argument.

**2. `TestHttpServer_Gzip_E2E_StreamedJsonRpc` is flaky** (`erpc/http_gzip_e2e_test.go:144`) — unrelated to this PR, worth a look separately. It asserts *"second identical finalized request should be a cache hit"* and got `MISS`. Passes 3/3 locally and passed on a bare re-run of the same commit, so the cache write from request 1 isn't reliably visible to request 2 on a loaded runner — an async-write race in the test, not the code under test.

Also in this push: manifesto repinned to the upstream merge commit (`v0.0.0-20260805101127-4242bb202b55`) now that [manifesto#17](https://github.com/blockchain-data-standards/manifesto/pull/17) landed, and the fork `replace` directive is gone.

PR #811, opened 2026-03-19
feat: dynamic quantile-based timeout policy

kasrakhosravi:
## Summary

- Extends `TimeoutPolicyConfig` with `Quantile`, `MinDuration`, `MaxDuration` fields — same shape as the hedge config.
- When `Quantile > 0`, timeout is computed dynamically from method latency percentiles (DDSketch p-quantile) at both network and upstream levels.
- When `Quantile` is 0 or omitted, behavior is unchanged (backward compatible).
- Cold start (no metrics yet) falls back to `Duration`, then `MaxDuration`.
- Adds `erpc_network_timeout_duration_seconds` histogram and `erpc_network_timeout_fired_total` counter.
- TypeScript config types updated.

## Architecture

**One path, not two.** Both fixed and quantile-based timeouts are applied via `context.WithTimeoutCause(ErrDynamicTimeoutExceeded)` at the request entry point in `Network.Forward` and `Upstream.Forward`. The failsafe-go `timeout.Policy` is no longer registered — previously it was used for fixed timeouts only, creating two parallel code paths and two error types to translate. Unifying the paths removes the `+5ms` race workaround and the dual error-matching in `TranslateFailsafeError`.

**Sentinel cause distinguishes policy firing from parent deadlines.** `common.ErrDynamicTimeoutExceeded` is set as the context cause exclusively by the timeout policy. The HTTP client and `TranslateFailsafeError` both match on this sentinel to route policy-driven timeouts to `ErrFailsafeTimeoutExceeded` (504), while parent-context deadlines (HTTP server timeouts, caller cancellation) pass through unchanged.

**`Duration` as fallback, not additive.** Unlike hedge where `Delay` is added to the quantile, `Duration` is used purely as a cold-start fallback. Timeout represents a total wait budget, not an offset.

**`TimeoutFunc` on `FailsafeExecutor`.** Replaces `*time.Duration` to allow per-request computation.

## Test plan

- [x] Fixed timeout backward compatibility (wraps as `ErrFailsafeTimeoutExceeded`)
- [x] Fixed timeout exceeding at both network and upstream scope
- [x] Quantile-based dynamic computation with histogram emission at both scopes
- [x] MinDuration boundary enforcement
- [x] Cold start fallback to Duration
- [x] Cold start fallback to MaxDuration (no Duration set)
- [x] Parent-context deadline does NOT get misclassified as policy timeout
- [x] Existing hedge and upstream tests pass (no regressions)
- [x] `go vet` clean

Closes ERPC-356

---

> [!NOTE]
> **Medium Risk**
> Touches core request forwarding/timeout/error-translation paths at both network and upstream scope (including batching), so regressions could change timeout behavior, classification, or retry budgets despite strong new test coverage.
> 
> **Overview**
> Adds a **dynamic (quantile-based) timeout mode** to the `timeout` failsafe policy by extending `TimeoutPolicyConfig` with `quantile`, `minDuration`, and `maxDuration`, including validation and updated TypeScript config types.
> 
> Unifies timeout enforcement for both network- and upstream-scope by applying `context.WithTimeoutCause(..., common.ErrDynamicTimeoutExceeded)` in `Network.Forward` and `Upstream.Forward` (removing failsafe-go’s timeout policy), and updates HTTP JSON-RPC client (single + batch) to propagate `context.Cause` so policy-driven timeout classification survives batching and transport error paths.
> 
> Introduces new observability for dynamic timeouts: `erpc_network_timeout_duration_seconds` histogram (computed timeout values) and `erpc_network_timeout_fired_total` counter with scope attribution, plus docs and Grafana dashboard updates; adds comprehensive tests covering fixed backward-compat, quantile computation, cold-start fallbacks, lifecycle-scoped network timeouts, and misclassification/double-counting regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2345a9c3bcccc5492ad369846c51b408ad2a60ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -446,12 +446,6 @@ func (c *GenericHttpJsonRpcClient) processBatch(alreadyLocked bool) {
 		// Each request's own ctx carries the policy-driven cause (e.g.
 		// ErrDynamicTimeoutExceeded), while the shared batch ctx only has the
 		// earliest-deadline plain DeadlineExceeded. Prefer the per-request cause
-		// so the sentinel survives upstream-level error classification. Give the
-		// req.ctx a brief moment to propagate its cancellation if it has the same
-		// deadline as batchCtx (they fire within microseconds of each other).
-		// Each request's own ctx carries the policy-driven cause (e.g.
-		// ErrDynamicTimeoutExceeded), while the shared batch ctx only has the
-		// earliest-deadline plain DeadlineExceeded. Prefer the per-request cause
 		// so the sentinel survives upstream-level error classification.
 		for _, req := range requests {
 			reqErr := err
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -179,10 +179,16 @@ func (c *GenericHttpJsonRpcClient) SendRequest(ctx context.Context, req *common.
 	case err := <-errChan:
 		return nil, err
 	case <-ctx.Done():
-		err := ctx.Err()
+		// Prefer context.Cause so policy-driven sentinels (e.g. ErrDynamicTimeoutExceeded)
+		// survive upstream-level error translation instead of being flattened to plain
+		// context.DeadlineExceeded.
+		err := context.Cause(ctx)
+		if err == nil {
+			err = ctx.Err()
+		}
 		// TODO For both of these conditions failsafe library can introduce carrying
 		//      the "cause" so we know this cancellation is due to Hedge policy for example.
-		if errors.Is(err, context.DeadlineExceeded) {
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, common.ErrDynamicTimeoutExceeded) {
 			err = common.NewErrEndpointRequestTimeout(time.Since(startedAt), err)
 		} else if errors.Is(err, context.Canceled) {
 			err = common.NewErrEndpointRequestCanceled(err)
@@ -225,8 +231,11 @@ func (c *GenericHttpJsonRpcClient) queueRequest(id interface{}, req *batchReques
 	// If the request context is already canceled, fail it immediately and do not queue
 	if err := req.ctx.Err(); err != nil {
 		c.batchMu.Unlock()
-		// propagate a normalized error
-		if errors.Is(err, context.DeadlineExceeded) {
+		// Prefer context.Cause so policy-driven sentinels survive.
+		if cause := context.Cause(req.ctx); cause != nil {
+			err = cause
+		}
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, common.ErrDynamicTimeoutExceeded) {
 			req.err <- common.NewErrEndpointRequestTimeout(0, err)
 		} else {
 			req.err <- common.NewErrEndpointRequestCanceled(err)
@@ -246,7 +255,7 @@ func (c *GenericHttpJsonRpcClient) queueRequest(id interface{}, req *batchReques
 	c.batchRequests[id] = req
 	ctxd, ok := req.ctx.Deadline()
 	if ctxd.After(time.Now()) && ok {
-		// Use the earliest deadline among queued requests so the batch cancels promptly
+		// Use the earliest deadline among queued requests so the batch cancels promptly.
 		if c.batchDeadline == nil || ctxd.Before(*c.batchDeadline) {
 			duration := time.Until(ctxd)
 			c.logger.Trace().Dur("deadline", duration).Msgf("setting batch deadline to earliest request deadline")
@@ -331,7 +340,10 @@ func (c *GenericHttpJsonRpcClient) processBatch(alreadyLocked bool) {
 	for id, br := range requests {
 		if err := br.ctx.Err(); err != nil {
 			delete(requests, id)
-			if errors.Is(err, context.DeadlineExceeded) {
+			if cause := context.Cause(br.ctx); cause != nil {
+				err = cause
+			}
+			if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, common.ErrDynamicTimeoutExceeded) {
 				br.err <- common.NewErrEndpointRequestTimeout(0, err)
 			} else {
 				br.err <- common.NewErrEndpointRequestCanceled(err)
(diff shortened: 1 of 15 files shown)


erpc/networks_registry.go
```diff
@@ -97,11 +96,19 @@ func NewNetwork(
...
+			policyArray := upstream.ToPolicyArray(pls, "consensus", "retry", "hedge")
+			executorPolicies := policyArray
+			if timeoutCfg != nil && timeoutCfg.Quantile == 0 {
+				timeoutPolicy, err := upstream.CreateTimeoutPolicy(&lg, timeoutCfg.Duration.Duration())
+				if err != nil {
+					return nil, err
+				}
+				executorPolicies = append(make([]failsafe.Policy[*common.NormalizedResponse], 0, len(policyArray)+1), timeoutPolicy)
+				executorPolicies = append(executorPolicies, policyArray...)
 			}
```
aramalipoor, line 123: What is happening here exactly? 🤔

erpc/networks.go
```diff
@@ -463,7 +465,31 @@ func (n *Network) Forward(ctx context.Context, req *common.NormalizedRequest) (*
...
+		timeoutDuration = failsafeExecutor.timeoutCfg.Duration.Duration()
+		if failsafeExecutor.timeoutCfg.Quantile > 0 {
+			timeoutDuration, _ = upstream.CalculateQuantileBasedDuration(
+				req,
+				failsafeExecutor.timeoutCfg.Quantile,
+				timeoutDuration,
+				failsafeExecutor.timeoutCfg.MinDuration.Duration(),
+				failsafeExecutor.timeoutCfg.MaxDuration.Duration(),
+			)
+			timeoutPolicy, err := upstream.CreateTimeoutPolicy(failsafeExecutor.logger, timeoutDuration)
```
aramalipoor: Wouldn't this be static/fixed value only on initial setup? We might need to customize the timeout failsafe policy code https://github.com/aramalipoor/failsafe-go/tree/erpc/timeout

erpc/networks_registry.go
kasrakhosravi: I think this is no longer relevant bcause now we are using NewTimeoutFunc directly

erpc/networks.go
kasrakhosravi: you're right — that earlier version was rebuilding a whole failsafe policy per request just to change the duration, which was pointless since failsafe-go only takes a fixed value anyway. Current version just uses context.WithTimeout for the dynamic part and keeps the failsafe policy static as a safety net.

do you still think it makes sense to fork and cusomise timeout for `failsafe-go`?

common/config.go
```diff
@@ -1051,7 +1051,10 @@ func (c *CircuitBreakerPolicyConfig) Copy() *CircuitBreakerPolicyConfig {
 }
 
 type TimeoutPolicyConfig struct {
-	Duration Duration `yaml:"duration,omitempty" json:"duration" tstype:"Duration"`
+	Duration    Duration `yaml:"duration,omitempty" json:"duration" tstype:"Duration"`
+	Quantile    float64  `yaml:"quantile,omitempty" json:"quantile"`
+	MinDuration Duration `yaml:"minDuration,omitempty" json:"minDuration" tstype:"Duration"`
+	MaxDuration Duration `yaml:"maxDuration,omitempty" json:"maxDuration" tstype:"Duration"`
```
kasrakhosravi, line 1057: quick thought on min/max: for production use-cases, I'd suggest starting with just quantile of e.g 0.99 + duration as cold-start fallback — no minDuration/maxDuration. p99 is conservative enough that it almost never kills legitimate requests, and it keeps the config surface minimal.

I know dynamic hedge has min and max clamping, but I think the use-case is different there. thoguths?

erpc/networks.go
aramalipoor: The caller-level timeout I feel is ok for now, but yeah ultimately it's time to retire failsafe-go and reimplement the bits that is important to erpc, lean, robust, DRY without all context hacks we had to do overtime.

common/config.go
aramalipoor: Having min/max is mainly for defense in depth in case things get out of hand, I think they still make sense. And for quantile perhaps we can even do p95, i.e. 5% of the requests would be killed to retry instead.

aramalipoor:
@copilot resolve the merge conflicts in this pull request

common/config.go
kasrakhosravi: Fair enough — keeping min/max. You're right that defense-in-depth matters here: without clamping, a bad GC pause or cold-cache spike can inflate p99 and feed back into itself.

erpc/networks.go
```diff
@@ -515,17 +515,11 @@ func (n *Network) Forward(ctx context.Context, req *common.NormalizedRequest) (*
...
-					//      but allow the failsafe execution to fail with timeout first for proper error handling.
-					*failsafeExecutor.timeout+5*time.Millisecond,
-				)
-
-				defer cancelFn()
+				if td := failsafeExecutor.timeout(execSpanCtx, effectiveReq); td != nil {
+					var cancelFn context.CancelFunc
+					execSpanCtx, cancelFn = context.WithTimeoutCause(execSpanCtx, *td, upstream.ErrDynamicTimeoutExceeded)
+					defer cancelFn()
+				}
```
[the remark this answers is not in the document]
kasrakhosravi: This is an intentional architectural change, not a bug. The PR removes the failsafe-go `timeout.Policy` and unifies both fixed and quantile timeouts through `context.WithTimeoutCause` inside the callback. aramalipoor acknowledged this approach in the thread above ("The caller-level timeout I feel is ok for now").

The per-attempt model is the correct semantic for dynamic quantile-based timeouts — each attempt should get its own deadline derived from current latency percentiles. For fixed timeouts, this changes the behavior: the timeout now applies per-attempt rather than as a total lifecycle budget. This is documented in the PR description under "One path, not two."

If total-budget semantics are needed for fixed timeouts, that should be a follow-up (e.g. re-adding a lifecycle-scoped context timeout at the executor level for fixed mode).

upstream/failsafe.go
```diff
@@ -852,11 +902,15 @@ func TranslateFailsafeError(scope common.Scope, upstreamId string, method string
...
-	// In those cases we return the standard error object as is.
-	if serr, ok := execErr.(common.StandardError); ok {
+	// Timeout-policy firing takes precedence over StandardError short-circuiting.
+	// When the dynamic timeout fires mid-HTTP-request, the client wraps the
+	// cause as ErrEndpointTransportFailure (a StandardError), which would
+	// otherwise hide the timeout classification from the user. The sentinel
+	// cause is only ever set by our own context.WithTimeoutCause, so matching
+	// on it never picks up parent-context deadlines.
+	if errors.Is(execErr, ErrDynamicTimeoutExceeded) && !common.HasErrorCode(execErr, common.ErrCodeFailsafeTimeoutExceeded) {
+		err = common.NewErrFailsafeTimeoutExceeded(scope, execErr, startTime)
```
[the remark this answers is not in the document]
kasrakhosravi: Valid finding. Fixed in ae817c3: `TranslateFailsafeError` now checks `retrypolicy.ExceededError` before the `ErrDynamicTimeoutExceeded` sentinel, so exhausted retries whose last attempt timed out are correctly classified as retry-exceeded (with the timeout as the recursively translated cause) rather than as a bare timeout.

erpc/networks.go
```diff
@@ -739,6 +733,16 @@ func (n *Network) Forward(ctx context.Context, req *common.NormalizedRequest) (*
...
 	if execErr != nil {
+		if errors.Is(execErr, upstream.ErrDynamicTimeoutExceeded) {
+			finality := req.Finality(ctx)
+			telemetry.MetricNetworkTimeoutFiredTotal.WithLabelValues(
+				n.projectId,
+				req.NetworkLabel(),
+				method,
+				finality.String(),
+				string(common.ScopeNetwork),
+			).Inc()
```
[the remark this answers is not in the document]
kasrakhosravi: Valid finding. Fixed in ae817c3: the network-level metric now uses `!common.HasErrorCode(execErr, common.ErrCodeFailsafeTimeoutExceeded)` so it only increments when the sentinel is raw (network-scope timeout), not when it leaks through the unwrap chain from an already-translated upstream-scope timeout.

clients/http_json_rpc_client.go
```diff
@@ -422,26 +434,36 @@ func (c *GenericHttpJsonRpcClient) processBatch(alreadyLocked bool) {
...
+		// Each request's own ctx carries the policy-driven cause (e.g.
+		// ErrDynamicTimeoutExceeded), while the shared batch ctx only has the
+		// earliest-deadline plain DeadlineExceeded. Prefer the per-request cause
+		// so the sentinel survives upstream-level error classification. Give the
+		// req.ctx a brief moment to propagate its cancellation if it has the same
+		// deadline as batchCtx (they fire within microseconds of each other).
+		// Each request's own ctx carries the policy-driven cause (e.g.
+		// ErrDynamicTimeoutExceeded), while the shared batch ctx only has the
+		// earliest-deadline plain DeadlineExceeded. Prefer the per-request cause
+		// so the sentinel survives upstream-level error classification.
```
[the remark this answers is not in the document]
kasrakhosravi: Valid — removed the duplicate comment block in ae817c3.

upstream/upstream.go
```diff
@@ -669,6 +667,16 @@ func (u *Upstream) Forward(ctx context.Context, nrq *common.NormalizedRequest, b
...
 			common.SetTraceSpanError(span, execErr)
+			if errors.Is(execErr, common.ErrDynamicTimeoutExceeded) {
+				finality := nrq.Finality(ctx)
+				telemetry.MetricNetworkTimeoutFiredTotal.WithLabelValues(
+					u.ProjectId,
+					nrq.NetworkLabel(),
+					method,
+					finality.String(),
+					string(common.ScopeUpstream),
+				).Inc()
```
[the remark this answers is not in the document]
kasrakhosravi: Valid finding. Fixed in 207a64d: the upstream metric now guards on `failsafeExecutor.timeout != nil` before checking for the sentinel. If the upstream has no timeout configured, the sentinel must have originated from the network scope and should not be counted at scope=upstream. This ensures network-level timeouts are attributed to scope=network (not suppressed by the HasErrorCode guard after being misattributed upstream).

Note: when both network and upstream have timeouts configured and the network fires first, there is still a theoretical imprecision — the upstream counter fires for a network-originated timeout. A complete fix would require separate sentinels (ErrNetworkTimeoutExceeded / ErrUpstreamTimeoutExceeded). Acceptable trade-off for now.

erpc/networks.go
```diff
@@ -470,6 +470,19 @@ func (n *Network) Forward(ctx context.Context, req *common.NormalizedRequest) (*
 		attribute.String("failsafe.matched_finalities", fmt.Sprintf("%v", failsafeExecutor.finalities)),
 	)
 
+	// Network-level timeout is lifecycle-scoped: it wraps the entire failsafe
+	// execution including retries and hedges. Applying it here (outside the
+	// executor) matches the documented semantics — a network timeout of 5s with
+	// 3 retries still bounds total wall-clock to 5s. Upstream-level timeout,
+	// applied per-attempt inside Upstream.Forward, is independent.
+	if failsafeExecutor.timeout != nil {
```
kasrakhosravi, line 479: @aramalipoor I am thinking if it makes sense to have the dynamic timeout be for the full lifecycle of network request or should it be per attempt?

per attempt sound wasteful and unintuitive 
but per lifecycle does not. feel natural with a p99 latency derivation

aramalipoor: whichever is easiest and least code, both are close enough in terms of impact

kasrakhosravi: hmm interesting. the lifecycle is def cleaner and less code

aramalipoor: nice, let's go for it

PR #822, opened 2026-04-08
feat: add full WebSocket support with subscriptions

jleeh:
## Summary

- **Server-side WS listener**: clients connect via `ws://` on the same HTTP port. Regular JSON-RPC and batch requests reuse the full `Forward()` chain -- caching, upstream scoring, failsafe policies, rate limiting, multiplexing, EVM hooks, and metrics all work identically to HTTP.
- **Upstream WS client**: connect to upstream providers via `ws://` and `wss://` endpoints with persistent connections, request ID matching, reconnection with exponential backoff, and ping/pong keepalive.
- **Subscription manager**: `eth_subscribe`/`eth_unsubscribe` with ID remapping, notification fan-out to multiple clients, subscription deduplication (shared upstream subscriptions), rate limiting, metrics, upstream scoring, `useUpstream` directive support, and automatic cleanup on client disconnect.

## Upstream Health Monitoring

**Client disconnect on upstream drop:**
- When an upstream WS connection drops, all client connections with active subscriptions on that upstream are closed with `CloseGoingAway (1001)`
- Clients receive a clean close frame and can reconnect immediately instead of holding zombie subscriptions

**Periodic upstream health check:**
- Every `healthCheckInterval` (configurable, default 10s), checks if each upstream serving subscriptions is still the highest-scored WS upstream for its network
- Uses eRPC's existing scoring system -- if a better upstream becomes available due to lag, errors, throttling, or any other tracked factor, clients are disconnected to re-subscribe on the better upstream

**Connection guard:**
- `eth_subscribe` is rejected if no healthy upstream WS is available (prevents zombie subscriptions during startup)
- If `eth_subscribe` fails for any reason, client connection is closed with `1001 Going Away` for fast reconnect

**Upstream selection filtering:**
- `selectWsUpstream` checks three health signals before selecting:
  - `IsConnected()`: WS TCP connection is established
  - `IsCordoned()`: not marked unhealthy by the scoring system
  - `LatestBlock() > 0`: state poller has successfully polled at least once

## Configuration

- `maxSubscriptionsPerConnection`: limit per client WS connection (default: 100)
- `healthCheckInterval`: how often to check upstream health for subscriptions (default: 10s)

## Documentation

New `docs/pages/operation/websocket.mdx` covering connecting, subscriptions, connection health and automatic failover, WS upstreams, server configuration, and directives.


jleeh:
Updated the PR to fix the go sec and unit test failures. The only other unit test that failed was `TestConsensusPolicy/all_server_errors_infra_retry_exhausted` but that's not touched by these changes.

aramalipoor:
Thanks @jleeh for the PR, would need find time to dig into it. This has been a very complex yet most wanted feature.

1. Can you share which AI model you used for the implementation? It'll help narrow down scope of revview.
2. Have you used it in production for a long-enough period of time (the usual challenge with websocket implementations)?

jleeh:
No problem @aramalipoor, understand it'll take some time.

1. Was built with Opus 4.6
2. We've rolled this out into a new production environment and have started ramping up load. Will be thousands of req/s over the next coming weeks so will be fixing issues as we find them. Just now pushed up a fix as graceful shutdown of ws conns on shutdown wasn't hooked in. So far though, seems pretty good.

Happy to keep on maintaining this as needed. We've migrated to full erpc in our new environment and needed ws so thought I'd take the step in getting it built out as yeah every EVM proxy lacks it.

jleeh:
Updated the PR @aramalipoor with fixes and tidied up the code to the point I'm happy with it, as it was pretty messy before.

The main design decision with ws is for the `eth_subscribe` method and how it handles upstreams. In aid to keep it simple, rather than having logic where erpc would transparently resubscribe clients if a new endpoint has a higher score, it will disconnect all clients using that upstream forcing reconnect logic onto clients.

Main reason for this is that transparent resubscribes could cause compatibility issues, for example with head gaps or missing logs on subscription. Idea being if any client using ws gets disconnected, they should handle all of this already with backfilling so adding any implicit behaviour could cause problems that are hard to debug and understand.

Main idea with the WS implementation is that any non-subscription RPC calls will go to the existing HTTP pipeline and be returned back via WS. Any subscription calls are sent over WS and erpc will monitor upstream scores and disconnect clients if a better one becomes available, forcing them to reconnect and then getting the higher rated upstream on reconnect.

Happy for it to be reviewed now as the core of it shouldn't be changed, may just push fixes for any new issues we find. Once we're sending larger load and are happy with it, can reply again for when we're comfortable with the feature.

aramalipoor:
> Main reason for this is that transparent resubscribes could cause compatibility issues, for example with head gaps or missing logs on subscription. 

Actually for this very reason when initially I thought about "websocket" and "subscription" my idea was to create an indexer logic, which actually does listen to upstreams (whether http polling or websocket) and proactively indexes blocks/logs with reorg handling included (borrowing logic from Ponder.sh and Subgraphs maybe). And then we can emit the stream either via websockets or even things like kafka or sqs/sns.

This way erpc would be a more meaningful "websocket/subscription" proxy vs just a websocket rerouting without adding much benefit. i.e. a client code can simply implement such logic (if node A websocket is broken, switch to node B... or just keep retrying node A). 

I know it feels more sophisticated than this implementation but perhaps using AI we can achieve it faster, what are your thoughts? 🤔 

P.S. in some internal projects we achieved good stable indexer logic with pretty simple structure (haven't found the time to port it to erpc)

jleeh:
Do agree that having an indexer would be a very nice feature and would eliminate reconnects, I feel for this first version at least it adds a lot of complexity for what could be done after this PR.

Even in this more basic implementation, there's still some big wins versus just traditional ws. Any non-subscribe rpc call over ws benefits from the full feature set of erpc. Also, since the ws implementation implements `ClientInterface` then it's also used within the `Forward` pipeline so ws backends can also be used for http calls if they score higher. In theory you could run full erpc with just a ws backend and both http & ws will still work (not tried this).

Honest opinion is that an indexer would be great, but it'd be a significant amount of work that could add a lot more bugs in edge cases with not as much benefit when client disconnects give most of the benefit just with slower response if an upstream has issues.

jleeh:
Going to close this PR @aramalipoor, as currently implementing the indexer. Was seeing too many reconnects & errors with current logic on fast block producing chains, so going to rework this and will raise when ready.

aramalipoor:
Thanks for sharing your observations @jleeh. Yes, this would be super cool and I'm pretty sure many users of erpc would love to see it finally land 🤝 

It would be ideal if the implementation is transport-agnostic, basically ingress/egress could be anything, the indexer core would be same indexing and same emitting logic. That way whether we implement websocket for listening now, or later add Kafka as a sink it'll be all reusing same components, same reorg guarantees, etc.

Feel free to brainstorm the features/architecture here as well.

jleeh:
Raised the new pr here @aramalipoor:
https://github.com/erpc/erpc/pull/836

Implemented the indexer that is ingress/egress agnostic, ready for any egress/ingress adapters like Kafka, gRPC etc. I've been running this in our new production environment over a large variety of chains from L1's and L2's and it seems solid. Also made a few other tweaks around metric cardinality at scale with a lot of networks & upstreams, adding block number & finalised block number to shared state.

PR #826, opened 2026-04-09
fix: mark consensus dispute/low-participants errors as non-retryable

kasrakhosravi:
## Summary
- Set `retryableTowardNetwork: false` on `ErrConsensusDispute` and `ErrConsensusLowParticipants`
- Prevents the retry policy from re-executing entire consensus rounds when consensus cannot reach agreement, which cascades upstream pressure

## Test plan
- [x] Unit test: `ErrConsensusDispute` is not retryable toward network
- [x] Unit test: `ErrConsensusLowParticipants` is not retryable toward network
- [x] All existing `IsRetryableTowardNetwork` tests pass (no regression)

---

> [!NOTE]
> **Low Risk**
> Low risk: only changes retry classification for two consensus error constructors and adds unit coverage; behavior change is limited to reducing network-level retries in these specific failure modes.
> 
> **Overview**
> Prevents network-level retries from re-running full consensus rounds when consensus cannot be reached.
> 
> `NewErrConsensusDispute` and `NewErrConsensusLowParticipants` now set `details.retryableTowardNetwork=false`, making `IsRetryableTowardNetwork` treat these consensus outcomes as non-retryable. Adds unit tests asserting both consensus errors are not retryable.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bb958de218f53c0cd1d8c64f1ff4721e40761f0d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/common/errors.go
+++ b/common/errors.go
@@ -2499,7 +2499,8 @@ var NewErrConsensusDispute = func(message string, participants []ParticipantInfo
 			Message: message,
 			Cause:   errors.Join(causes...),
 			Details: map[string]interface{}{
-				"participants": participants,
+				"participants":           participants,
+				"retryableTowardNetwork": false,
 			},
 		},
 	}
@@ -2533,7 +2534,8 @@ var NewErrConsensusLowParticipants = func(message string, participants []Partici
 			Message: message,
 			Cause:   errors.Join(causes...),
 			Details: map[string]interface{}{
-				"participants": participants,
+				"participants":           participants,
+				"retryableTowardNetwork": false,
 			},
 		},
 	}
--- a/common/errors_retry_test.go
+++ b/common/errors_retry_test.go
@@ -268,6 +268,39 @@ func TestIsRetryableTowardNetwork_UserReportedScenario(t *testing.T) {
 	})
 }
 
+func TestIsRetryableTowardNetwork_ConsensusErrors(t *testing.T) {
+	t.Run("ErrConsensusDispute_NotRetryable", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


common/errors.go
```diff
@@ -2499,7 +2499,8 @@ var NewErrConsensusDispute = func(message string, participants []ParticipantInfo
 			Message: message,
 			Cause:   errors.Join(causes...),
 			Details: map[string]interface{}{
-				"participants": participants,
+				"participants":           participants,
+				"retryableTowardNetwork": false,
```
aramalipoor, line 2503: Shouldn't we actually retry again (after the defined delay) in case the upstreams are in sync? What would this gains us in prod, I guess probably lower latency? But anyway the client would retry anyway so 'actual' latency would be same no?

aramalipoor: Same for low participants

kasrakhosravi: the main issue is that network-level delay: 0 ([recommended default](https://docs.erpc.cloud/config/projects/networks#config)) causes cascading retries on consensus disputes — each retry spawns a full consensus round against the same upstream pool that just failed

we can probably combat this by setting retry delay at network level

aramalipoor: yeah that would make more sense imo, because a dispute in nature "is" retryable towards network... nodes just need to sync up eventually

kasrakhosravi: closing this

PR #836, opened 2026-04-16
feat: WebSocket support and transport-agnostic event indexer

jleeh:
## Summary

Adds full WebSocket client and server support to eRPC, with a transport-neutral event-stream indexer underneath so future ingress/egress implementations (Kafka, webhooks, gRPC) reuse the same dedup, reorg, lifecycle, and fan-out machinery.

### Client-facing WebSocket server
- HTTP/WS share the same port; `ws://` / `wss://` on the project URL
- Regular JSON-RPC single and batch over WebSocket
- `eth_subscribe` / `eth_unsubscribe` with per-connection sub limits
- Auth and directives work identically on the upgrade request
- Per-subscription buffered writer goroutines with oldest-drop so slow clients never block upstream delivery

### Upstream WebSocket client (`clients/ws_json_rpc_client.go`)
- Persistent connection with exponential-backoff reconnect (1s → 30s)
- Ping/pong keepalive with read-deadline extension on pong
- Lazy initial connect — failed startup doesn't fail the upstream
- Pending-request tracking with ID normalisation for cross-format IDs
- Keyed `SetOnDisconnect` / `SetOnReconnect` so subscription management doesn't leak callbacks over long-lived connections

### Transport-neutral indexer (`indexer/`)
- `StreamEvent` / `IndexedEvent` types with a tagged-union `EventKind`
- `EventIngress` and `EventEgress` interfaces + a narrow `NetworkHandle` view
- Dedup, lifecycle classification (soft / finalized), per-source state-poller updates, and egress fan-out
- Refcounted filter subscriptions — N clients on the same filter share one upstream subscription
- Canonical-chain tracker: ring buffer of recent heads per network, `parentHash` continuity drives reorg detection, reorged-out logs are re-emitted with `"removed": true`
- Tunable via a new top-level `indexer:` config (`canonicalChainDepth`, `dedupWindowSize`)

### Adapters (`indexer/adapters/`)
- `wsupstream` — wraps `*clients.WsJsonRpcClient` as an `EventIngress`. Owns `eth_subscribe`/`eth_unsubscribe` RPCs and re-subscribes on reconnect.
- `wsclient` — wraps a client WS connection as an `EventEgress`. Per-sub notify buffer + writer goroutine, interest-index for O(1) routing.
- `nullingress` — fixture-driven ingress used in integration tests. A concrete second implementation that prevents `EventIngress` from silently drifting toward WS-specific assumptions.

### `erpc/` integration
- `SubscriptionManager` is a thin client-facing shim over the indexer. Holds per-connection `wsclient` adapters and routes `Subscribe` / `Unsubscribe` / `CleanupConnection` through indexer APIs.
- `networkHandle` adapter wires `FinalityDepth` and `SuggestLatestBlock` back into `*Network` and `EvmStatePoller`, preserving existing shared-counter semantics for latest / finalized block numbers.

### Supporting fixes bundled
- **Network-level head consistency** — latest and finalized block numbers are cross-pod monotonic via `CounterInt64SharedVariable` with forward-progress-only semantics. Fallback-group upstreams no longer inflate the shared block counter ahead of primaries, avoiding tag-translation misses.
- **NodeGroup block-head sharing** (new `nodeGroup` field on `UpstreamConfig`) — HTTP and WS upstreams that target the same backend can be grouped so a `newHeads` notification on the WS sibling fires `SuggestLatestBlock` on the HTTP sibling's state poller. Fixes `ErrUpstreamBlockUnavailable` false rejections on fast-block chains where a request hedged to the HTTP transport arrives before its own poller has caught up.
- **Shared-state refresh-fn timeout honours caller ctx** — `TryUpdateIfStale` now uses `max(fallbackTimeout, caller_ctx_remaining)` for the refresh goroutine rather than the remote-storage `fallbackTimeout` (3s default). The previous behaviour silently capped state-poller fetches at 3s regardless of the upstream's `failsafe.timeout`, producing a flood of `context deadline exceeded` warnings on slow chains with no config knob to raise the bound.
- **Per-request failover tiering** (new `failover.onDefaultsExhausted` on network / network-defaults) — the request loop tries default-group upstreams first and only advances to fallback-group upstreams if every default returned a retryable error within the same request. Previously the auto-applied `SelectionPolicy` handled this on a ~1-minute evaluation interval — too slow for request-level escalation. When enabled, the fallback-filtering auto-policy is suppressed so fallbacks remain visible to the request loop.
- **Metric cardinality** — `histogramLabelMode: compact` collapses high-cardinality labels on duration histograms for ~80% series reduction on large deployments.

## Docs

- New `/operation/websocket` page: client examples, subscription features, reorg handling, indexer configuration, server tuning
- New `failover` field documented in `/config/projects/networks` (per-network + networkDefaults) with cross-reference from `/config/projects/selection-policies`
- New `nodeGroup` field documented in `/config/projects/upstreams`
- Monitoring page documents `histogramLabelMode`
- Production page covers WS subscription memory profile and behaviour notes

## Test plan

- [x] 16 WebSocket integration tests in `erpc/ws_server_test.go` (basic RPC, batch, subscription lifecycle, recovery, dedup, rate limits, graceful shutdown, method filtering, regression guards) — all green
- [x] Unit tests for every indexer primitive: dedup windows, filter-key extraction, canonical-chain tracker, reorg emission, lifecycle classification
- [x] End-to-end integration test wiring a real `Indexer` to the null ingress
- [x] Local end-to-end verification against live public WebSocket endpoints on both a slow-block chain and a 1Hz fast-block chain: subscribe ack, live `newHeads` stream with parent continuity, concurrent multi-client fan-out identical across clients, `logs` subscription filter hashing through refcounted upstream sub, clean resource footprint (~0.1% CPU / 61 MB RSS idle)
- [x] Unit tests for NodeGroup sharing (`TestNetwork_NodeGroupSharesBlockHeadState`), sharedState ctx-deadline honouring (`TestCounterInt64_TryUpdateIfStale_FnTimeoutFromCtxDeadline`), and failover tiering (`TestTierUpstreamsByGroup`, `TestFailoverConfig_Enabled`, `TestNetworkConfig_SetDefaults_FailoverSkipsAutoSelectionPolicy`, `TestNetworkConfig_SetDefaults_FailoverInheritsFromDefaults`)
- [x] Observed-in-production for multiple days on a multi-pod deployment — no regressions versus the prior release; NodeGroup fix eliminated a recurring error class on a fast-block chain; failover fix eliminated tail timeouts on a slow-chain upstream by escalating to public fallbacks within the request


jleeh:
Still working on this and ironing out any issues as I roll it to more chains.

Found a deadlock in one of the upstream deps:
https://github.com/aramalipoor/failsafe-go/pull/1

Changed this branch to use my forked version in the meantime.

jleeh:
Following up here @aramalipoor and will say this PR is ready for review. Squashed the last bug yesterday, and running this in production with no issues. Got to say, this is a huge QoL improvement and the WS support is really good.

I'm a bit slammed over the next week or two, but will aim to get the conflicts resolved and branch up to date.

jleeh:
I take back what I said @aramalipoor, there's a memory leak and will get a fix up for it when I can.

aramalipoor:
> I take back what I said @aramalipoor, there's a memory leak and will get a fix up for it when I can.

It's all good, take your time this is a rather big feature. By the way the memory leak could be related to existing things as well, most of the previous memory leaks were related to Response and Result allocations/release.

jleeh:
Confident this is ready now @aramalipoor, some stats from us using this in our production env:

  **Production stats from running this branch:**

  Single eRPC deployment fronting all RPC traffic for internal services in one cluster.

  | Property | Value |
  |---|---|
  | Pods | 7 |
  | Image | `3d33933` |
  | Uptime on this image | ~36 h, 0 restarts |
  | EVM networks served | 82 |
  | Total upstreams configured | 555 (HTTP + WS) |
  | WS upstreams | 204 |
  | Avg upstreams per network | ~6.8 |

  **Traffic, last 24 h:**

  | Metric | Value |
  |---|---|
  | Network-level RPS, sustained | ~4,250 req/s |
  | Network-level RPS, peak 1 min | 5,012 req/s |
  | Upstream-level RPS, sustained | ~3,843 req/s |
  | Per-pod RPS | ~600 req/s |
  | Network success rate | 97.3% |
  | Latency p50 | 37 ms |

  **Top method mix:**

  | Method | Req/s |
  |---|---|
  | `eth_call` | 1,804 |
  | `eth_getBlockByNumber` | 1,529 |
  | `eth_getLogs` | 240 |
  | `eth_getTransactionCount` | 117 |
  | `eth_getBalance` | 114 |

  **Per-pod footprint:**

  | Metric | Value |
  |---|---|
  | Memory working set | avg 1.36 GiB (`GOMEMLIMIT=1500MiB`, cgroup cap 2 GiB) |
  | CPU | avg 0.88 cores |

  **Notes on the WS work specifically:**

  - WS client connections are long-lived; the indexer dedups subscribes — N clients each subscribing to `newHeads` on the same network result in one upstream `eth_subscribe`.
  - WS subscriptions span all 82 networks where the configured upstream advertises a WS endpoint (204 of 555 upstreams).

jleeh:
Raised a PR for the failsafe-go fork that fixes a memory leak in there:
https://github.com/aramalipoor/failsafe-go/pull/2

Although worth saying, this is already fixed in the upstream repo in the later versions, but this is a simple change without needing to pull in then all the diffs.

jleeh:
Ty for the failsafe merge, updated the PR with the updated dep.

aramalipoor:
@jleeh Thanks for the great work and deploying on real prod as well. Quick question about the indexer logic (I haven't dug deep into code yet):

1) Are we using http-based endpoints as well (with polling, like Ponder) for indexing besides websockets?
2) Where is the main place that data integrity checks are enforce, if any is enforced? e.g. if we make calls to Network.Forward (for poll-based indexing) that would include all the goodies and magic of erpc meaning that getBlockReceipts/getLogs would have nice good checks on them.
3) Are we doing "logs" subscriptions in this PR, and if yes which method are we relying on? Our pas experience shows doing getBlockReceipts is much more robust than getLogs. BTW if we're only doing blocks for now it's great, because maybe even better to keep the first version forcused.

jleeh:
Thanks, to answer the questions:

1. No HTTP based polling yet implemented, should be relatively easy to add as an adapter in the indexer that could be done outside of this PR once merged. Didn't want to change much around HTTP for potential regressions I didn't catch.
2. For WS subscription calls, they're sent through the normal `upstream.Forward()` pipeline so include all of the existing integrity checks. Although, for subscription events, they're not returned through that pipeline rather handled by dedup logic in the indexer which is reorg aware.
3. Yes, `newHeads`, `logs`, and `newPendingTransactions` supported. Agreed on getBlockReceipts, could be supported by the HTTP poller.

azgms:
Adding a comment to flag a focused review path for the last 5 commits on this branch — they're an independent failover-resiliency fix layered on top of the per-request failover-tiering work (commit `3a64586`) that's already in this PR.

### Commits to review (top of branch)

* `2973d4e` — fix(failover): close tracker-signal gap and add per-request fallback escape

That commit is the consolidation. It addresses three production incidents:

1. **Bitlayer scaledown, 2026-05-22** — `checkUpstreamBlockAvailability` rejections updated Prometheus counters but never called the health tracker. A primary gate-rejected on every request showed `errorRate=0` to the JS selectionPolicy, so the policy never cordoned the broken primary and fallbacks stayed cordoned.

2. **State-poller CB-open invisible to tracker** — `PollLatestBlockNumber`/`PollFinalizedBlockNumber` failures (broken pipe, 503) logged warn and returned. The CB counted them (half-open ⟷ open cycling), but CB-open short-circuits then bypass `tryForward` and the tracker stays starved of signal.

3. **B2 mainnet reth-stall, 2026-05-25** — primary stalled ~5099 blocks behind chain head. Non-retryable gate-skip (distance > `MaxRetryableBlockDistance`) left `lastErr` nil; the per-request escape hatch's `lastErr != nil` gate failed; client got `ErrUpstreamsExhausted` despite Ankr being able to serve.

### What the fix does

* `erpc/networks.go` — `handleBlockSkip` records `(request, failure)` on gate-skips (both retryable and non-retryable — non-retryable on THIS upstream doesn't mean don't try OTHER upstreams). Outer escalation loop wraps the inner per-upstream loop: when exhausted with retryable errors AND `failover.onDefaultsExhausted: true`, replace `upsList` with fallback-group upstreams (cordon ignored) and re-iterate with the selectionPolicy permit check bypassed.
* `architecture/evm/evm_state_poller.go` — record `(request, failure)` on `ErrFailsafeCircuitBreakerOpen` short-circuit (narrowed to CB-open so we don't double-count normal `tryForward` failures).
* `upstream/registry.go` — new `GetFallbackEscapeUpstreams(ctx, networkId, method)` returning fallback-group upstreams that are bootstrapped + CB-closed + method-allowed, intentionally ignoring `IsCordoned` (the cordon is what the escape bypasses).
* `common/request.go` — `escalatedToFallbacks atomic.Bool` flag prevents escalation loops across failsafe retries / hedges.
* `telemetry/metrics.go` — new `erpc_network_fallback_escape_total{project,network,category}` counter.

### Observability

`erpc_network_fallback_escape_total` is the canary. Steady-state expected zero (escape only fires on exhaustion; healthy primaries return on first iteration). Sustained non-zero during an outage is expected during the brief window before the selectionPolicy promotes fallbacks into `upsList` directly. Sustained non-zero in steady state indicates a regression or config drift.

### Tests

* `erpc/networks_failover_escape_test.go` — `TestFailover_GateSkipsAccumulateErrorRate` (tracker recording), plus `TestFailover_EscapeHatch` with 5 sub-tests: `EscapesToFallbackOnFirstFailingRequest`, `NoEscapeWhenPrimariesHealthy` (zero steady-state cost), `NoEscapeWhenFailoverDisabled` (operator opt-out), `OnlyEscalatesOncePerRequest` (no escalation loop), `EscapesOnNonRetryableGateSkip` (the B2 case). Each sub-test verified to FAIL without the fix and PASS with it.
* `upstream/registry_fallback_escape_test.go` — unit tests for `GetFallbackEscapeUpstreams`: filter-by-group, ignores-cordon, filter-by-IgnoreMethods, no-fallbacks-returns-empty.

### Verification in cluster

Image built from this commit was deployed as `docker.io/azgms/erpc:2973d4e` (temporary personal namespace because this branch isn't yet on `main`, so CI hasn't published to `linkpool/erpc`). During a Bitlayer scaledown reproduction test:
* 9/9 pods healthy, 0 restarts.
* During scaledown burst: 0 client-facing failures, fallbacks served traffic.
* B2 reth-stall request (the exact failing `eth_getBalance` from the 2026-05-25 incident) now succeeds via Ankr.

aramalipoor:
@jleeh @shpookas again thanks for this great work. We were wondering what is the state of this PR in your side?

We have few other customers that are working off your branch, and we would like to do few changes and get this in.

Are you OK with us taking this forward and create a new PR off of this one?

jleeh:
Hey @aramalipoor, sorry for not replying sooner.

In its current state without latest upstream, this PR is ready to be reviewed and merged. We've been running in production for months with minimal issues. The more recent fixes we've added are small issues we encountered in real workloads and patched.

aramalipoor:
@jleeh that's awesome! thanks for getting back, do you think you can resolve the conflicts with the latest main? Or do you prefer we branch off your PR and new PR with updates? (It would still preserve your contribution commits though) 🙌

jleeh:
Sure @aramalipoor, we will do that. Was just talking with the team and we did find a new WS issue yesterday around sockets being closed. Will look at fixing that, testing and then working on getting the branch up to date.

aramalipoor:
An interesting topic around this rebase/merge with main would be teh ChainView component in the integrity module which is almost a mini indexer on its own, in final vision probably I'd merge them (e.g. taking the event indexer as source of truth for that). If it was easier to do this in rebase let's do it, otherwise we can keep both mechanisms separately if that's easier for now, I'll do a follow up PR for merging them.

PR #847, opened 2026-04-23
fix: wrap permanent init failures with NewTaskFatal

kasrakhosravi:
## Summary

Bootstrap initialization can fail for two kinds of reasons, and the Initializer (\`util/initializer.go\`) should treat them differently:

**Permanent** (deterministically won't self-heal on retry):
- chainId mismatch (configured != detected) — misconfiguration
- chainId parse error — upstream returned a non-numeric value

**Transient** (may self-heal when the provider recovers):
- \`EvmGetChainId\` RPC failure — network blip, provider outage, rate limit during startup, TLS handshake race

Previously all four shared a single error type without \`IsTaskFatal\`, so every failure (permanent or not) was retried indefinitely — a misconfigured upstream hung dependent requests for ~30s per retry cycle.

## Why not a blanket fix on the error type

The first iteration of this PR added \`IsTaskFatal()\` returning true on \`ErrUpstreamClientInitialization\` directly. That was too aggressive: it would have made transient RPC failures during chainId detection permanently fatal, so a provider outage at startup would have silently killed an upstream until the process restarted. Thanks to the review feedback on that — caught before merge.

## What this PR does instead

Keep \`ErrUpstreamClientInitialization\` semantically neutral. Wrap only the two permanently-broken call sites in \`upstream/upstream.go\` with the existing \`common.NewTaskFatal()\` helper:
- chainId parse error
- chainId mismatch

Leave the RPC-failure path unwrapped so the auto-retry loop can recover once the provider returns.

## Test plan

- [x] Bare \`ErrUpstreamClientInitialization\` must NOT satisfy \`IsTaskFatal\` (transient-friendly).
- [x] \`NewTaskFatal\`-wrapped version must satisfy \`IsTaskFatal\` AND stay reachable via \`errors.As(&ErrUpstreamClientInitialization)\`.
- [x] Integration: wrapped form stops retries after 1 attempt, initializer enters \`StateFatal\`.
- [x] Integration: unwrapped form keeps retrying (>5 attempts in a 150ms window), initializer does NOT enter \`StateFatal\` — the negative-control test that would have caught the over-aggressive v1.
- [x] \`./common/...\`, \`./util/...\`, \`./upstream/...\` suites pass.

---

> [!NOTE]
> **Medium Risk**
> Changes bootstrap retry/fatal behavior for upstream initialization, which can affect service startup and error surfacing if call sites misclassify errors as fatal or transient.
> 
> **Overview**
> Adjusts upstream feature detection to distinguish *transient* vs *permanent* upstream initialization failures: RPC/network chainId detection errors remain retryable, while non-numeric chainId responses and configured/detected chainId mismatches are now wrapped with `common.NewTaskFatal()` so `util.Initializer` stops retrying and enters a fatal state.
> 
> Adds tests that pin this contract: `ErrUpstreamClientInitialization` is not fatal by default, becomes fatal when wrapped, and the real initializer retries transient causes but stops immediately on fatal-wrapped ones. Also documents in `common/errors.go` why `ErrUpstreamClientInitialization` intentionally does not implement `IsTaskFatal` directly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1793265287fa7ee5d283d012daf50a6ddda588d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/common/errors.go
+++ b/common/errors.go
@@ -750,6 +750,14 @@ type ErrUpstreamClientInitialization struct {
 	BaseError
 }
 
+// Note: ErrUpstreamClientInitialization deliberately does not implement
+// IsTaskFatal — the same constructor is used for both permanent
+// (chainId-mismatch, parse-error, unsupported-type) and transient (RPC/network
+// failure during chainId detection) causes. Call sites that know the cause is
+// permanent wrap with common.NewTaskFatal() so the Initializer stops retrying.
+// Leaving this unimplemented keeps transient failures retryable — a provider
+// outage during startup should be recoverable once the provider returns.
+
 var NewErrUpstreamClientInitialization = func(cause error, upstream Upstream) error {
 	return &ErrUpstreamClientInitialization{
 		UpstreamAwareError: UpstreamAwareError{
--- a/common/errors_task_fatal_test.go
+++ b/common/errors_task_fatal_test.go
@@ -1,0 +1,136 @@
+package common
+
+import (
+	"context"
+	"errors"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/erpc/erpc/util"
+	"github.com/rs/zerolog"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// buildUpstreamInitErr constructs a bare ErrUpstreamClientInitialization with
+// the given cause — the shape produced by upstream.go for the RPC-failure path.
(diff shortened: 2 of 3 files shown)


aramalipoor:
This ErrUpstreamClientInitialization could happen if the upstream is temporarily down, is that correct? In such case we should either have a different error, or have a way to pass down the "cause" and make the IsFatal() conditional

PR #857, opened 2026-04-30
fix: short-circuit on agreed empty results when no competing group

kasrakhosravi:
## Problem

The `unassailable_lead` short-circuit rule only fires when the winning group is `ResponseTypeNonEmpty`. When the agreement threshold is met by empty responses (e.g. `eth_getLogs` on a block with no matching events), consensus blocks on `responseChan` waiting for all `maxParticipants` to return — even when one of them is slow or stuck on a degraded upstream. The wait runs to the outer network-level failsafe timeout, turning a sub-second request into a tens-of-seconds user-facing failure.

Reproduces consistently in production: a single chain configured with `consensusConfig(threshold=1, max=3, preferNonEmpty=true)` for `eth_getLogs` produces 2,000+ such hangs per hour. Trace shows all `Upstream.Forward` spans completing in <200ms while `Consensus.CollectResponses` runs for the full network timeout floor. Independent of any specific upstream slowness — it's the consensus aggregator itself blocking.

## Fix

Allow short-circuit when the winning group is empty **AND** no other valid group could catch up or create a tie. When a competing non-empty group is present, still wait for all participants — that preserves `preferNonEmpty` semantics (don't prematurely pick empty over a potential non-empty winner).
```

## Tests

- New: `TestConsensus_ShortCircuit_EmptyResponses_StuckParticipant` — two participants return matching empty responses fast, third blocks indefinitely; asserts caller is released in <500ms. Fails without the fix.
- All existing `consensus/...` tests pass unchanged. The existing `TestConsensus_ShortCircuit_CallerGetsWinnerBeforeSlowParticipants` covers the analogous non-empty case.

## Test plan

- [x] `go test ./consensus/...` (passes locally)
- [x] Build all packages clean

<!-- CURSOR_SUMMARY -->
---

> [!NOTE]
> **Medium Risk**
> Adjusts consensus short-circuiting behavior, which can change request completion timing and winner selection in edge cases involving empty results and slow participants. Guardrails and a targeted regression test reduce risk, but it impacts core consensus flow.
> 
> **Overview**
> Fixes `unassailable_lead` short-circuiting to allow **agreed empty results** (e.g. `eth_getLogs` returning `[]`) to return early when the agreement threshold is met and **no other valid response group exists** that could catch up or tie.
> 
> Updates the rule’s condition to treat empty winners differently: it now blocks short-circuiting if any competing valid group is present (preserving preference semantics), and adds a regression test (`TestConsensus_ShortCircuit_EmptyResponses_StuckParticipant`) that ensures consensus doesn’t wait on a stuck participant once empty agreement is reached.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e9ea5adb2f3388215ad8ad79568bd831a6c3b4f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>
<!-- /CURSOR_SUMMARY -->

### DIFF
--- a/consensus/executor_test.go
+++ b/consensus/executor_test.go
@@ -719,3 +719,58 @@ func TestRecordMetricsAndTracing_NilAnalysis_DoesNotPanic(t *testing.T) {
 		e.recordMetricsAndTracing(req, time.Now(), result, nil /* analysis */, labels, span)
 	})
 }
+
+// emptyResponse returns an empty JSON-RPC array result, the canonical shape for
+// eth_getLogs on a block with no matching events. Used to exercise the empty-
+// result branch of the unassailable_lead short-circuit rule.
+func emptyResponse() *common.NormalizedResponse {
+	jrpc, _ := common.NewJsonRpcResponse(1, []interface{}{}, nil)
+	return common.NewNormalizedResponse().WithJsonRpcResponse(jrpc)
+}
+
+// TestConsensus_ShortCircuit_EmptyResponses_StuckParticipant verifies that when
+// the agreement threshold is met by empty responses (e.g. eth_getLogs on a
+// block with no matching events) and no competing non-empty group exists,
+// consensus short-circuits without waiting for the slow/stuck remaining
+// participants. Without this, consensus blocks on responseChan until the outer
+// failsafe timeout fires — the production pattern observed on Monad eth_getLogs.
+func TestConsensus_ShortCircuit_EmptyResponses_StuckParticipant(t *testing.T) {
+	logger := zerolog.New(zerolog.NewTestWriter(t))
+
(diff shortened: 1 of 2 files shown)


aramalipoor:
@kasrakhosravi this would basically defeat the purpose of "prefer non-empty" flag, no? Maybe the way to prevent a slow upstream impacting consensus is a "max wait" config (or maybe two maxWaitOnEmptyResult and maxWaitOnNonEmptyResult... with better naming if possible).

This way we set a 100ms or 500ms wait time, and avoid such edge-cases, until the bad upstream recovers, or existing upstreams end up syncing their data and not serving empty anymore.

PR #878, opened 2026-05-12
fix: exclude hedge-cancelled attempts from upstream error rate denominator

andreclaro:
## Problem

When a hedged request loses the race, `RecordUpstreamRequest` increments `RequestsTotal` but `RecordUpstreamFailure` is skipped — `ErrCodeUpstreamHedgeCancelled` is correctly not treated as an upstream fault. This creates an asymmetry: hedge-cancelled attempts inflate the denominator of `ErrorRate`, `ThrottledRate`, and `MisbehaviorRate` without contributing to the numerator.

During periods of heavy hedging, a degraded upstream can appear significantly healthier than it is. If 50% of requests to an upstream are hedge-cancelled, its observed error rate is halved — suppressing the signal the selection policy uses to route traffic away from it.

## Solution

Add `HedgeCancelledTotal` to `TrackedMetrics` and call `RecordUpstreamHedgeCancelled` from `recordHedgeDiscard` in `networks.go` — the single unambiguous point where a hedge loss is definitively known (`hedges > 0 && ErrCodeEndpointRequestCanceled`).

Rate calculations subtract `HedgeCancelledTotal` from `RequestsTotal` via a new `effectiveRequests()` helper. `RequestsTotal` is preserved as a true throughput counter so observability is not reduced — `hedgeCancelledTotal` is also exposed in the health JSON endpoint.

## Changes

- `common/upstream.go` — add `RecordUpstreamHedgeCancelled` to `HealthTracker` interface
- `common/upstream_fake.go` — no-op stub
- `health/tracker.go` — `HedgeCancelledTotal` field, `effectiveRequests()` helper, updated rate calculations, `Reset`, `MarshalJSON`, new tracker method
- `erpc/networks.go` — call `u.Tracker().RecordUpstreamHedgeCancelled(u, method)` in `recordHedgeDiscard`
- `health/tracker_test.go` — new test verifying corrected denominator

## Testing

```
make test-fast
```

New test: 10 requests, 5 hedge-cancelled, 3 real failures → `ErrorRate() == 0.6` (3/5), not `0.3` (3/10).

### DIFF
--- a/common/upstream.go
+++ b/common/upstream.go
@@ -27,6 +27,7 @@ type HealthTracker interface {
 	RecordUpstreamMisbehavior(up Upstream, method string)
 	RecordUpstreamRequest(up Upstream, method string)
 	RecordUpstreamFailure(up Upstream, method string, err error)
+	RecordUpstreamHedgeCancelled(up Upstream, method string)
 	Cordon(upstream Upstream, method string, reason string)
 	Uncordon(upstream Upstream, method string, reason string)
 }
--- a/common/upstream_fake.go
+++ b/common/upstream_fake.go
@@ -275,6 +275,10 @@ func (t *FakeHealthTracker) RecordUpstreamFailure(up Upstream, method string, er
 	// No-op for testing
 }
 
+func (t *FakeHealthTracker) RecordUpstreamHedgeCancelled(up Upstream, method string) {
+	// No-op for testing
+}
+
 func (t *FakeHealthTracker) Cordon(upstream Upstream, method string, reason string) {
 	// No-op for testing
 }
--- a/erpc/networks.go
+++ b/erpc/networks.go
@@ -1226,6 +1226,7 @@ func (n *Network) recordHedgeDiscard(
 	hedges int,
 ) {
 	ulg.Debug().Err(err).Msgf("discarding hedged request to upstream")
+	u.Tracker().RecordUpstreamHedgeCancelled(u, method)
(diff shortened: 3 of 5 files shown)


common/upstream.go
```diff
@@ -27,6 +27,7 @@ type HealthTracker interface {
 	RecordUpstreamMisbehavior(up Upstream, method string)
 	RecordUpstreamRequest(up Upstream, method string)
 	RecordUpstreamFailure(up Upstream, method string, err error)
+	RecordUpstreamHedgeCancelled(up Upstream, method string)
```
aramalipoor, line 30: We can skip them when calling "RecordUpstreamFailure" we already have a place where we list "ignored" errors: https://github.com/erpc/erpc/blob/main/upstream/upstream.go#L527

andreclaro: the goal here is not just to ignore the error but also to exclude the request from the rate denominator (_RequestsTotal_).

aramalipoor: Good point that makes sense. Should we actually do this for other skips as well? Maybe a new RecordUpstreamSkip (RecordUpstreamLose)? Notable the link I shared above.

Another thought is: Should we track such "hedge loss" as its own signal of "this upstream is getting slow, therefore put it lower in the list"? If an upstream is constantly losing hedges shouldn't we push it down so other upstreams come higher?

A simple approach to go about this is to consider hedge-loss as a failure (not even skipping) so basically removing some existing logic vs adding more.

PR #879, opened 2026-05-12
feat: expose blockHeadLagSeconds and finalizationLagSeconds in selection policy

andreclaro:
## Summary

`blockHeadLag` and `finalizationLag` in the selection policy eval function are
raw block-count deltas. Policy authors who want to reason in time units must
independently know the chain's block time — which varies by network and changes
post-merge/upgrade.

This adds `blockHeadLagSeconds` and `finalizationLagSeconds` to the upstream
metrics object passed to `evalFunction`, derived from the existing EMA-based
block time estimate (`GetNetworkBlockTime`). The two new fields make time-based
lag thresholds expressible directly in policy scripts without any chain-specific
knowledge in the config.

## Changes

- In `evaluateMethod`, call `GetNetworkBlockTime(networkId)` once per evaluation
  tick (network-level EMA, shared across all upstreams).
- Derive `blockHeadLagSeconds = blockHeadLag * blockTime.Seconds()` and
  `finalizationLagSeconds = finalizationLag * blockTime.Seconds()` per upstream.
- Both fields are `0.0` when the block time EMA is not yet warm (fewer than 3
  samples), matching the existing behaviour of `GetNetworkBlockTime` returning 0.
- Existing `blockHeadLag` and `finalizationLag` fields are unchanged.

## Example policy usage

```js
// exclude upstreams more than 120 seconds behind head
upstreams.filter(u => u.metrics.blockHeadLagSeconds <= 120)
```

## Tests

`TestPolicyEvaluatorLagSeconds` covers three cases:

- **BlockHeadLagSecondsZeroBeforeEMAWarm**: with a cold EMA, `blockHeadLagSeconds`
  is 0 regardless of block lag, so all upstreams pass any threshold.
- **BlockHeadLagSecondsFiltering**: after warming the EMA to ~2s block time, an
  upstream 6 blocks behind (~12s) is cordoned by a `< 10s` threshold, then
  uncordoned once it catches up.
- **FinalizationLagSecondsFiltering**: same pattern for `finalizationLagSeconds`.

Note: `RecordUpstreamRequest` is called for both upstreams before seeding lag
values. `SetLatestBlockNumber` writes into `upstreamsByNetwork`, which is only
populated once `getUpsMetrics` has been called for an upstream — without this,
lag writes are dropped and the assertions would not hold.

## Notes

- No new `TrackedMetrics` fields — the seconds values are derived at eval time,
  not stored. The block time is a network-level quantity; embedding it in the
  per-upstream struct would be a layering violation and could go stale between
  lag updates and block-time EMA ticks.
- Prometheus users can already compute this via
  `erpc_upstream_block_head_lag * on(network) group_left() (erpc_network_dynamic_block_time_milliseconds / 1000)`.

### DIFF
--- a/docs/pages/config/projects/selection-policies.mdx
+++ b/docs/pages/config/projects/selection-policies.mdx
@@ -254,9 +254,19 @@ export type UpstreamMetrics = {
     // (e.g. 10 blocks ≈ 120s on Ethereum, ≈ 2.5s on Arbitrum).
     blockHeadLag: number;
 
+    // Seconds this upstream is behind the network's highest known block head,
+    // derived from blockHeadLag × the EMA-estimated block time.
+    // Zero until the block-time EMA has collected at least 3 samples.
+    blockHeadLagSeconds: number;
+
     // Number of finalized blocks this upstream is behind the network's highest known
     // finalized block. Block-number delta, not seconds.
     finalizationLag: number;
+
+    // Seconds this upstream is behind the network's highest known finalized block,
+    // derived from finalizationLag × the EMA-estimated block time.
+    // Zero until the block-time EMA has collected at least 3 samples.
+    finalizationLagSeconds: number;
 };
 
 // Method is either `*` (all methods) or a specific method name.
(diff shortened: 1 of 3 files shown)


aramalipoor (approved):

andreclaro:
Already added! :)

PR #882, opened 2026-05-13
feat: state-readiness probe to detect head-vs-trie race

kasrakhosravi:
## Problem

For state-read methods (`eth_getBalance`, `eth_getCode`, `eth_getStorageAt`, `eth_getTransactionCount`, `eth_call`, …), an upstream's reported head block isn't a sufficient proof that its state DB is consistent at that block. Fast-ingesting nodes — especially load-balanced ones — can advance their head pointer to block N before their state trie has committed slots for N, and silently return zero-valued or stale-prev-block results.

The retry shortcut at [`upstream/failsafe.go:703`](upstream/failsafe.go) trusted the head pointer (`reason="block_available_but_empty"` → `retry=false`), so a wrong-empty result for a state read was returned to the client without any retry against other upstreams.

## What this PR adds

A new availability primitive — `AvailbilityConfidenceStateReady` — that requires the upstream's state DB to be **verifiably consistent** at the requested block, not just observed at its head. Backed by a per-network state-readiness probe that the state poller runs alongside its existing cycle.

The probe is a single named strategy with typed inputs (no expression DSL):

```yaml
network:
  evm:
    chainId: 137
    stateProbe:
      strategy: changingStorage
      address: "0x..."   # contract with a high-activity slot
      slot:    "0x..."   # slot that changes on most blocks
```

`changingStorage` reads `eth_getStorageAt(address, slot, latestBlock)` each cycle and advances `stateReadyBlock` iff the result is **non-empty** AND **differs from the previous successful probe**. The first check catches `0x0`/`"0x"`/`null` returned at the advertised head; the second catches the silent-stale-prev-block case where a node serves the value from block N-1 while claiming to be at block N. Both assertions live in the strategy implementation, not in user config.

Future strategies can land as code (same shape, no config-language churn).

## Behavior

- **No behavior change without opt-in.** When `EvmNetworkConfig.StateProbe` is nil, `StateReadyBlock()` returns `LatestBlock()`, the new confidence behaves identically to `BlockHead`, and the retry shortcut decides the same outcome as before.
- **Wired at the leak site.** [`upstream/failsafe.go:703`](upstream/failsafe.go) — the `block_available_but_empty` shortcut now requests `StateReady` confidence when `evm.IsStateReadMethod(method)`. Other methods keep using `emptyResultConfidence`.
- **Cold-start permissive.** First probe with a non-empty result advances `stateReadyBlock` without waiting for a second cycle.
- **Failure threshold (default 10).** Consecutive probe failures disable the probe for that upstream and fall back to `LatestBlock` semantics, so a misconfigured probe can't black-hole an upstream.

## What this PR does NOT change

- The upstream-selection gate at [`erpc/networks.go:1281`](erpc/networks.go) stays on `BlockHead`. Promoting it to `StateReady` for state-read methods is a follow-up — the tolerance policy needs its own design discussion.

## Tests

- `architecture/evm/state_probe_test.go` — full assertion truth table for `changingStorage` plus a 3-cycle scenario test.
- `upstream/upstream_state_ready_test.go` — `EvmAssertBlockAvailability` rejects under `StateReady` when state-readiness lags head, accepts under `BlockHead` (legacy parity), and the no-probe-configured fallback.
- `common/config_state_probe_test.go` — load-time config validation.
- `architecture/evm/util_state_read_test.go` — `IsStateReadMethod` coverage.

`go test ./common/... ./architecture/evm/... ./upstream/...` passes; `go vet ./...` clean.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new block-availability confidence used in retry/availability decisions for state-read JSON-RPC calls; mistakes could incorrectly suppress retries or mark upstreams stale despite backward-compatible fallbacks when the probe is unconfigured/disabled.
> 
> **Overview**
> Adds an opt-in *state-readiness* signal to detect head-vs-trie races where an upstream reports block N at its head but cannot yet serve consistent state at N.
> 
> The EVM state poller now optionally runs a per-network `stateProbe` (initial `changingStorage` strategy) to advance a new `StateReadyBlock`, exposes probe diagnostics, and disables the probe after a configurable consecutive-failure threshold.
> 
> Upstream block availability and the retry shortcut for empty state-read responses now consult the new `AvailbilityConfidenceStateReady` for methods classified by `evm.IsStateReadMethod`, with explicit fallback to legacy `LatestBlock`/`BlockHead` semantics when no probe is configured. Extensive unit tests cover config validation, probe classification/dispatch, and `EvmAssertBlockAvailability` behavior under head-vs-state divergence.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 47f76ee6f6566c66d6fb805fd110b98f20c1ec5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/architecture/evm/evm_state_poller.go
+++ b/architecture/evm/evm_state_poller.go
@@ -92,6 +92,23 @@ type EvmStatePoller struct {
 	earliestSchedulerStarted     map[common.EvmAvailabilityProbeType]bool
 	earliestInitialDetectionDone map[common.EvmAvailabilityProbeType]bool // tracks if THIS instance did initial detection
 	earliestMu                   sync.RWMutex
+
+	// State-readiness probe tracking. stateReadyBlock is updated only when the
+	// configured strategy (EvmNetworkConfig.StateProbe) passes its assertions
+	// at the upstream's current latest block. When no probe is configured,
+	// StateReadyBlock() falls through to LatestBlock(). Protected by stateMu
+	// alongside the other detection-state fields.
+	//
+	// lastStateProbeResult holds the JSON-RPC result string from the most
+	// recent successful probe. It's the "prev" the changingStorage strategy
+	// compares against to detect silent-stale-prev-block staleness.
+	stateReadyBlock          int64
+	stateProbeFailureCount   int
+	stateProbeSuccessfulOnce bool
+	skipStateProbeCheck      bool
+	stateProbeDetectionIssue string
+	lastStateProbeResult     string
+	stateProbeInProgress     sync.Mutex
 }
 
 func NewEvmStatePoller(
@@ -343,6 +360,20 @@ func (e *EvmStatePoller) Poll(ctx context.Context) error {
 		e.initializeEarliestBlockDetectionAndStartScheduler(ctx)
 	}()
 
+	// State-readiness probe (only when configured on the network). Runs
+	// concurrently with latest/finalized poll — but reads latestBlockShared,
+	// so the first cycle may see block=0 and no-op. The next cycle will catch up.
+	wg.Add(1)
+	go func() {
+		defer wg.Done()
+		if e.networkStateProbeConfig() == nil {
+			return
+		}
+		if _, err := e.PollStateReady(ctx); err != nil {
+			e.logger.Debug().Err(err).Msg("state-readiness probe returned error")
+		}
+	}()
+
 	wg.Wait()
 
 	if len(errs) > 0 {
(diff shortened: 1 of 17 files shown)


aramalipoor:
@kasrakhosravi what are your thoughts on a free-form probe function, similar to selection policy? The upside is any network can have it's own bespoke logic, and can be as nuanced as it's needed, the downside is overhead of such check and complexity of code.

A middle-ground could be a basic interpolation logic and letting user define method/params/assertions and then we can interpolate the "{latestBlockNumber}" or "{finalizedBlockNumber}" etc there. Or maybe just the assertions to be free-form.

The most pragmatic solution is we hard-code these probes ourselves, and just ask user for input. e.g. "walletBalanceProbe" or "smartContractCodeProbe" etc. but not sure how accurate that would be, a node could silently return balance of previous block. This Tenderly edge-case of return 0x0 for a balance (and not falling back to prev block value) is only 1 incident.

PR #886, opened 2026-05-13
fix: exclude hedge attempts from per-upstream rate counters

aramalipoor:
## Summary

Resolves the per-upstream rate suppression described in #878 by keeping hedge attempts out of **both** `RequestsTotal` and `ErrorsTotal` at the tracker layer. The latency-based scoring (`ResponseQuantiles`) is left to handle the "slow but functional" upstream demotion case on its own — no double-punishment.

Closes #878.

## Background

When failsafe spawns a hedged parallel attempt to a secondary upstream, the upstream layer used to bump `RequestsTotal` for it. Cancellations (the common outcome — primary or hedge wins, the other is cancelled) were already in the tracker's skip list, so they didn't bump `ErrorsTotal`. Result: hedge attempts inflated the denominator of `ErrorRate` / `ThrottledRate` / `MisbehaviorRate` without contributing to the numerator, suppressing the signal a degraded upstream should send to selection.

## Alternatives considered (and rejected)

| Approach | Trade-off | Outcome |
|---|---|---|
| **Count hedge-loss as a failure** (raise the numerator) | Stacks ErrorRate on top of the latency penalty `ResponseQuantiles` already applies; double-punishes the slow-but-functional case. ([William's concern](https://github.com/erpc/erpc/pull/878#discussion_r3233186251); [aram's reinforcement](https://github.com/erpc/erpc/pull/878#discussion_r3235366204)) | Rejected |
| **Subtract hedge cancellations from the denominator** (#878's original) | Equivalent outcome but introduces new bookkeeping (`HedgeCancelledTotal`, `effectiveRequests()`). | Rejected as the heavier path |
| **Exclude hedges from both sides of the ratio** (this PR) | Keeps the rate honest; latency already feeds scoring. | Picked |

## Design

- New context key [`common.HedgeAttemptKey`](common/request.go) — bool flag carried on the per-attempt context.
- Set at the network layer in [`erpc/networks.go`](erpc/networks.go) where failsafe's hedge counter lives (the upstream's own failsafe has no hedge policy, so `exec.Hedges()` at the upstream layer is always zero — context plumbing is the canonical signal).
- Read at the upstream layer in [`upstream/upstream.go`](upstream/upstream.go) to gate both `RecordUpstreamRequest` and `RecordUpstreamFailure`.
- The `ErrCodeEndpointRequestCanceled` early-return branch in the upstream-layer error chain is preserved — at this layer a cancellation could be a hedge loss *or* a client disconnect, so don't blame the upstream either way. Hedge discards stay observable via `MetricNetworkHedgeDiscardsTotal` at the network layer.

| Counter | Primary attempt | Hedge attempt |
|---|---|---|
| `RequestsTotal` | ✓ | **excluded** |
| `ErrorsTotal` | ✓ (real errors) | **excluded** |
| `ResponseQuantiles` (success only, filtered by `isSuccess`) | ✓ | ✓ — successful hedge wins still feed latency scoring |
| `MetricNetworkHedgeDiscardsTotal` (Prometheus, network layer) | n/a | ✓ — unchanged |

## Test coverage

| Goal | Test |
|---|---|
| Cancellation errors (hedge loss or client disconnect) don't count as upstream failures | [`TestRecordUpstreamFailure_IgnoresHedgeCancellationErrors`](health/tracker_test.go) |
| Full matrix of skip codes still ignored (RequestCanceled / HedgeCancelled live here too) | [`TestRecordUpstreamFailure_AllSkipCodesIgnored`](health/tracker_test.go) — 9-case matrix |
| Cancellations don't bleed into `ErrorRate` / `ThrottledRate` / `MisbehaviorRate` | [`TestRates_RealErrorsOnlyAffectRates`](health/tracker_test.go) |
| Hedge cancellations don't degrade selection score | [`TestUpstreamsRegistry_HedgeCancellationDoesNotDegradeScore`](upstream/registry_test.go) |
| Real errors *do* degrade, hedge cancellations don't | [`TestUpstreamsRegistry_RealErrorsDegradeButHedgeCancellationsDont`](upstream/registry_test.go) |
| End-state: slow upstream IS demoted, but via latency, not via ErrorRate | [`TestUpstreamsRegistry_SlowUpstreamDemotedByLatency`](upstream/registry_test.go) |
| End-to-end: hedge attempt that wins doesn't tick `RequestsTotal`, losing primary's cancellation doesn't tick `ErrorsTotal` | [`TestNetwork_HedgeAttemptsExcludedFromTrackerCounters`](erpc/networks_hedge_test.go) — 2 scenarios (primary-wins + hedge-wins) |

## Test plan

- [x] `go test ./health/...` — passes
- [x] `go test ./upstream/...` — passes
- [x] `go test ./common/...` — passes
- [x] `go test ./clients/...` — passes
- [x] Targeted erpc-package hedge tests (`TestNetwork_HedgePolicy`, `TestHedgeConsensus_*`, `TestHttpServer_HedgedRequests`, `TestUpstreamSelectionWithHedgeAndRetry`, `TestNetworkIntegrity_Hedge*`, and the new `TestNetwork_HedgeAttemptsExcludedFromTrackerCounters`) — passes
- [x] `go vet ./...` — clean
- [ ] `make test-fast` on CI

### DIFF
--- a/health/tracker.go
+++ b/health/tracker.go
@@ -540,7 +540,12 @@ func (t *Tracker) RecordUpstreamFailure(up common.Upstream, method string, err e
 	// - Unsupported: capability, not quality
 	// - CapacityExceeded: remote 429, already penalized via ThrottledRate
 	// - ClientSideException: user sent a bad request, not upstream's fault
-	// - RequestCanceled / HedgeCancelled: hedge lost the race, not an upstream fault
+	// - RequestCanceled / HedgeCancelled: indistinguishable from a client
+	//   disconnect at this layer (both surface as context cancellation). Hedge
+	//   attempts are excluded from RequestsTotal / ErrorsTotal at the call
+	//   site in upstream.tryForward, so cancellations only reach here for
+	//   primary attempts — where they almost always mean "client gave up,"
+	//   not "upstream failed."  Slowness is already captured by ResponseQuantiles.
 	if common.HasErrorCode(
 		err,
 		common.ErrCodeEndpointExecutionException,
--- a/architecture/evm/eth_query_test.go
+++ b/architecture/evm/eth_query_test.go
@@ -63,7 +63,7 @@ func (u *queryTestUpstream) Logger() *zerolog.Logger {
 }
 func (u *queryTestUpstream) Vendor() common.Vendor         { return nil }
 func (u *queryTestUpstream) Tracker() common.HealthTracker { return nil }
-func (u *queryTestUpstream) Forward(ctx context.Context, nq *common.NormalizedRequest, byPassMethodExclusion bool) (*common.NormalizedResponse, error) {
+func (u *queryTestUpstream) Forward(ctx context.Context, nq *common.NormalizedRequest, byPassMethodExclusion bool, isHedgeAttempt bool) (*common.NormalizedResponse, error) {
 	return nil, nil
 }
 func (u *queryTestUpstream) Cordon(method string, reason string)   {}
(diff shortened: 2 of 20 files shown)


health/tracker.go
```diff
@@ -550,8 +553,6 @@ func (t *Tracker) RecordUpstreamFailure(up common.Upstream, method string, err e
 		common.ErrCodeEndpointUnsupported,
 		common.ErrCodeEndpointCapacityExceeded,
 		common.ErrCodeEndpointClientSideException,
-		common.ErrCodeEndpointRequestCanceled,
```
andreclaro, line 553: why removing `ErrCodeEndpointRequestCanceled`? In this case, a client disconnecting mid-request will affect upstream error rate, right?

aramalipoor: Yes that's a good call, I just revised the logic and did exclude hedge cancellations from requests/failures, because slow upstreams are already punished by latency score no need to double punish due to hedge loss as well.

PR #895, opened 2026-05-20
fix: reject emptyish results as winners for non-accept methods

kasrakhosravi:
The hedge keep predicate currently accepts any non-object-null response, including `{"result": null}`. For methods in `DefaultMarkEmptyAsErrorMethods` (`eth_getBlockByNumber`, `eth_getTransactionByHash`, `eth_getTransactionReceipt`, …) `null` means "this upstream doesn't have the data yet", not "doesn't exist". A fast empty leg therefore wins the race and cancels in-flight siblings that could have returned the actual data, then the retry layer re-fans-out the whole rotation.

This change mirrors the upstream-sweep policy (`networks.go:608-625`) inside the hedge keep: when the response is emptyish and the method is not in `emptyResultAccept`, reject so the race continues. Methods that legitimately return empty (the accept list expanded in #894) are unchanged — fast empty still short-circuits the hedge.

When every leg ends up emptyish for a non-accept method, `RunHedged` already returns the last non-kept result, so terminal behaviour is preserved; the retry layer remains responsible for deciding whether to fan out again.

## Tests

- `TestHedge_EmptyishLosesToNonEmpty_ThreeUpstreams` — direct bug isolation: 3 upstreams across 2 fan-outs, one sweep ends emptyish, the other has the block; the block must win. Fails against the previous behaviour.
- Per-method coverage for `eth_getBlockByNumber`, `eth_getTransactionByHash`, `eth_getTransactionReceipt`.
- Accept-list regression guards (`eth_getLogs`, `eth_getBalance`) asserting fast empty still short-circuits the hedge in well under the hedge delay.
- All-emptyish-legs surfaces a response to the caller (no hang).
- Non-empty primary baseline — hedge does not need to fire when the primary is already valid.

---

> [!NOTE]
> **Medium Risk**
> Changes the hedge race winner selection for JSON-RPC calls, which can alter latency and upstream selection behavior for block/tx lookups at the network layer. While scoped and covered by new tests, it affects a core execution path under load.
> 
> **Overview**
> Prevents `hedge` fan-out from declaring `{"result": null}` / other *emptyish* responses as the race winner for methods not in the `emptyResultAccept` allowlist, so a fast empty leg no longer cancels slower siblings that may return real block/tx data.
> 
> Adds a dedicated test suite validating the new hedge keep behavior across common lookup methods (`eth_getBlockByNumber`, `eth_getTransactionByHash`, `eth_getTransactionReceipt`), the all-legs-empty terminal case, and regressions ensuring accept-listed methods (`eth_getLogs`, `eth_getBalance`) still short-circuit quickly on legitimate empty results.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0f56890584b983f6e19052048cde35e119c77f51. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/erpc/network_executor.go
+++ b/erpc/network_executor.go
@@ -538,6 +538,35 @@ func (e *networkExecutor) runHedge(
 		if r == nil || r.IsObjectNull(ctx) {
 			return false
 		}
+		// Mirror the upstream-sweep empty-result policy so a fast
+		// {"result": null} from one hedge leg does not cancel siblings
+		// that may still return real data. When the method legitimately
+		// returns empty (eth_getLogs, eth_call, point state reads, …)
+		// the method is in emptyResultAccept and we keep the fast empty
+		// winner — preserving prior behaviour.
+		//
+		// For methods like eth_getBlockByNumber / eth_getTransactionByHash /
+		// eth_getTransactionReceipt, null means "this upstream does not
+		// have it yet" (tip lag, reorg, pruned). Letting that null win
+		// the hedge cancels the in-flight legs that could have returned
+		// the data, then forces the retry layer to redo the whole fan-
+		// out — amplifying latency on the cold path. Reject emptyish
+		// here so the hedge keeps racing for a non-empty sibling; if all
+		// legs finish empty the failsafe hedge falls through to the
+		// last response, matching the pre-existing terminal behaviour.
+		if r.IsResultEmptyish(ctx) {
+			method, _ := req.Method()
+			accepted := false
+			for _, m := range e.emptyResultAccept {
+				if m == method {
+					accepted = true
+					break
+				}
+			}
+			if !accepted {
+				return false
+			}
+		}
 		kept = true
 		return true
 	}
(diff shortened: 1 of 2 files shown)


aramalipoor (approved):
LGTM 🚀  Let's just make sure the upstream's score is properly punished due to such 'bad empty' results... and of course the 'wrong empty' metrics is also properly handled, hopefully we don't need any code change for these 2.

kasrakhosravi:
> 'wrong empty' metrics is also properly handled, hopefully we don't need any code change for these 2.

good point. `MetricUpstreamWrongEmptyResponseTotal` increments and `RecordUpstreamMisbehavior fires` when one upstream returns null while another has the block, and misbehavior feeds scoring at 5.0× (highest tier alongside latency). 

Only catch: it only fires when the network response is non-empty, so I think this hedge fix is what unblocks the signal for this bug class to feed better data to the scoring mechanism.

PR #900, opened 2026-05-28
feat(network): cluster-based served tip with strict monotonic clamp and less redundant empty result retries

aramalipoor:
## Summary

- Replaces `EvmHighestLatestBlockNumber` / `EvmHighestFinalizedBlockNumber` MAX-based logic with a cluster picker that returns the **MIN block in the largest agreement cluster of upstreams**, clamped strict-monotonic at the network level via a shared `CounterInt64SharedVariable`.
- Eliminates the production -32014 storm where eRPC was advertising tips that only the single "leader" upstream had, then routing the client's follow-up `getBlock(N)` to a slower vendor still at N-3.

## Why

Groundcover trace analysis on Base `eth_getBlockByNumber unfinalized` showed:
- 99% of slow-tail (>2s) requests are for **explicit numeric blocks** (`0x2c6...`) — clients had already gotten that block number from us.
- Per-trace: 4-5 upstreams attempted, **4 returned -32014 for the same block for 3+ seconds** while only flashblocks/prism had it immediately.
- Root cause: eRPC's own `eth_blockNumber` + `latest`-interpolation logic returned `MAX(upstream tips)` — a value only the leader could serve — then routing distributed follow-ups to whichever upstream the policy ranked highest.

## The fix in one line

> The lowest block in the largest-agreement cluster of upstream tips, clamped to never go below what we've already served.

## Algorithm

1. Drop syncing / zero / velocity-rejected observations.
2. Sort ascending; cluster greedily by Δ (auto-derived from chain block time, clamped `[2, 10]`).
3. Pick the dominant cluster: `max by (size desc, then min desc)` — size wins, chain-forward direction breaks ties.
4. Return MIN of the dominant cluster.
5. Strict-monotonic via per-network `TryUpdate` on a `CounterInt64SharedVariable` (same primitive each per-upstream poller uses).

Velocity gate (internal): drops single-upstream fantasy-future tips before clustering. Protects 1- and 2-upstream networks from a buggy leader pulling the served tip to garbage.

## Worked behavior

| Tips | Δ | Clusters | Dominant | Returned | Notes |
|---|---|---|---|---|---|
| `[100]` | 2 | `{100}` | `{100}` | **100** | Single upstream |
| `[100, 99]` | 2 | `{99,100}` | size 2 | **99** | 1-block jitter; both servable |
| `[100, 100, 95]` | 2 | `{95}`, `{100,100}` | `{100,100}` | **100** | 1 lagger excluded |
| `[100, 50, 49]` | 2 | `{49,50}`, `{100}` | `{49,50}` | **49** proposed → monotonic holds previous | Lagging cluster wins by size; monotonic refuses regression |
| `[101, 100, 99, 50, 49]` | 2 | `{49,50}`, `{99,100,101}` | size 3 | **99** | Larger close-cluster wins |
| `[100, 50]` | 2 | `{50}`, `{100}` | tiebreak higher-min | **100** | Lone leader allowed; monotonic still gates regression |

## What semantically changes

| | Before | After |
|---|---|---|
| `eth_blockNumber` response (with integrity enforcement) | Returns MAX(tips) — servable by only the leader | Returns cluster MIN — servable by every dominant-cluster member |
| `latest` / `finalized` tag interpolation | Resolves to MAX(tips) | Resolves to cluster MIN |
| `eth_getBlockByNumber` enforcement | Compares against MAX(tips) | Compares against cluster MIN |
| Strict-monotonic served tip | No | Yes, network-level via `TryUpdate` |
| Block-aware per-attempt routing | `checkUpstreamBlockAvailability` (unchanged) | `checkUpstreamBlockAvailability` (unchanged) |

The existing `checkUpstreamBlockAvailability` per-attempt skip handles the routing-level skip already, so this PR intentionally does NOT add a new routing pre-filter.

## API additions

`UpstreamsRegistry.SharedStateRegistry()` — accessor that exposes the existing shared-state backing store so `Network` can allocate its own served-tip counters without threading the registry through every `NewNetwork` call site.

## Test plan

- [x] 24 pure-function tests for the cluster picker — every scenario in the design matrix
- [x] 6 integration tests for `Network.EvmHighestLatestBlockNumber` new semantics + the monotonic clamp
- [x] Full `./architecture/evm/` test suite passes
- [x] `TestHttpServer_EvmGetBlockByNumber` — 19 enforcement subtests all pass unchanged
- [x] `TestNetwork_HighestLatestBlockNumber` / `TestNetwork_HighestFinalizedBlockNumber` — all existing subtests pass
- [x] `TestInterpolation_UpstreamSkipping_DisabledByMethodConfig` passes

### Known issue (pre-existing flakiness)

`TestNetwork_HedgeAttemptsExcludedFromTrackerCounters/HedgeLoser_ElapsedRecordedAsLowerBoundLatency` is timing-sensitive: it relies on rpc2's state poller not having populated by the time the test issues its first request, so block availability gating doesn't skip rpc1. This race exists independently of this PR (the same `checkUpstreamBlockAvailability` would skip rpc1 whenever rpc2's poller is ready). The test should be reworked to either pin both upstreams to the same latest block, or set `EnforceBlockAvailability: false` for `eth_getBalance` — orthogonal to the served-tip change.

### DIFF
--- a/architecture/evm/common.go
+++ b/architecture/evm/common.go
@@ -54,11 +54,12 @@ func upstreamPostForward_markUnexpectedEmpty(
 }
 
 // emptyResultIsFutureBlock reports whether `rq` targets a concrete block number
-// more than one block beyond the network's served latest — a not-yet-produced
-// block for which every upstream legitimately returns empty. Exactly head+1 is
-// allowed (the next block may land on a retry); head+2 and beyond are "future".
-// Returns false (fail-open) when the head is unknown or the request does not
-// target a concrete numeric block (tags and block-hash lookups are never future).
+// beyond the network's served latest by more than MaxFutureBlockRetryDistance —
+// a not-yet-produced block for which every upstream legitimately returns empty.
+// The distance defaults to 1 (head and head+1 stay retryable); a negative
+// distance disables the guard. Returns false (fail-open) when the head is
+// unknown or the request does not target a concrete numeric block (tags and
+// block-hash lookups are never future).
 func emptyResultIsFutureBlock(ctx context.Context, rq *common.NormalizedRequest) bool {
 	if rq == nil {
 		return false
@@ -71,12 +72,24 @@ func emptyResultIsFutureBlock(ctx context.Context, rq *common.NormalizedRequest)
 	if nw == nil {
 		return false
 	}
+	cfg := nw.Config()
+	if cfg == nil || cfg.Evm == nil {
+		return false
+	}
+	distance := int64(1)
+	if cfg.Evm.MaxFutureBlockRetryDistance != nil {
+		distance = *cfg.Evm.MaxFutureBlockRetryDistance
+		if distance < 0 {
+			// Negative disables the bound: retry all empties (legacy behavior).
+			return false
+		}
+	}
 	head := nw.EvmHighestLatestBlockNumber(ctx)
 	if head <= 0 {
 		// Fail open: without a known head we cannot tell future from behind.
 		return false
 	}
-	return bn > head+1
+	return bn > head+distance
 }
 
 // normalizeEmptyArrayResponse returns a new NormalizedResponse with result `[]`,
--- a/architecture/evm/common.go
+++ b/architecture/evm/common.go
@@ -25,6 +25,15 @@ func upstreamPostForward_markUnexpectedEmpty(
 		}
 	}
 
+	// Future-block guard: do not retry an empty result for a concrete block more
+	// than one beyond the network's served latest — it is not yet produced, so
+	// every upstream legitimately returns empty. Return the truthful empty instead
+	// of churning retries until the request times out. Exactly head+1 is still
+	// retried (the next block may land on a retry).
+	if emptyResultIsFutureBlock(ctx, rq) {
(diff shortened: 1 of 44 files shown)


aramalipoor:
CI note: the `TestNetwork_EvmGetLogs` failure reported above was on attempt 1 only — a pre-existing `gock` concurrency flake where mock matching races under the parallel `eth_getLogs` sub-request fan-out (the full range splits into 3 concurrent sub-requests, one served from cache). It's unrelated to this PR, which doesn't touch getLogs splitting or that test; the test passes locally under `-race`. Re-ran `units` → passed, and all checks are green now.

aramalipoor:
De-flaked `TestGrpcBdsClient_WatchdogReplacesWedgedConn` in dda27482 — it asserted the wedged pool slot was replaced within a 2s `require.Eventually`, which flakes on loaded Blacksmith runners when the 250ms-capped calls + watchdog goroutine get scheduled late (passes locally and on standard runners). Widened the max wait to 10s; it still returns the instant the slot swaps, so the fast path is unchanged. Unrelated to this PR's failsafe changes.

PR #904, opened 2026-06-02
feat: bound retry-on-empty for point-lookups by distance to chain head

kasrakhosravi:
## Problem

Point-lookup methods listed in `markEmptyAsErrorMethods` (e.g. `eth_getBlockByNumber`) convert an empty/null upstream result into a retryable missing-data error, so the network rotates to another upstream that might have the data. That is correct when the block exists but a given upstream is briefly behind.

It is counterproductive when a client requests a block number **ahead of the chain head**: the block has not been produced yet, so *every* upstream legitimately returns empty. erpc then retries until the block lands or the request hits its ceiling timeout — surfacing timeouts on what should be a fast `null`. Tightening the global retry caps (`maxAttempts` / `emptyResultMaxAttempts`) instead reduces the latency but regresses the genuinely‑laggy‑upstream case, trading timeouts for a surge of missing‑data errors.

The missing signal is **distance from the chain head**, which the network already tracks via `EvmHighestLatestBlockNumber`.

## Change

Add an opt-in per-network bound: `evm.maxFutureBlockRetryDistance`.

An empty point-lookup result is treated as retryable missing-data only while the requested block is within the configured distance of the network's highest known latest block. Beyond that, the block is treated as not‑yet‑produced and the empty result is returned as the truthful answer (e.g. a `null` block) without retrying.

- Only concrete numeric block numbers are considered. Tags (`latest` / `pending` / `finalized` / …) and block-hash lookups are never treated as future.
- Fails open when the head is unknown (e.g. a cold state poller), preserving existing behavior.
- `nil` disables the bound (default) — behavior is unchanged unless a network opts in. `0` means only the head block itself is considered present; negative values are rejected by config validation.

The bound is enforced in two places:

1. `upstreamPostForward_markUnexpectedEmpty` — the empty→missing-data trigger (primary).
2. `shouldRetryWithReason` — the network-level retry decision, on the `RetryEmpty` directive and missing-data paths (defense in depth).

## Observability

New counter `network_future_block_empty_kept_total{network, method}` counts empties preserved as truthful because the requested block is beyond the head, for validating rollout.

## Tests

- **Predicate** (`EmptyResultTruthfulForFutureBlock`): below / at / within-slack / beyond-slack, zero slack, negative and nil bound, tags, cold head (fail-open), nil-safety, non-EVM config.
- **Hook** (`HandleUpstreamPostForward`): future block keeps the empty result; at/below head still marks missing-data; disabled bound and cold head preserve current behavior; explicit `retryEmpty=false` unaffected.
- **Retry decision** (`shouldRetryWithReason`): future-block empty/missing-data not retried; at/below head retried as before; disabled bound and cold head retry as baseline.
- **Config validation**: nil and zero valid, negative rejected.

---

> [!NOTE]
> **Medium Risk**
> Changes retry and missing-data behavior for mark-empty point-lookups on all EVM networks by default (distance 1); mis-tuned head tracking could affect edge cases, though cold/unknown head fails open to prior behavior.
> 
> **Overview**
> Adds **`evm.maxFutureBlockRetryDistance`** so empty/null **point-lookup** results (methods in `markEmptyAsErrorMethods`) are only treated as retryable **missing data** while the requested **numeric** block is within that distance of the network’s highest known latest head. Blocks farther ahead are assumed **not yet produced**, so empties are returned as-is (e.g. `null`) instead of rotating upstreams until timeout.
> 
> **`EmptyResultTruthfulForFutureBlock`** implements the check (mark-empty methods only, concrete block numbers—not tags/hashes, fail-open when head is unknown). It is wired into **`upstreamPostForward_markUnexpectedEmpty`** (primary) and **`networkExecutor.shouldRetryWithReason`** for `RetryEmpty` / missing-data paths (defense in depth). Unset config uses default distance **1**; negative values are rejected at validate time.
> 
> New Prometheus counter **`network_future_block_empty_kept_total`**. Config surface updated in Go and TypeScript; tests cover predicate, post-forward hook, retry logic, and validation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5ab85a93897779c35a37805fa79e6a856bedb7d6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/architecture/evm/common.go
+++ b/architecture/evm/common.go
@@ -4,12 +4,14 @@ import (
 	"context"
 
 	"github.com/erpc/erpc/common"
+	"github.com/erpc/erpc/telemetry"
 )
 
 // upstreamPostForward_markUnexpectedEmpty converts empty results for point-lookups
 // (blocks, transactions, receipts, traces, etc.) to missing-data so network retry can rotate.
 func upstreamPostForward_markUnexpectedEmpty(
 	ctx context.Context,
+	n common.Network,
 	u common.Upstream,
 	rq *common.NormalizedRequest,
 	rs *common.NormalizedResponse,
@@ -23,6 +25,20 @@ func upstreamPostForward_markUnexpectedEmpty(
 		if rd := rq.Directives(); rd != nil && !rd.RetryEmpty {
 			return rs, re
 		}
+	}
+
+	// When the request reaches for a block beyond the known chain head by more than the
+	// configured distance, the block has almost certainly not been produced yet: the empty
+	// result is truthful and must not be converted into retryable missing-data (which would
+	// otherwise retry across upstreams until the block lands or the request times out).
+	if EmptyResultTruthfulForFutureBlock(ctx, n, rq) {
+		method, _ := rq.Method()
+		var networkId string
+		if n != nil {
+			networkId = n.Id()
+		}
+		telemetry.MetricNetworkFutureBlockEmptyKeptTotal.WithLabelValues(networkId, method).Inc()
+		return rs, re
 	}
 
 	// Build a simple message and include raw result in details for diagnostics.
--- a/architecture/evm/eth_getBlockByNumber_test.go
+++ b/architecture/evm/eth_getBlockByNumber_test.go
@@ -11,7 +11,8 @@ import (
 
 // testNetwork is a simple test implementation of common.Network interface for this file
 type testNetwork struct {
-	cfg *common.NetworkConfig
+	cfg                *common.NetworkConfig
+	highestLatestBlock int64
 }
 
 func (t *testNetwork) Architecture() common.NetworkArchitecture {
@@ -51,7 +52,7 @@ func (t *testNetwork) Logger() *zerolog.Logger {
 }
 
 func (t *testNetwork) EvmHighestLatestBlockNumber(ctx context.Context) int64 {
-	return 0
+	return t.highestLatestBlock
 }
 
 func (t *testNetwork) EvmHighestFinalizedBlockNumber(ctx context.Context) int64 {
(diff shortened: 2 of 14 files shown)


aramalipoor:
Closing as **superseded by #912** (now on `main`), which implements the same "don't retry-till-timeout on a not-yet-produced block" behavior through a broader mechanism.

**Mapping to what landed on `main`:**

- `evm.maxFutureBlockRetryDistance` (this PR) → **`evm.emptyResultConfidence: blockHead | finalizedBlock`**. `main` now explicitly deprecates and ignores the former — `common/defaults.go`: *"evm.maxFutureBlockRetryDistance is deprecated and ignored; use evm.emptyResultConfidence (blockHead|finalizedBlock) instead."*
- The empty→missing-data predicate is `emptyResultBeyondConfidence` (`architecture/evm/common.go`), wired into the same two enforcement points this PR targeted (`markUnexpectedEmpty` + `shouldRetryWithReason`).
- `main` goes one step further: `tryShortCircuitFutureBlock` (`erpc/networks.go`) returns the truthful `null` **before dispatch** — comparing the requested block against the served-tip / max head — so a future-block read skips the upstream fan-out entirely, not just the retry.

**Net:** `main` is a superset on a cleaner surface — a confidence *level* (and the served tip as the reference) rather than a fixed distance `N`. This PR is also currently `CONFLICTING` because `main` reworked these exact files.

The only piece here not on `main` is the `network_future_block_empty_kept_total{network,method}` counter; since future-block reads now short-circuit pre-dispatch, it's largely moot, but it can be ported standalone if that observability is wanted.

PR #906, opened 2026-06-03
fix(consensus): respect use-upstream directive in participant pool

kasrakhosravi:
The consensus executor builds its participant pool from the full network upstream list, so the `use-upstream` directive is only enforced per-slot in `NextUpstream`. With consensus enabled, a request pinned via `use-upstream` can still fan out and require agreement across upstreams it never targeted.

This filters the participant pool by the directive at the top of the executor, before any slot is spawned. A pattern that matches nothing leaves the pool untouched (per-slot enforcement still applies). Unit test included.

---

> [!NOTE]
> **Medium Risk**
> Changes upstream selection and consensus fan-out for pinned requests; behavior is intentional but affects routing under consensus and stateful-method checks.
> 
> **Overview**
> Requests pinned with **`use-upstream`** (header/query) now narrow the **upstream pool** when `Network.Forward` builds `upsList`, not only when `NextUpstream` picks a slot.
> 
> A new **`filterUpstreamsByUseUpstream`** helper applies wildcard matching on upstream ids, preserves order, and **does not empty** the pool when nothing matches (same as before for “no match” cases). With **consensus** enabled, participant fan-out, `maxParticipants`, sequential sweeps, and the **stateful-method** upstream count all see the same reduced set the client intended.
> 
> Unit tests cover exact id, globs, no-op patterns, and the no-match fallback.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9f85b7d61162ffcc4b84e7ff5d3ae62ca3e1d90d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/erpc/networks.go
+++ b/erpc/networks.go
@@ -486,6 +486,14 @@ func (n *Network) Forward(ctx context.Context, req *common.NormalizedRequest) (*
 			upsList = append(upsList, u)
 		}
 	}
+	// Honor the use-upstream directive at the pool level so every downstream
+	// consumer operates on the targeted upstream(s): the consensus participant
+	// fan-out (and its maxParticipants count), the sequential sweep, and the
+	// stateful-method guard. NextUpstream also enforces this per-slot; filtering
+	// here keeps the request's upstream set consistent with what is dispatched.
+	if dr := req.Directives(); dr != nil && dr.UseUpstream != "" {
+		upsList = filterUpstreamsByUseUpstream(upsList, dr.UseUpstream)
+	}
 	upstreamSpan.SetAttributes(attribute.Int("upstreams.count", len(upsList)))
 	if common.IsTracingDetailed {
 		ids := make([]string, len(upsList))
--- a/erpc/use_upstream.go
+++ b/erpc/use_upstream.go
@@ -1,0 +1,34 @@
+package erpc
+
+import "github.com/erpc/erpc/common"
+
+// filterUpstreamsByUseUpstream restricts a request's candidate upstream pool to
+// those whose id matches the `use-upstream` directive.
+//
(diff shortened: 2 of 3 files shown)


consensus/executor.go
```diff
@@ -127,6 +127,19 @@ func (e *executor) Run(
...
+	// pool (and the maxParticipants count) is the full network set and consensus
+	// races/requires agreement across upstreams the caller never targeted — the
+	// directive is only enforced per-slot in NextUpstream. Runs before the quota
+	// reorder and before any slot consumes an upstream (req.UpstreamIdx == 0).
+	if dirs := originalReq.Directives(); dirs != nil && dirs.UseUpstream != "" {
+		all := originalReq.Upstreams()
+		if filtered := filterParticipantsByUseUpstream(all, dirs.UseUpstream); len(filtered) < len(all) {
+			originalReq.SetUpstreams(filtered)
+		}
+	}
```
aramalipoor: This seems like a generic logic, why should it be only in consensus executor here? and if we should do it only for consensus then we shouldn't probably change the Request object (i.e. do this in the consensus execution flow vs changing original request)

kasrakhosravi: Good points — agreed on both. Moved it out of the consensus executor: the filter now runs once in `Network.Forward`, right where the request's upstream pool is established (just before `SetUpstreams`), so it's generic — consensus fan-out, the sequential sweep, and the stateful-method guard all honor the directive. No new mutation of the request from inside the consensus flow; it's the existing pool assignment. Pushed as a small `filterUpstreamsByUseUpstream` helper + unit test; `executor.go` is reverted to no change.

PR #912, opened 2026-06-06
feat(routing): tag-aware use-upstream + selector-scoped served tip

aramalipoor:
## What & why

For chains where one chain id is served by **two isolated node groups** (e.g. Base `flashblocks` vs `normal`, HyperEVM `systx` vs `standard`), a more-ahead group currently defines the network's `latest` for *every* request — so a request pinned to the other group resolves `latest` to a block that group doesn't have yet (→ "block not found" churn). This PR lets `use-upstream` target a group **by tag** and decides `latest`/`finalized` **among that group**.

## Changes

- **Tag-aware `use-upstream`** — `common.MatchesSelector` / `UpstreamMatchesSelector` match the upstream **id OR any tag** for positive patterns; negation (`!…`) stays id-only so an exclusion can't be rescued by a non-matching tag. Wired into the three existing selection sites (`NextUpstream`, `acceptsRequest`, `shouldHandleMethod`). Existing id-only configs are a strict subset (unchanged).
- **Selector-scoped served tip** — the request is bound to `ctx` at the top of `Forward`; `tipCandidateUpstreams` filters the tip-input set by the request's selector, so `EvmHighestLatest/FinalizedBlockNumber` (both max-mode and cluster-mode), the future-block short-circuit, and `guaranteedMethodFloor` all resolve among the selected group.
- **Per-group monotonic shared state** — a selector that **exactly matches a configured tag** materializes a lazy, cross-pod strict-monotonic per-group counter (`servedLatestBlock/<net>/sel:<tag>`), so each group has its own monotonic tip. Selectors that don't match a real tag (id globs, unknown/garbage tags) create **no** shared state and fall back to a stateless cluster-min over the subset.

## Design notes for reviewers

- **Scope is deliberately narrow:** the background **selection policy and its block-head-lag exclusion are unchanged**. Isolation applies only to the `latest`/`finalized` decision (where the cross-group problem actually bites). Lagging upstreams are still excluded exactly as today.
- **DDoS / memory safety (3k upstreams × 400 networks, 1 pod):** durable per-group state is created **only** for selectors that exactly match a tag configured on the network's upstreams (a bounded, operator-controlled universe), and is capped per network (`maxServedTipPartitions = 16`). Attacker-supplied selectors can never grow the shared-state map. There's a test asserting exactly this (`OnlyConfiguredTagsMaterializeState`). The no-selector hot path is untouched (no new allocation).
- **`family` is not special** — any tag (or id) works as a selector; `family:` is just a convention an operator may pick.

## Tests
- `TestMatchesSelector` (id/tag/glob/bare-tag/negation-regression-guard)
- `TestNetwork_Forward_UseUpstreamTagSelector` (routes within group, failover containment, unmatched fast-fail)
- `TestServedTip_SelectorScoped` (max-mode, cluster-mode, DDoS-safety of materialization)
- Regression: served-tip / forward / `common` / `upstream` suites pass.

## Heads-up (pre-existing, not from this PR)
`TestNetwork_HedgeAttemptsExcludedFromTrackerCounters/HedgeLoser_ElapsedRecordedAsLowerBoundLatency` fails on this machine — verified it fails **identically on clean `main` (dd95fe0c)** with none of these changes (recorded latency `0`). Flagging in case it's a real regression on the default branch.

Design spec: `specs/selector-isolation/{feature.md,plan.md}`.


aramalipoor:
Re-ran `units`. The failure (`TestGrpcBdsClient_WatchdogReplacesWedgedConn`) is a flaky timing-sensitive gRPC watchdog test unrelated to this PR — no `clients/` files are touched here, and it passes locally 3/3. Rerun in progress.

PR #913, opened 2026-06-08
feat(selection-policy): add includeIf primitive and pool-level quantifiers

kasrakhosravi:
## What

Adds an `includeIf` primitive to the selection-policy stdlib — the dual of `excludeIf`. Where `excludeIf` **drops** a per-upstream offender, `includeIf` **admits** a selected set of upstreams from the full universe when an aggregate condition over the surviving pool holds. It never removes an upstream.

```js
upstreams
  .excludeTag('tier:reserve')                                            // reserve out of normal rotation
  .includeIf('tier:reserve', (p) => p.every(blockSecondsLagAbove(30)))   // admit if every survivor lags
  .includeIf('tier:reserve', (p) => p.length < 2)                        // ...or if too few remain
  .sortByScore(PREFER_FASTEST)
```

`includeIf(target, condition)` — the **target comes first** so the policy reads "include *these* if *condition*". The target is a tag pattern for the common case, or a selector object (`{ id, tag, vendor, type, position }`) whose facets AND together. The condition is a boolean or a function over the **current** chain array (the pool that survived earlier steps), written with native `every` / `some` / `length` / `filter` over the existing per-upstream predicate factories — no parallel quantifier vocabulary.

## Why

This enables a **break-glass reserve tier**: upstreams kept out of normal rotation and ranked in alongside the survivors only when the serving pool is collectively unfit — too few left, all lagging, all slow. Because it **adds** rather than excludes, a single slow primary is never evicted; the reserve is offered alongside the degraded pool and a later `sortByScore` decides the order.

The explicit `excludeTag` + `includeIf` pair is deliberate: `excludeTag` states "this tier is NOT in normal rotation", `includeIf` states "this is the only condition under which it comes back" — the policy documents its own intent (now also spelled out in the docs).

## Behavior details

- At least one selector facet must resolve, so a facet-less target (`{}`, `{ position: 'head' }`) is a no-op — it never admits the whole universe by accident.
- Already-present upstreams are not duplicated; admitted upstreams default to the tail so survivors keep priority.
- Malformed args and throwing conditions degrade to a no-op rather than failing the eval (which would drop the network back to the default policy).
- Native `every` is true on `[]`, so a bare `p.every(...)` admits on an empty pool; guard with `p.length > 0 &&` when that's not wanted.
- Pure JS over data already exposed to the eval — no engine or Go changes. Additive only.

## Tests

- Go unit tests (`internal/policy/stdlib/include_if_test.go`): admit / no-admit gating, tag and object targets, all selector facets, `position`, dedup, the surviving-pool view of the condition, `every`/`some`/`length`/`filter` conditions, empty-pool semantics, throwing-condition safety, and the end-to-end reserve-tier pattern.
- TS type-level tests (`erpc.test-d.ts`): the chained surface plus negative `@ts-expect-error` cases.
- Docs (`selection-policies.mdx`): reference entry, the "Reserve tier" worked pattern, and the rationale for keeping `excludeTag` explicit.

### DIFF
--- a/docs/pages/config/projects/selection-policies.mdx
+++ b/docs/pages/config/projects/selection-policies.mdx
@@ -448,12 +448,11 @@ Use `evalScope: 'network-method-finality'` to get one slot per `(method, finalit
     .stickyPrimary({ hysteresis: 0.30, minSwitchInterval: '30s' })
 ```
 
-This differs from a `preferTag` fallback: `preferTag` only reaches the
-fallback tier when the primary tier is **empty**, whereas `includeIf`
-brings the reserve in while primaries are still serving but
-**collectively degraded** (lagging / slow), letting `sortByScore` pick
-between them. Pair with `routing.probe: 'off'` on the reserve upstreams
-if you also want to keep shadow-probe traffic off them while they're out.
+The reserve is brought in while primaries are still serving but
+**collectively degraded** (lagging / slow) — not only when they're
+entirely gone — letting `sortByScore` pick between them. Pair with
+`routing.probe: 'off'` on the reserve upstreams if you also want to keep
+shadow-probe traffic off them while they're out.
 
 ### Per-method override
 
--- a/docs/pages/config/projects/selection-policies.mdx
+++ b/docs/pages/config/projects/selection-policies.mdx
@@ -441,9 +441,9 @@ Use `evalScope: 'network-method-finality'` to get one slot per `(method, finalit
     .removeCordoned()
     .excludeIf(all(samplesAbove(10), errorRateAbove(0.7)))
     .excludeTag('tier:reserve')
-    .includeIf((p) => p.length > 0 && p.every(blockSecondsLagAbove(30)), { tag: 'tier:reserve' })
-    .includeIf((p) => p.length > 0 && p.every(latencyAbove(2000, 99)),   { tag: 'tier:reserve' })
-    .includeIf((p) => p.length < 1,                                      { tag: 'tier:reserve' })
+    .includeIf('tier:reserve', (p) => p.length > 0 && p.every(blockSecondsLagAbove(30)))
+    .includeIf('tier:reserve', (p) => p.length > 0 && p.every(latencyAbove(2000, 99)))
+    .includeIf('tier:reserve', (p) => p.length < 1)
     .sortByScore(PREFER_FASTEST)
     .stickyPrimary({ hysteresis: 0.30, minSwitchInterval: '30s' })
 ```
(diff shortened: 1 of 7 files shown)


docs/pages/config/projects/selection-policies.mdx
```diff
@@ -426,6 +426,33 @@ Use `evalScope: 'network-method-finality'` to get one slot per `(method, finalit
...
+// A reserve pool (e.g. a pricier or rate-limited provider, tagged
+// `tags: [tier:reserve]`) is kept OUT of normal rotation, then ranked in
+// only when the serving pool is collectively unfit — every survivor
+// lagging, or too few left. `includeIf` adds; it never evicts a survivor,
+// so a single slow primary alone does not pull the reserve in.
+(upstreams, ctx) =>
+  upstreams
+    .removeCordoned()
+    .excludeIf(all(samplesAbove(10), errorRateAbove(0.7)))
+    .preferTag('!tier:reserve', { minHealthy: 1, fallback: 'tier:reserve' })
```
aramalipoor: Do we really need "preferTag" as a primitive anymore?

docs/pages/config/projects/selection-policies.mdx
```diff
@@ -1130,6 +1206,41 @@ Composed predicates carry a joined `policyReason` —
...
+into a single boolean — an aggregate, "network-level" question. They
+produce the conditions [`includeIf`](#probing--forced-inclusion)
+consumes.
+
+<details>
+<summary>**Pool-level conditions**</summary>
+
+```ts
+forAll(pred)         // every upstream satisfies pred (false on an empty pool)
+forAny(pred)         // at least one satisfies pred
```
aramalipoor: Don't we already have "any" and "all" helpers?

kasrakhosravi: good call — dropped the quantifiers entirely. conditions are just native every/some/length/filter over the existing predicate factories now, e.g. `(p) => p.every(blockSecondsLagAbove(30))`. no parallel vocab shadowing any/all.

docs/pages/config/projects/selection-policies.mdx
```diff
@@ -897,6 +925,62 @@ Always include matching upstreams, even if prior filters dropped them.
...
+The function form of `condition` receives the **current** chain array
+— the pool that survived earlier steps — so it can ask aggregate
+("network-level") questions about what's left. The array is a normal
+JS array, so use native `every` / `some` / `length` / `filter` over the
+[predicate factories](#predicate-factories):
+
+```js
+upstreams
+  .excludeTag('tier:reserve')                                             // reserve out of normal rotation
+  .includeIf((p) => p.every(blockSecondsLagAbove(30)), { tag: 'tier:reserve' })  // admit if every survivor lags
```
aramalipoor: Should we make the target tag a first-class input?
```suggestion
  .includeIf('tier:reserve', (p) => p.every(blockSecondsLagAbove(30)))  // admit if every survivor lags
```
Because if it is not provided it wouldn't be intuitive what it means?

Somehow it's better to explicitly show that we'll NOT include these initially, and 'only include if' the condition applies (docs/comments? name of the helper?)

PR #914, opened 2026-06-08
feat(cache): serve eth_blockNumber from the gRPC cache connector

kasrakhosravi:
The backing reader exposes no eth_blockNumber RPC, so the gRPC cache connector skipped the method and it always fell through to upstreams. Translate it on read: fetch getBlockByNumber("latest") and return just the block number in the eth_blockNumber result shape (a quoted hex string). Reuses the existing latest-block fetch path; the result is canonical hex via a uint64 round-trip.

Freshness is enforced by the existing realtime cache gate, which falls back to the connector's reported head timestamp for response-less methods like this one.

---

> [!NOTE]
> **Low Risk**
> Small, read-only cache path change with explicit miss behavior and tests; no auth or persistence changes.
> 
> **Overview**
> The gRPC read-through cache connector can now satisfy **`eth_blockNumber`** instead of skipping it and always falling through to upstreams.
> 
> **`GrpcConnector.Get`** allowlists the method and, when BDS has no native RPC, derives the answer by calling **`fetchTaggedBlock`** with **`latest`**, then returns a JSON-RPC-shaped quoted hex string (`"0x…"`) with canonical formatting via a uint64 round-trip. If the latest block cannot be read or has no parsable **`number`**, it returns a **record-not-found** miss so callers can retry upstream. Freshness for this response-less method still relies on the existing realtime cache gate and **`CacheLatestBlockTimestamp`**.
> 
> New unit tests in **`grpc_blocknumber_test.go`** cover happy path, zero block, leading-zero normalization, RPC errors, and missing **`number`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 21a0c5883ed2520b3f1b9f5291efd7e3a75c22ef. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/data/grpc.go
+++ b/data/grpc.go
@@ -63,6 +63,9 @@ var supportedMethods = map[string]struct{}{
 	"eth_getTransactionReceipt": {},
 	"eth_getBlockReceipts":      {},
 	"eth_chainId":               {},
+	// eth_blockNumber has no native BDS gRPC method; Get derives it from the
+	// latest block (see the translation in Get).
+	"eth_blockNumber": {},
 }
 
 func NewGrpcConnector(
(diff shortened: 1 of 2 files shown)


data/grpc.go
```diff
@@ -63,6 +63,9 @@ var supportedMethods = map[string]struct{}{
 	"eth_getTransactionReceipt": {},
 	"eth_getBlockReceipts":      {},
 	"eth_chainId":               {},
+	// eth_blockNumber has no native BDS gRPC method; Get derives it from the
```
hawkaa, line 66: Should we add it perhaps?

PR #926, opened 2026-06-12
feat(cache): derive realtime cache TTL from estimated block time

kasrakhosravi:
Adds an optional `ttlBlockTimeMultiplier` on cache policies.

For **realtime** finality, the cached-result age limit becomes `estimatedBlockTime * ttlBlockTimeMultiplier` (using the network's EMA block time), instead of the static `ttl`. This lets head freshness scale with each chain's block cadence rather than a single fixed value. Falls back to `ttl` when the block time isn't known yet.

Opt-in: when `ttlBlockTimeMultiplier` is unset, behavior is unchanged.

- `common/config.go`: new `CachePolicyConfig.TTLBlockTimeMultiplier`
- `data/cache_policy.go`: accessor
- `architecture/evm/json_rpc_cache.go`: resolve effective realtime TTL from
  block time (via the network's metrics tracker) with static-TTL fallback
- regenerated `typescript/config` types

`go build ./...`, `go vet`, and cache/policy tests pass.

---

> [!NOTE]
> **Medium Risk**
> Changes realtime cache freshness and connector storage TTL for policies using block-time multipliers; misconfiguration could serve staler head data or evict entries early if read/write paths diverged (this PR explicitly aligns them).
> 
> **Overview**
> Introduces **`BlockTimeAdaptiveDuration`** for cache policy `ttl`: a plain duration stays fixed, or an object `{ blockTimeMultiplier, fallback }` scales realtime freshness to **EMA block time × multiplier** (with fallback / 2s cold-start default when block time is unknown).
> 
> For **realtime** finality, the read-side age guard and **connector `Set` expiry** both use `ResolveTTL` via `Network.EvmBlockTime()`, so storage TTL is not shorter than the window the guard would still serve. **Metrics** keep labeling the fixed/fallback TTL to avoid cardinality from per-sample resolved values. Config validation rejects `blockTimeMultiplier` on non-realtime policies.
> 
> Existing scalar TTL configs map through `FixedDuration`; behavior is unchanged unless the object form is used.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a06786be4e3440e64d5519b62b790b47c9d91b5e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -326,6 +326,12 @@ type CachePolicyConfig struct {
 	MinItemSize *string              `yaml:"minItemSize,omitempty" json:"minItemSize" tstype:"ByteSize"`
 	MaxItemSize *string              `yaml:"maxItemSize,omitempty" json:"maxItemSize" tstype:"ByteSize"`
 	TTL         Duration             `yaml:"ttl,omitempty" json:"ttl" tstype:"Duration"`
+
+	// TTLBlockTimeMultiplier, when > 0, derives the realtime-finality age limit
+	// from the network's estimated block time (blockTime * multiplier) instead of
+	// the static TTL, so head freshness tracks each chain's cadence. Falls back to
+	// TTL when the block time isn't known yet.
+	TTLBlockTimeMultiplier *float64 `yaml:"ttlBlockTimeMultiplier,omitempty" json:"ttlBlockTimeMultiplier,omitempty" tstype:"number"`
 }
 
 type ConnectorDriverType string
--- a/common/config.go
+++ b/common/config.go
@@ -325,17 +325,13 @@ type CachePolicyConfig struct {
 	AppliesTo   CachePolicyAppliesTo `yaml:"appliesTo,omitempty" json:"appliesTo" tstype:"'get' | 'set' | 'both'"`
 	MinItemSize *string              `yaml:"minItemSize,omitempty" json:"minItemSize" tstype:"ByteSize"`
 	MaxItemSize *string              `yaml:"maxItemSize,omitempty" json:"maxItemSize" tstype:"ByteSize"`
-	TTL         Duration             `yaml:"ttl,omitempty" json:"ttl" tstype:"Duration"`
 
-	// TTLBlockTimeMultiplier, when > 0, derives the realtime-finality age limit
-	// from the network's estimated block time (blockTime * multiplier), so head
-	// freshness tracks each chain's cadence.
-	//
-	// It composes with TTL rather than conflicting: block time is used when
-	// known; otherwise (cold start, or a network with no estimate) it falls back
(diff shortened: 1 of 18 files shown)


common/config.go
```diff
@@ -326,6 +326,12 @@ type CachePolicyConfig struct {
 	MinItemSize *string              `yaml:"minItemSize,omitempty" json:"minItemSize" tstype:"ByteSize"`
 	MaxItemSize *string              `yaml:"maxItemSize,omitempty" json:"maxItemSize" tstype:"ByteSize"`
 	TTL         Duration             `yaml:"ttl,omitempty" json:"ttl" tstype:"Duration"`
+
+	// TTLBlockTimeMultiplier, when > 0, derives the realtime-finality age limit
+	// from the network's estimated block time (blockTime * multiplier) instead of
+	// the static TTL, so head freshness tracks each chain's cadence. Falls back to
+	// TTL when the block time isn't known yet.
+	TTLBlockTimeMultiplier *float64 `yaml:"ttlBlockTimeMultiplier,omitempty" json:"ttlBlockTimeMultiplier,omitempty" tstype:"number"`
```
aramalipoor: probably it should fail if used together? or maybe ttl can be used by default if it can't derive the block time

aramalipoor: A fund idea would be to create a new DynamicDuration type (we might already have that?) and use it here and other places.

This type can be either a fixed value, or an object defining a dynamic block-time-based value?

kasrakhosravi: Done — went with this. `ttl` is now a single polymorphic `DynamicDuration` (scalar **or** object), replacing the separate `ttlBlockTimeMultiplier` field:

```yaml
ttl: 2s                                        # fixed
ttl: { blockTimeMultiplier: 1, fallback: 2s }  # blockTime * 1, 2s cold-start floor
```

Modeled it after `AdaptiveDuration` (scalar/object wire form) as a sibling type rather than overloading it, since the resolution input differs (block time vs latency quantile). `Resolve(blockTime, coldStartDefault)` handles the fallback chain; `GetTTL()` still returns the fixed component for storage expiry. Generated TS types updated to `Duration | DynamicDuration`, and added unit tests for the scalar/object unmarshal (YAML + JSON) and Resolve. If you'd prefer it folded into `AdaptiveDuration` instead, happy to switch.

aramalipoor: Same PR, also avoid creating a new object if we can use the existing object. Also if we're using AdaptiveDuration then make sure it fails when used in the wrong context.

Probably better approach is QuantileAdaptiveDuration BlockTimeAdaptiveDuration?

aramalipoor (approved):

PR #928, opened 2026-06-12
fix: consensus wait caps armed by skipped participants resolve rounds before any real attempt

kasrakhosravi:
## Problem

Two compounding issues affect consensus fan-out when some upstreams statically skip the matched method — a common setup being archive nodes configured with `ignoreMethods: ["eth_send*"]` alongside full nodes that handle broadcasts:

1. **`ignoreMethods` is enforced too late.** Upstream selection installs the full policy-ordered list on the request; method-ineligible upstreams are only rejected at forward time (`ErrUpstreamMethodIgnored` via `shouldSkip`). Until then they consume consensus participant slots and retry/hedge attempts just to produce instant skips.

2. **Wait caps are armed by responses that never reached an upstream.** Since 0.1.0, `maxWaitOnResult`/`maxWaitOnEmpty` default to adaptive values whenever consensus is configured. `considerWaitCap` arms `maxWaitOnEmpty` on the *first response of any kind* — including skip results that resolve locally in microseconds and that the analyzer itself classifies as non-votable infrastructure errors (`ResponseTypeInfrastructureError`).

The failing sequence:

- consensus spawns `maxParticipants` slots; each slot is `retry(hedge(tryOneUpstream))` over a shared cursor
- slots that draw skip-configured upstreams (or find the cursor exhausted) burn their entire network retry budget on skips in microseconds (`gave up retrying on network-level after 84µs`) and return instantly
- the first such result arms `maxWaitOnEmpty`; the cap fires before any real participant has answered
- the round resolves with only infrastructure errors → the client receives a "method ignored" flavored error (`-32601`)

With `fireAndForget: true` the detached broadcasts still complete in the background, so a transaction can land on chain while the submitter is told the request failed — the worst possible outcome for `eth_sendRawTransaction` (clients rebroadcast, bump nonces, etc.).

## Fix

Two complementary changes:

1. **Selection-time eligibility** (`erpc/networks.go`): filter the candidate list through the (memoized) `ShouldHandleMethod` before `SetUpstreams`, so method-ineligible upstreams never become attempts or consensus participants. When *every* upstream ignores the method, the unfiltered list is kept so the request still fails with the descriptive per-upstream "method ignored" error instead of a generic "no upstreams found" (existing behavior, still covered by `TestNetwork_Forward_InfiniteLoopWithAllUpstreamsSkipping`).

2. **Skip-aware wait-cap arming** (`consensus/executor.go`): `considerWaitCap` now ignores results that represent no actual upstream attempt — config-static skips (method ignored, shadow upstream, syncing, `use-upstream` mismatch) and empty-cursor outcomes (`ErrNoUpstreamsLeftToSelect`, exhausted/retry-exceeded wrappers whose leaf causes are all skips). This also covers what selection filtering alone cannot: when `maxParticipants` exceeds the number of eligible upstreams, the surplus slots still return instant exhausted-cursor errors that would otherwise arm the cap. Genuine attempt failures (timeouts, 5xx, connection resets) still arm the caps, and all-skip rounds still terminate immediately because every response counts toward the spawn total.

Also adds the previously missing `ErrCodeUpstreamNotAllowed` constant.

## Tests

- `TestWaitCap_NotArmedBySkippedParticipants` — reproduces the failure (30ms caps, one skip + one exhausted-cursor result + one 150ms real participant). Fails on the previous logic — the round resolved at ~30ms with only infrastructure errors — and passes now by waiting for the real participant.
- `TestWaitCap_AllParticipantsSkipped_ResolvesImmediately` — all-skip rounds terminate promptly without relying on any cap.
- `TestWaitCap_ArmedByRealAttemptFailures` — pins that genuine failures still arm `maxWaitOnEmpty`.
- `TestIsNoAttemptError` — classification table including mixed skip+real exhausted causes and wrapper recursion.
- `TestFilterMethodEligible` — order preservation, zero-copy pass-through when nothing is dropped, all-ineligible fallback, input immutability (selection-cache slices are read-only).
- Full `consensus` suite passes with `-race`; `erpc` forward/skip/consensus suites pass.

---

> [!NOTE]
> **Medium Risk**
> Changes consensus timing and upstream selection for fan-out requests; misclassification of skip vs real errors could affect when rounds resolve, but behavior is heavily tested including end-to-end consensus paths.
> 
> **Overview**
> Fixes consensus rounds that could end early with only “method ignored” / infrastructure errors when **instant skip** participant results armed `maxWaitOnEmpty` / `maxWaitOnResult` before any real upstream attempt completed.
> 
> **Upstream selection** (`erpc/networks.go`): before binding the request’s upstream list, drops upstreams that `ShouldHandleMethod` would reject (`ignoreMethods` / `allowMethods`), so they no longer consume consensus/retry slots or emit phantom skips. If every upstream is ineligible, the unfiltered list is kept so clients still get per-upstream “method ignored” errors.
> 
> **Wait-cap arming** (`consensus/executor.go`): `considerWaitCap` ignores **no-attempt** results (static skips, shadow/syncing/`use-upstream` mismatches, empty-cursor / all-skip exhausted wrappers) via `isNoAttemptResult` / `isNoAttemptError`; real transport failures still arm caps. Adds `ErrCodeUpstreamNotAllowed` in `common/errors.go`.
> 
> Tests cover wait-cap behavior, error classification, `filterMethodEligible`, and an end-to-end network consensus path for ignored methods.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 439f5392a816563940f17acedb0f0e64d7f60e07. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/consensus/executor.go
+++ b/consensus/executor.go
@@ -450,8 +450,22 @@ func (e *executor) runAnalyzer(
 		if maxWaitOnEmpty <= 0 && maxWaitOnResult <= 0 {
 			return
 		}
+		// Results that never reached an upstream (config-static skips like
+		// ignored methods or shadow upstreams, and empty-cursor outcomes —
+		// see isNoAttemptResult) are produced locally in microseconds and
+		// carry no signal about how long the round's real participants
+		// need. Arming the caps off them would start the countdown before
+		// any live attempt exists: under fan-out configs where some
+		// upstreams statically skip the method, the round would resolve
+		// with only non-votable infrastructure errors while the real
+		// participants are still in flight. The collection loop still
+		// counts these results toward maxToSpawn, so rounds where every
+		// participant skips terminate immediately without any cap.
+		if isNoAttemptResult(resp) {
+			return
+		}
 		now := time.Now()
-		// First response of any kind arms maxWaitOnEmpty.
+		// First response from an actual upstream attempt arms maxWaitOnEmpty.
 		if maxWaitOnEmpty > 0 && waitDeadline.IsZero() {
 			armTimer(now.Add(maxWaitOnEmpty))
 		}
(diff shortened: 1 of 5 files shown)


consensus/executor.go
```diff
@@ -450,8 +450,22 @@ func (e *executor) runAnalyzer(
...
+		// need. Arming the caps off them would start the countdown before
+		// any live attempt exists: under fan-out configs where some
+		// upstreams statically skip the method, the round would resolve
+		// with only non-votable infrastructure errors while the real
+		// participants are still in flight. The collection loop still
+		// counts these results toward maxToSpawn, so rounds where every
+		// participant skips terminate immediately without any cap.
+		if isNoAttemptResult(resp) {
+			return
+		}
```
aramalipoor, line 466: Should we have some network-level full (real-ish) test to see if the ultimate end goal is achieved? Like intentional full/archive node mix where full nodes ignore certain methods and we expect they to be fully excluded and not cause maxWait or other side effects like counters/metrics impact etc.

aramalipoor (approved):

PR #933, opened 2026-06-12
fix(tests): pin hedge-loser quantile exclusion and close the CI shard gap that hid the stale test

aramalipoor:
## Verdict: stale test, not a scoring regression — plus the CI hole that hid it

`TestNetwork_HedgeAttemptsExcludedFromTrackerCounters/HedgeLoser_ElapsedRecordedAsLowerBoundLatency` fails deterministically (`p70 == 0`), in isolation and in the full suite, on every commit since it merged — verified at #888's own merge commit (70a1f178).

### Why it fails

PR #888 evolved during development and shipped two contradictory artifacts in one commit:

- **Production code** (`upstream/upstream.go`, `ObserveDuration` call sites) deliberately **excludes** canceled attempts from `ResponseQuantiles`, with an explicit rationale: elapsed-by-cancel is "however long the race winner took", not the loser's latency; recording it would bias cross-upstream comparisons (`latencyDeviationAbove`) even when per-method latencies are identical. The "hedge-loser stays optimistic forever" problem is instead handled by `probeExcluded` shadow traffic, which feeds real successful samples.
- **The test** pins the earlier in-development iteration (record elapsed-by-cancel as a "lower bound" sample) that this design replaced before merge.

So the scoring/rotation behavior is working as designed; the test is the leftover. Rewritten as `HedgeLoser_ElapsedExcludedFromQuantiles`, pinning the shipped design: clean `ErrorsTotal` and an **empty** quantile no matter how many hedge races the upstream loses, with a comment trail pointing at the `upstream.go` rationale and the `probeExcluded` mechanism.

### How a deterministically-failing test merged: the CI shard gap

CI runs `make test-fast`, which shards the `erpc` package by regex. The catch-all shard's hand-maintained `SKIP_ALL` list was **broader than what the named shards actually run** — e.g. it skipped all of `TestNetwork_*` while `SHARD_NETWORK` only runs an enumerated subset. Result: six tests ran in **no shard at all**, despite the comment claiming "the catch-all automatically picks up any new/unassigned tests":

```
TestConsensusPolicy_RequiredParticipants
TestNetwork_Forward_UseUpstreamTagSelector
TestNetwork_HedgeAttemptsExcludedFromTrackerCounters   ← the broken one
TestNetwork_LatePrimaryResponseAfterHedgeWin_NoDoubleCounting
TestNetwork_LongTermHedgingDynamics_PromotesFasterUpstream
TestNetwork_TimeoutPolicy
```

Fix: compose `SKIP_ALL` from the shard regexes themselves (plus the two standalone phases, `^TestConsensusPolicy_DSL` and `^TestHttp_`), so the shards and the skip can never drift apart again.

### Verification

- Partition simulation over all **347** `erpc`-package test functions: every test now runs in exactly one place (zero orphans, zero double-runs).
- All six formerly-gapped tests pass at current `main` with the rewritten hedge test (`go test -run '^(…six names…)$'` → ok, 32.5s).
- `make -n test-fast` parses and expands the composed skip correctly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -76,17 +76,25 @@ test-fast:
 	@# of TestNetwork_* was skipped while SHARD_NETWORK only ran an enumerated
 	@# subset — so any test under a skipped prefix but outside a shard ran
 	@# NOWHERE in CI (which is how a deterministically-failing test merged).
+	@# SHARD_CONTAINER_CACHE isolates the testcontainers-backed cache tests:
+	@# their duration is dominated by docker (image pull + dynamodb-local JVM
+	@# startup) and swings from ~15s to 9m+ on degraded runners. In the serial
+	@# catch-all that variance ate the entire 10m process budget and starved
+	@# every test scheduled after them; in a dedicated shard with extra
+	@# headroom it can't take anyone else down.
 	@bash -c '\
 	  SHARD_CONSENSUS="^(TestConsensusPolicy$$|TestConsensusGoroutine|TestConsensusSelectsLargest|TestInterpolation_)"; \
 	  SHARD_NETWORK="^(TestNetwork_Forward$$|TestNetwork_(SelectionScenarios|InFlightRequests|SkippingUpstreams|EvmGetLogs|ThunderingHerd|HighestFinalizedBlockNumber|HighestLatestBlockNumber|CacheEmptyBehavior|BatchRequests|SingleRequestErrors|CapacityExceededErrors|TraceExecutionTimeout|Multiplexer|SendRawTransaction))"; \
 	  SHARD_HTTPSERVER="^TestHttpServer_"; \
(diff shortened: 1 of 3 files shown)


aramalipoor:
**CI investigation for the two `units` timeouts on this branch** (both pre-date the new commit):

Both runs timed out in the catch-all shard **before any of the six newly-unhidden tests executed** (they sit late in file order; run 2's process died while still in the cache-connector tests). The actual time sink was the testcontainers-backed tests running on degraded runners:

| Test | healthy run (main) | run 1 | run 2 |
|---|---|---|---|
| `TestEvmJsonRpcCache_DynamoDB` | 14.7s | ≥9m46s (killed mid-test) | 9m06s (passed) |
| `TestEvmJsonRpcCache_Redis` | 2.6s | — | ≥40s (killed mid-test) |

Docker image pull + dynamodb-local JVM startup dominate these tests and swing ~40x between runners. In the serial catch-all that variance consumed the whole 10m process budget and starved everything scheduled after them — a latent flake that exists on `main` too; this branch just kept drawing slow runners.

Pushed a086624a: the two container-backed tests get their own shard with 15m headroom (units job allows 30m). The composed `SKIP_ALL` makes this a two-line change, and the partition check still shows all 347 erpc-package tests running exactly once.

PR #949, opened 2026-06-18
fix(evm): don't count cancelled cache fan-out reads as success_miss

hawkaa:
## What this does

`EvmJsonRpcCache.Get` fans out cache reads across connectors and drains results until a hit lands, every connector reports a genuine miss, or the fan-out is aborted. The drain loop exits via `fanCtx.Done()` when the parent context is cancelled (e.g. a request-level failsafe/hedge timeout) or the 30s defensive backstop fires. In that case `jrr` stays nil and the post-fan-out block **unconditionally** recorded `cache_get_success_miss_total` plus duration, using `time.Since(start)` as the latency.

That counts a *cancelled* read as a slow cache miss — inflating the miss count and attributing the cancellation-timeout latency to the connector, rather than reflecting real cache-miss behavior.

This change mirrors the existing per-goroutine "cancelled" guard (`fanCtx.Err() != nil`) in the post-fan-out block: it tracks an `aborted` flag set when the drain loop bails on `fanCtx.Done()`, and skips the miss metric (falling through to upstream exactly as before) when no connector confirmed a genuine miss. Genuine all-connector misses are still recorded.

## Testing

- New regression test `TestEvmJsonRpcCache_FanOut_CancelledFanOutIsNotAMiss` asserts no `success_miss` increment when the parent context is cancelled. It fails without the fix (count 1) and passes with it (0).
- Full `architecture/evm` suite and all `erpc` cache tests pass.

---

> [!NOTE]
> **Low Risk**
> Observability-only change for aborted fan-outs; request path still falls through to upstream the same way, and real cache misses keep being recorded.
> 
> **Overview**
> When parallel cache reads exit because **`fanCtx`** is done (caller cancel/deadline or the 30s fan-out backstop) and no hit is found, **`EvmJsonRpcCache.Get`** no longer increments **`cache_get_success_miss_total`** or records miss duration. It sets trace **`cache.miss_reason=cancelled`** and still returns **`nil, nil`** so upstream handling is unchanged.
> 
> The drain loop sets an **`aborted`** flag on the **`fanCtx.Done()`** path (after the buffered winner race drain). Genuine all-connector misses still use the existing miss metrics.
> 
> Adds **`TestEvmJsonRpcCache_FanOut_CancelledFanOutIsNotAMiss`** (ERPC-553) to assert no **`success_miss`** increment when a short parent timeout cancels hanging connector reads.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3521309b599dbf2dac37e4ab68fd6fb9418ecc9c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/architecture/evm/json_rpc_cache.go
+++ b/architecture/evm/json_rpc_cache.go
@@ -370,6 +370,7 @@ func (c *EvmJsonRpcCache) Get(ctx context.Context, req *common.NormalizedRequest
 		lastMiss   *fanResult
 		lastReject *fanResult
 		lastError  *fanResult
+		aborted    bool
 	)
 drain:
 	for received := 0; received < spawned && jrr == nil; {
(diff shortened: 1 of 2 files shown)


aramalipoor (approved):

hawkaa:
@aramalipoor i see the test is failing, but I'm not convinced it has anything to do with my change. How should we proceed with this PR?

PR #951, opened 2026-06-18
fix(blockdaemon): remove implicit hostname-based vendor detection

andreclaro:
## Problem

`BlockdaemonVendor.OwnsUpstream` matched any upstream whose endpoint URL contained `svc.blockdaemon.com` as a substring, silently activating the vendor module. Once activated, `GenerateConfigs` required `apiKey` unconditionally — so users who provided a full `https://svc.blockdaemon.com/...` endpoint with their own `Authorization` header got an \`apiKey is required\` error, and any pre-configured auth was discarded.

## Fix

Keep hostname detection in `OwnsUpstream` (consistent with chainstack, blockpi, quicknode) but gate the `apiKey` requirement on `endpoint == ""` — the same pattern chainstack uses:

- **`endpoint` is empty** (eRPC constructs the URL): `apiKey` is required; eRPC builds the `svc.blockdaemon.com` URL and injects `Authorization: Bearer <apiKey>`.
- **`endpoint` is provided** (user brings their own URL): `apiKey` is optional; if present it injects the auth header (unless one is already set); if absent the pre-configured `Authorization` header is preserved unchanged.

```go
func (v *BlockdaemonVendor) OwnsUpstream(ups *common.UpstreamConfig) bool {
    return strings.HasPrefix(ups.Endpoint, "blockdaemon://") ||
        strings.HasPrefix(ups.Endpoint, "evm+blockdaemon://") ||
        strings.Contains(ups.Endpoint, "svc.blockdaemon.com")
}
```

## Testing

- Updated `TestBlockdaemonVendor_OwnsUpstream`: `svc.blockdaemon.com` HTTPS endpoints are correctly claimed.
- Added `passes_through_preset_endpoint_with_apiKey`: auth header injected when `apiKey` provided alongside a full endpoint.
- Added `passes_through_preset_endpoint_without_apiKey`: pre-configured `Authorization` header preserved when no `apiKey` is set.

### DIFF
--- a/thirdparty/blockdaemon.go
+++ b/thirdparty/blockdaemon.go
@@ -82,12 +82,11 @@ func (v *BlockdaemonVendor) GenerateConfigs(ctx context.Context, logger *zerolog
 		upstream.JsonRpc = &common.JsonRpcUpstreamConfig{}
 	}
 
-	apiKey, ok := settings["apiKey"].(string)
-	if !ok || apiKey == "" {
-		return nil, fmt.Errorf("apiKey is required in blockdaemon settings")
-	}
-
 	if upstream.Endpoint == "" {
+		apiKey, _ := settings["apiKey"].(string)
+		if apiKey == "" {
+			return nil, fmt.Errorf("apiKey is required in blockdaemon settings when endpoint is not set")
+		}
 		if upstream.Evm == nil {
 			return nil, fmt.Errorf("blockdaemon vendor requires upstream.evm to be defined")
 		}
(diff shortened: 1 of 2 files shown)


thirdparty/blockdaemon.go
```diff
@@ -148,9 +148,5 @@ func (v *BlockdaemonVendor) GetVendorSpecificErrorIfAny(req *common.NormalizedRe
 }
 
 func (v *BlockdaemonVendor) OwnsUpstream(ups *common.UpstreamConfig) bool {
-	if strings.HasPrefix(ups.Endpoint, "blockdaemon://") || strings.HasPrefix(ups.Endpoint, "evm+blockdaemon://") {
-		return true
-	}
-
-	return strings.Contains(ups.Endpoint, "svc.blockdaemon.com")
```
kasrakhosravi, line 155: chainstack/blockpi/quicknode keep this same hostname match — cleaner to gate the `apiKey` check in `GenerateConfigs` on `endpoint == ""` (like chainstack) than to drop detection here.

thirdparty/blockdaemon.go
```diff
@@ -111,11 +111,13 @@ func (v *BlockdaemonVendor) GenerateConfigs(ctx context.Context, logger *zerolog
...
-	if upstream.JsonRpc.Headers == nil {
-		upstream.JsonRpc.Headers = make(map[string]string)
-	}
-	if _, exists := upstream.JsonRpc.Headers["Authorization"]; !exists {
-		upstream.JsonRpc.Headers["Authorization"] = "Bearer " + apiKey
+	if apiKey != "" {
+		if upstream.JsonRpc.Headers == nil {
+			upstream.JsonRpc.Headers = make(map[string]string)
+		}
+		if _, exists := upstream.JsonRpc.Headers["Authorization"]; !exists {
```
andreclaro: This can also be lower case ... fixing it

andreclaro: other thirdparty modules should also inject apiKey into the header whenever possible :)

PR #953, opened 2026-06-19
feat: add project-level allowClientDirectives filter

jscaltreto:
## Summary

eRPC's 23 `X-ERPC-*` request directives (skip-cache-read, use-upstream, skip-consensus, etc.) are unconditionally honored from any client — there's no way to restrict which directives a project accepts. For production-facing deployments, this means any client can bypass caching, pin to specific upstreams, or skip consensus validation. This PR adds a project-level `allowClientDirectives` config field that filters which directives are honored from HTTP requests, using the existing WildcardMatch pattern grammar.

## Detail

- **New config field**: `projects[].allowClientDirectives` (`*string`) on `ProjectConfig`. Nil (omitted) = all allowed (backward-compatible). Empty string = none allowed. Supports full WildcardMatch grammar: `"!skip-cache-read & !use-upstream"` blocks those two while allowing everything else.
- **Pre-compiled matcher**: The pattern is compiled once at project registration via `NewWildcardMatcher` (new addition to `common/matcher.go`) into a `MatcherFunc` closure stored on `PreparedProject`. No per-request tokenization or parsing.
- **Filtering in `EnrichFromHttp`**: Two closures (`getHeader`, `getQueryArg`) wrap the existing header/query reads. Each checks whether the value is non-empty first (short-circuit `&&`), so the matcher is only called when the client actually sent that directive — no cost on the common path where a directive header is absent.
- **`isDirectiveAllowed` method**: On `NormalizedRequest`. Nil matcher = allow all, empty query key with non-nil matcher = deny (fail-closed for unregistered keys), otherwise calls the compiled matcher.
- **`SetAllowClientDirectiveMatcher` setter**: Stores the `MatcherFunc` on `NormalizedRequest` before `EnrichFromHttp` runs, keeping the `EnrichFromHttp` signature unchanged — no changes to existing callers or test files.
- **Startup validation**: `ProjectConfig.Validate` calls `NewWildcardMatcher` to catch invalid patterns at load time.
- **`X-ERPC-Force-Trace` not filtered**: Force-trace bypasses OTel sampling at span creation in `StartHTTPServerSpan`, which runs before project resolution. Filtering it would require deferring the sampling decision — left as a follow-up.
- **Config-set `directiveDefaults` unaffected**: `ApplyDirectiveDefaults` runs before `EnrichFromHttp` and is not gated by the filter. Operator-configured defaults always apply.

## Testing

- 25 unit tests for `isDirectiveAllowed` covering all pattern permutations: nil, empty, wildcard, exact match, OR, negation, AND+negation, glob, glob+OR, grouped negation.
- 6 integration tests for `EnrichFromHttp` with filtering: nil allows all, empty blocks all, negation blocks specific directive, query params filtered, user-agent always extracted, directive defaults survive blocking.
- All existing tests pass unchanged (no signature changes to `EnrichFromHttp`).
- `go build ./...` clean, `go test ./common/...` green.

## Documentation

- `docs/pages/config/projects.mdx`: New row in config schema table, two new edge case entries (#27, #28), added to AISection key list.
- `docs/pages/operation/directives.mdx`: Visible body note about `allowClientDirectives`, updated parsing pipeline description in AISection, new best practice bullet, new edge case entry (#19), new `<PromptExample>` block.


common/config.go
```diff
@@ -570,10 +570,11 @@ type ProjectConfig struct {
 	Networks         []*NetworkConfig  `yaml:"networks,omitempty" json:"networks"`
 	RateLimitBudget  string            `yaml:"rateLimitBudget,omitempty" json:"rateLimitBudget"`
 	// Configure user agent tracking at the project level
-	UserAgentMode  UserAgentTrackingMode `yaml:"userAgentMode,omitempty" json:"userAgentMode"`
-	ForwardHeaders []string              `yaml:"forwardHeaders,omitempty" json:"forwardHeaders"`
-	IgnoreMethods  []string              `yaml:"ignoreMethods,omitempty" json:"ignoreMethods"`
-	AllowMethods   []string              `yaml:"allowMethods,omitempty" json:"allowMethods"`
+	UserAgentMode         UserAgentTrackingMode `yaml:"userAgentMode,omitempty" json:"userAgentMode"`
+	ForwardHeaders        []string              `yaml:"forwardHeaders,omitempty" json:"forwardHeaders"`
+	AllowClientDirectives *string               `yaml:"allowClientDirectives,omitempty" json:"allowClientDirectives"`
```
andreclaro, line 575: Should we have validation of `allowClientDirectives` at load time?  https://github.com/erpc/erpc/blob/main/common/validation.go

see RateLimiterConfig example: https://github.com/erpc/erpc/blob/main/common/validation.go#L164

common/request.go
```diff
@@ -728,166 +765,182 @@ func (r *NormalizedRequest) EnrichFromHttp(headers http.Header, queryArgs url.Va
 	}
 
 	// Headers have precedence over directive defaults, but should only override when explicitly provided.
-	if hv := headers.Get(headerDirectiveRetryEmpty); hv != "" {
+
+	getHeader := func(key string) string {
+		if hv := headers.Get(key); hv != "" && isDirectiveAllowed(headerToQueryKey[key], r.allowClientDirectives) {
```
andreclaro: minor: should we have a way to avoid avoid calling `isDirectiveAllowed` multiple times per request ?

We will be calling this function X time per request, where X is the number of defined directives in the request, right?  🤔

andreclaro: anyway, this doesn't impact us since we will configure `allowClientDirectives` to `""`

jscaltreto: it kind of has to be called for each directive in order to validate that the directive is allowed. But it's only called for directives that are actually supplied in the request. So if the user includes only one directive it just gets called the one time. 

The current design does re-parse the matcher with every call though, which could get expensive. Ideally we would compile the match pattern once per project and reuse it for every call.

jscaltreto: Refactored to pre-compile the matcher function at project registration time. Each header/query directive still calls `isDirectiveAllowed`, but it's much lighter now.

andreclaro (approved):

PR #948, opened 2026-06-18
feat(evm): data-integrity module — spec + unified check engine

aramalipoor:
Spec **and** implementation for the data-integrity module, in one PR. Built incrementally — every commit green.

## Implementation (foundation; `architecture/evm/integrity/`)

A dedicated package replaces scattered, duplicated inline validation with one clean framework:

- **`Decoded`** — the response result decoded **once** into lightweight EVM views (header/transactions/receipts/logs). Every check reads from it; none re-parse (today each validator parses into its own structs).
- **`Check` + registry** — small, self-registering validation units keyed by method.
- **`CheckSet`/`CheckConfig`** — the resolved selection (enabled + params) the engine consumes, kept independent of where it came from (directives today; the `level`/`enforceFrom`/generic-header model next).
- **`Validate`** — the engine: selects enabled+applicable checks, runs them over the single decode, maps the first violation to `ErrEndpointContentValidation`.

Separation of concerns: the integrity package owns validation; the `evm` package keeps only the hook wiring + a directive→`CheckSet` adapter. This commit migrates the index-magnitude check, deletes the standalone `evm/integrity.go`, and wires the hook through the engine. Behavior preserved; full `evm` suite green.

**Incremental plan (each green):** (1) ✅ foundation + `indexMagnitude`; (2) migrate `eth_getBlockReceipts` checks (field-lengths, bloom, contiguity, embedded-identity, counts) → registry, removing the duplicated inline helpers; (3) same for `eth_getBlockByNumber`; (4) new catalog checks + the config model; (5) authoritative tier (resolver + `network.Forward` force-fetch) + cross-block continuity.

## Design (the spec — `specs/data-integrity/`)

- **Orthogonal to consensus**; deliberately **no quorum level**.
- **One knob, four levels:** `off`/`intrinsic`/`corroborated`/`authoritative`; levels 0–2 add no upstream calls.
- **Data-point resolver** — use already-available data, force-fetch only what's missing and only at `authoritative` (via `network.Forward`, inheriting the network's failsafe).
- **Check catalog (§6)** — five families (commitment recompute, per-item authenticity, structural, shape, cross-block continuity) + completeness + encoding correctness + retryable-vs-deterministic failure classes. Every check independently flagged/parameterized.
- **Per-chain nuances (Appendix A)** — chain-specific edge-cases grouped by behavior class.
[...]

PR #968, opened 2026-06-30
feat(selection-policy): per-boundary eval axis (block-availability lanes)

aramalipoor:
## What

Adds an orthogonal **`evalPerBoundary`** axis to the selection policy. When enabled, a single-block request is evaluated against the **lane** of upstreams whose configured block-availability bounds can actually serve the request's block. An upstream that can't serve the range is **absent from the pool** (a capability fact), and per-lane health/exclusion state is kept separate from other lanes.

This is useful when a network mixes upstreams with different served ranges — e.g. archive nodes alongside pruned / recent-only / snap-synced nodes that only retain block bodies above some floor. Today such a node can be ranked into a pool for a historical block it can't actually serve; the policy then scores and excludes over a population that includes upstreams that can't answer, and a separate per-upstream availability filter hides them afterward. The boundary axis makes the policy operate on the population that can actually serve the block.

## Design

- **A lane is identified by its eligible upstream *set*** (sorted-id key), not by a block range — so the slot key is stable as numeric bounds drift; it changes only when membership changes. Non-adjacent ranges with identical membership coalesce onto one slot.
- **Not folded into `EvalScope`.** Boundary must not change the health-tracker grain (it's a decision/pool axis, not a metrics axis), so the engine reads `EvalPerBoundary` directly. No new metric label dimensions.
- **One source of truth for reachability.** Lane membership reuses the same `EvmBlockAvailabilityBounds` predicate as the per-upstream availability gate, so the two never disagree.
- **Additive and safe.** Off by default; when off, routing is identical to before. Requests with no resolvable block number (e.g. `eth_getBlockByHash`) and all-eligible requests fall back to the full-pool wildcard slot. Lane slots are narrow and reclaimed by the idle sweep, so lane-key cardinality can't grow the slot map without bound.

## Changes

- `common/config.go`: `SelectionPolicyConfig.EvalPerBoundary *bool`.
- `internal/policy`: `slotKey` gains a `boundary` dimension; `GetOrderedInLane` + `PerBoundaryEnabled`; lane slots get a membership-filtered `upstreamsFn`; idle-sweep exemption tightened to the all-wildcard slot.
- `erpc/networks.go`: derive a request's eligible lane (`eligibleUpstreamIDsForBoundary` + pure `eligibleLane`) and pass it on the forward path, gated by `PerBoundaryEnabled`.

## Tests

Lane-key canonicalization/stability, boundary gating across scopes, lane pool scoping, axis-off no-op, idle eviction of lane slots, and the lane-derivation collapses (no-bounds / all-eligible → full-pool; out-of-range exclusion).

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -2440,7 +2440,19 @@ type SelectionPolicyConfig struct {
 	EvalPerMethod *bool `yaml:"evalPerMethod,omitempty" json:"evalPerMethod,omitempty" tstype:"-"`
 	// EvalPerFinality — same shape and translation behavior as
 	// EvalPerMethod, for the finality axis.
-	EvalPerFinality *bool    `yaml:"evalPerFinality,omitempty" json:"evalPerFinality,omitempty" tstype:"-"`
+	EvalPerFinality *bool `yaml:"evalPerFinality,omitempty" json:"evalPerFinality,omitempty" tstype:"-"`
+	// EvalPerBoundary scopes selection-policy evaluation by block-availability
+	// "lane" — the set of upstreams whose configured block range can actually
+	// serve the request's block. Unlike EvalPerMethod / EvalPerFinality this is
+	// deliberately NOT folded into EvalScope: it must not change the
+	// health-tracker grain (boundary is a decision/pool axis, not a metrics
+	// axis), so the engine reads it directly as an orthogonal slot dimension.
+	// When on, a request whose block excludes some upstream is evaluated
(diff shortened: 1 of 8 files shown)


aramalipoor:
Superseded: folded into the data-integrity branch (#948) so per-boundary rides the same shadow image and we can test both against shadow traffic together. Commit cherry-picked as e7689dc2 on `claude/data-integrity-spec`.

PR #972, opened 2026-07-01
feat(clients): share http.Transport connection pool across projects per upstream

kasrakhosravi:
## Summary

Each eRPC project builds its own `ClientRegistry`, so a deployment that runs **multiple projects declaring the same upstream endpoint** ends up with one full `http.Transport` per (project, upstream) — i.e. N independent idle-connection pools (`MaxIdleConns=1024` / `MaxIdleConnsPerHost=256`), TLS session caches, and per-host buffers for a single physical endpoint.

This PR shares one connection pool per endpoint across projects.

## What changes

- New `clients/transport_pool.go`: a process-global `TransportPool` keyed by `UniqueUpstreamKey`.
- `NewGenericHttpJsonRpcClient` obtains its `*http.Transport` from the pool instead of allocating one.

Two projects that declare the identical upstream (same id + endpoint + headers → same `UniqueUpstreamKey`) now share **one** `http.Transport`. Distinct upstreams within a project still get distinct pools, so per-upstream isolation is unchanged; only cross-project duplication is removed.

## Why it's non-breaking

- **No signature changes.** The pool is package-internal to `clients`, so no registry/constructor call sites move.
- **Test path untouched.** `util.IsTest()` still returns `http.DefaultTransport`, so HTTP mocking (gock/httpmock, which patch `DefaultTransport`) keeps intercepting; the pool is production-only.
- **Client identity unchanged.** projectId `User-Agent`, upstream labels, spans, and metrics are all still per-project — only the underlying transport is shared, and `http.Transport` is safe for concurrent use.

## Not included (on purpose)

Sharing the EVM state poller. Its expensive work (latest/finalized/earliest block polling) is **already** deduplicated across projects via the `SharedState` counters (keyed by `UniqueUpstreamKey`, which omits projectId, + `TryUpdateIfStale` skipping the fetch when fresh), and each project's tracker already receives the shared block numbers via the counter's `OnValue` fan-out. The only remaining per-project poller state is `eth_syncing` (`EvmSyncingState`), which gates upstream selection (`networks.go`, `upstream.go`, `upstream_executor.go`) — sharing the poller goroutine would require moving that into `SharedState` and touching the selection path, so it's left per-project.

## Testing

`clients/transport_pool_test.go` (run with `-race`): same-key dedup, distinct-key isolation, empty key, high-concurrency single/mixed keys, transport-config pinning, the package-global instance, and end-to-end dedup through `UniqueUpstreamKey`.

- `go build ./...` ✅
- `go vet ./...` ✅
- `go test -race ./clients/` ✅

---

> [!NOTE]
> **Medium Risk**
> Introduces global shared outbound connection state keyed by upstream identity; behavior is standard for `http.Transport` but affects all co-resident projects hitting the same endpoint.
> 
> **Overview**
> **Production HTTP JSON-RPC clients no longer allocate a dedicated `http.Transport` per project/upstream pair.** Outbound tuning is centralized in `newDefaultTransport()` and reused via a process-wide `TransportPool` keyed by `common.UniqueUpstreamKey` (id, endpoint, network, headers).
> 
> `NewGenericHttpJsonRpcClient` now sets `http.Client.Transport` from `sharedTransportPool.GetOrCreate(...)` instead of building a transport inline. **Distinct upstreams still get distinct pools**; only identical upstream definitions across co-resident projects share one idle-connection set and TLS cache. The **test path is unchanged** (`http.DefaultTransport` when `util.IsTest()`), so gock/httpmock keep working.
> 
> Adds `transport_pool_test.go` covering deduplication, per-key isolation, concurrency under `-race`, transport config pinning, and `UniqueUpstreamKey` end-to-end behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5459a21828287764bbe2a675eeebbd39a8ce9f4f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/clients/http_json_rpc_client.go
+++ b/clients/http_json_rpc_client.go
@@ -102,30 +102,21 @@ func NewGenericHttpJsonRpcClient(
 		errorExtractor:  extractor,
 	}
 
-	// Default fallback transport (no proxy). Optimized for high-latency,
-	// high-RPS scenarios to prevent connection churn. DialContext via
-	// util.DefaultOutboundDialer enables kernel-level TCP keepalive so
-	// wedged outbound flows are detected within ~45s (3 missed probes)
-	// instead of the OS default tcp_keepalive_time of 2h on Linux.
-	transport := &http.Transport{
-		DialContext:           util.DefaultOutboundDialer().DialContext,
-		MaxIdleConns:          1024,
-		MaxIdleConnsPerHost:   256,
-		MaxConnsPerHost:       0, // Unlimited active connections (prevents bottleneck)
-		IdleConnTimeout:       90 * time.Second,
-		ResponseHeaderTimeout: 30 * time.Second,
-		TLSHandshakeTimeout:   10 * time.Second,
-		ExpectContinueTimeout: 1 * time.Second,
-	}
-
 	if util.IsTest() {
+		// Tests rely on http.DefaultTransport so HTTP mocking (gock/httpmock,
+		// which patch DefaultTransport) can intercept outbound requests.
 		client.httpClient = &http.Client{
 			Transport: http.DefaultTransport,
 		}
 	} else {
+		// Share ONE *http.Transport (connection pool) per unique upstream across
+		// every project. Two projects configuring the identical endpoint reuse the
+		// same pool instead of each holding a full idle-connection set + TLS cache.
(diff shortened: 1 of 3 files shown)


aramalipoor:
lgtm

PR #977, opened 2026-07-02
feat(bds): chain-identity enforcement + connection freshness for gRPC BDS clients

aramalipoor:
## Why

A gRPC BDS endpoint can silently become **cross-wired**: a DNS record or load balancer briefly pointing at another chain's server (stale records, reused addresses behind `dns:///` targets). Today nothing on the request path detects this — chainId is validated once at upstream bootstrap through one pooled connection, so a later cross-wire serves another chain's heads and data until the connection happens to break. #976 makes the tracker recover from a bogus head sample; this PR prevents bogus heads/data from being accepted at all, and bounds how long any stale connection can live.

## What

**Per-request chain assertion** — the client stamps its configured `chainId` into every BDS request that carries the optional field (`GetBlockByNumber/ByHash`, `GetLogs`, `GetTransactionByHash/Receipt`, `GetBlockReceipts`, all `Query*`). Chain-aware servers reject mismatches per request — atomic, survives mid-connection re-targeting. `eth_chainId` responses are also checked and return an error on mismatch.

**Armed from the very first request, on every path:**
- *configured chainId*: stored in the constructor before the pool exists — the first request ever is stamped;
- *auto-detected chainId (config 0)*: armed in upstream bootstrap right after detection — no request can route earlier (the upstream's networkId derives from the chainId);
- *cache connector*: armed with the probed chainId **before** the client is published into `clientByNetwork` — connector traffic only reaches the client through that map.

**Fail-closed on detection** — the pool maintainer re-verifies the ChainId behind each connection every 60s; a mismatch **closes the connection first** (every in-flight and future RPC on it errors immediately — nothing more can be served even by a server that doesn't validate chainId), then swaps in a verified replacement; a still-closed slot is retried every tick. A *proven* mismatch at pool construction fails the client outright. No new metrics: mismatches log at ERROR and surface as request errors through the existing error tracking.

**Connection max-age** — pooled conns are recycled after 5m (±20% jitter, one age-recycle per tick, 30s linger for in-flight RPCs), so re-resolution keeps conns aligned with current DNS truth.

**Server keepalive** — erpc's own gRPC server sets `MaxConnectionAge 10m + Grace 30s` (+ enforcement policy permitting 30s client pings) so its consumers re-resolve periodically too.

## Notes

- `SetExpectedChainId` is a concrete method on `*GenericGrpcBdsClient` (narrow type assertion at call sites), NOT an addition to the `GrpcBdsClient` interface — existing implementations stay compatible.
- No behavior change for servers that ignore the optional chainId field or clients without a configured/probed chainId (enforcement stays disarmed at 0).
- Dial/verify happen outside the pool lock, so `Pick()` is never blocked behind a probe.

## Validation

`go build ./...` clean; `clients`, `data`, `upstream`, and the erpc bootstrap/query shards pass locally; rebased on #976.

### DIFF
--- a/architecture/evm/evm_state_poller.go
+++ b/architecture/evm/evm_state_poller.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"runtime/debug"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -459,6 +460,12 @@ func (e *EvmStatePoller) PollLatestBlockNumber(ctx context.Context) (int64, erro
 		e.latestBlockFailureCount = 0
 		e.stateMu.Unlock()
 
+		// A major move must pass a fresh chain-identity check before entering
+		// the shared counter / tracker (see verifyChainIdOnMajorHeadMove).
+		if !e.verifyChainIdOnMajorHeadMove(ctx, "latest", e.latestBlockShared.GetValue(), blockNum) {
+			return 0, nil
+		}
+
 		// Directly update tracker with the correct timestamp for this locally-fetched block
 		// This happens BEFORE the OnValue callback is triggered, ensuring only the fetching node emits the metric
 		e.tracker.SetLatestBlockNumber(e.upstream, blockNum, blockTimestamp)
@@ -486,7 +493,6 @@ func (e *EvmStatePoller) SuggestLatestBlock(blockNumber int64) {
 			Msg("skipping latest block suggestion as it's not newer")
 		return
 	}
-
 	newValue := e.latestBlockShared.TryUpdate(e.appCtx, blockNumber)
 	e.logger.Trace().
 		Int64("blockNumber", blockNumber).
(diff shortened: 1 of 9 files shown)


aramalipoor:
Fixed the two CI failures:
- **units**: `SetExpectedChainId` is no longer part of the `GrpcBdsClient` interface (existing implementations — including test doubles injected via reflection — stay compatible); the cache connector arms it via a narrow type assertion instead.
- **security**: `#nosec G404` on the recycle-jitter `math/rand` use (spread jitter, not security-sensitive), matching repo precedent.

aramalipoor:
Added a second commit per discussion: **chain-identity verification on major head moves** — a polled head that moves beyond `common.DefaultToleratedBlockHeadRollback` (1024, either direction) must pass a fresh `eth_chainId` probe before entering the shared counter/tracker. Proven mismatch → sample rejected + upstream **cordoned** (existing primitive; `erpc_uncordonUpstream` restores it); probe failure → sample dropped for one poll cycle; match → accepted unchanged (deep reorgs / catch-ups cost one extra cheap call). Out-of-band `Suggest*` moves beyond the tolerance are dropped and re-confirmed by the poller within a debounce. No new config, no new metrics — reuses the shared tolerance constant, `EvmGetChainId`, cordon, and a typed `ErrEndpointChainIdMismatch` so the BDS client's own mismatch detection counts as proof.

Complements #976: that heals the tracker *after* a bogus sample; this keeps the bogus sample out entirely.

aramalipoor:
CI fix: dropped the `Suggest*` out-of-band gates from the major-move guard — suggestion-driven upward bursts are designed behavior (response enrichment, halted-chain resume bursts; pinned by `TestServedTip_NoNetworkLevelStickiness` and `TestEnrichment_AfterInterpolation_UsesOriginalLatestTagIntent`) and can't be verified in those hot paths. A bogus suggestion self-heals within one poll cycle (the next verified poll's >tolerance rollback is accepted as a correction per #976 semantics). The poll-path identity gate — where the real protection lives — is unchanged.

PR #982, opened 2026-07-07
fix: retry trace/debug "insufficient funds" toward the network

kasrakhosravi:
## Problem

`trace_block` and other tracing methods can return `"insufficient funds for gas * price + value"` for a transaction that was **mined** in the traced block — i.e. it provably had funds at execution time. This is a **state-reconstruction artifact**: tracing re-executes historical transactions against the parent block's pre-state, and when a node can't fully materialize that state trie it computes a wrong/zero balance and emits `insufficient funds`.

The error is transient **across nodes** — an upstream that holds the state (e.g. an archive node, or a fresh full node) traces the same block correctly. But eRPC currently returns it to the client with `attempts: 1, retries: 0, hedges: 0` — no failover.

## Root cause

`ExtractJsonRpcError` classifies `"insufficient funds"` / `"insufficient balance"` as `ErrEndpointExecutionException` with `retryableTowardNetwork: false` for **all** methods. That `false` propagates into both failover gates:

- `network_executor.shouldRetryWithReason` returns `""` for a non-retriable execution exception → **no retry**.
- `network_executor.runHedge`'s `keep(err)` returns `!IsRetryableTowardNetwork(err)` = `true` → the hedge race **stops on the error**, cancelling in-flight siblings.

So no `retry` / `hedge` / `consensus` policy can recover it — the classification is terminal. PR #834 introduced this branch with the correct intent ("non-retriable **for `eth_sendRawTransaction`**"), but the implementation dropped the method guard, making it non-retriable for every method.

## Fix

Scope retryability to tracing methods (`trace_*`, `debug_*`, `eth_trace*`), mirroring the method-aware pattern already used by the adjacent "out of gas" / "Transaction rejected" branch. Writes (`eth_sendRawTransaction`) and live simulations (`eth_call` / `eth_estimateGas`) keep the deterministic, non-retried behavior — for those, `insufficient funds` is a real result, not a reconstruction artifact.

## Test coverage

- New `TestExtractJsonRpcError_InsufficientFunds_TracingMethodsRetryable` (table-driven): `trace_block` / `debug_traceTransaction` / `eth_traceBlock` → retryable toward network; `eth_sendRawTransaction` / `eth_call` → non-retried.
- Existing `TestNetwork_SendRawTransaction_Idempotency/InsufficientFundsIsNotRetriedAcrossUpstreams` still passes (#834 behavior preserved).
- `docs/pages/reference/errors.mdx` error-classification table updated.

---

> [!NOTE]
> **Low Risk**
> Localized change to EVM error normalization and retry flags; writes and live simulations are explicitly excluded, with tests locking in both paths.
> 
> **Overview**
> **Tracing `trace_*` / `debug_*` / `eth_trace*` requests that return "insufficient funds" or "insufficient balance" are now classified as network-retryable** in `ExtractJsonRpcError`, so eRPC can fail over to another upstream (e.g. an archive node) instead of stopping after one attempt.
> 
> The change treats those replies as state-reconstruction artifacts when re-executing mined transactions, not as real balance failures. **`eth_sendRawTransaction`, `eth_call`, and `eth_estimateGas` keep the prior behavior** — insufficient funds stays non-retried as a deterministic execution result.
> 
> Adds table-driven coverage in `error_normalizer_test.go` and updates rule 13 in `docs/pages/reference/errors.mdx` for the method-scoped retry flag.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 8ad9b9fc1f23d4eeb63ca84911fec827ac20395f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -349,7 +349,7 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 
 		if strings.Contains(msg, "insufficient funds") ||
 			strings.Contains(msg, "insufficient balance") {
-			return common.NewErrEndpointExecutionException(
+			execErr := common.NewErrEndpointExecutionException(
 				common.NewErrJsonRpcExceptionInternal(
 					int(code),
 					common.JsonRpcErrorTransactionRejected,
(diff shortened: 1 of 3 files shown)


sfsf9797:
lgtm!!

aramalipoor (approved):

PR #984, opened 2026-07-08
fix(upstream): force-poll tip before rejecting on a stale configured upper bound

kasrakhosravi:
## Problem

A configured `blockAvailability.upper` bound is derived from the state poller's last-known latest block, which refreshes only once per `statePollerInterval`. On fast-producing chains the cached bound can lag the network tip, so a request for a freshly-produced block is rejected **before forwarding** — even though the upstream already has the data. The selection layer never sees the request and a different upstream serves it instead.

The legacy `blockHead` confidence path already guards against this with a **force-poll-once** escape when `forceFreshIfStale` is set (`upstream.go`, *"if the requested block is beyond the current latest block, it will force-poll the latest block number once"*). That escape was unreachable once a configured upper bound short-circuited the check first:

```go
minBound, maxBound := u.resolveAvailabilityBounds()
// ...
if maxBound != math.MaxInt64 && blockNumber > maxBound {
    // immediately returns false — never reaches the force-poll below
    return false, nil
}
```

## Change

Extend the same freshness escape to the configured-bound path: when the bound rejects and `forceFreshIfStale` is set, poll the tip once, recompute the bound, and re-evaluate before rejecting.

- Refactor `resolveAvailabilityBounds` → `resolveAvailabilityBoundsWithLatest(latest)` so the recomputation uses the freshly-polled tip without mutating poller state.
- Only the **upper** bound is re-evaluated. Lower bounds derived from `latest` can only *tighten* as `latest` advances, so re-polling them would never admit a previously-rejected block.
- When the fresh bound still rejects, behavior is unchanged (metric incremented, `false, nil`).
- The poll is best-effort: on error or a non-positive result the original stale bound decides, so a transient poll failure degrades to the prior behavior rather than failing the request.

## Tests

New `TestEvmAssertBlockAvailability_ConfiguredUpperBoundFreshnessEscape`:
- `forceFresh` recovers a tip block the cached bound rejects
- without `forceFresh`, the stale bound still rejects (no behavior change)
- `forceFresh` does not over-admit past the fresh tip
- an explicit `LatestBlockMinus` margin is honored after the re-poll

Full `./upstream/` suite and the `./erpc/` availability/race tests pass.

## Risk

Minimal. The new code path runs only when (a) a configured upper bound is present and (b) the caller passed `forceFreshIfStale=true`. The no-config / no-forceFresh paths are byte-for-byte unchanged. Worst case on a poll failure is the previous stale-bound behavior.

---

> [!NOTE]
> **Low Risk**
> Narrow new path (configured upper bound + forceFreshIfStale); poll errors degrade to existing reject behavior with no new failure modes for default callers.
> 
> **Overview**
> Fixes **false upstream skips** when `blockAvailability.upper` is derived from a **stale** state-poller latest on fast chains: requests for the real tip were rejected before forward, so selection never tried that upstream.
> 
> When a block fails the configured **upper** bound and the caller sets **`forceFreshIfStale`**, `EvmAssertBlockAvailability` now **polls latest once**, recomputes bounds via new **`resolveAvailabilityBoundsWithLatest(latest)`** (optional tip override without mutating poller cache), and only then rejects—matching the existing blockHead freshness escape that configured bounds used to short-circuit.
> 
> Paths without a configured upper bound or without `forceFreshIfStale` are unchanged; poll failures keep the prior stale-bound behavior. Tests cover recovery at tip, no over-admission, and `LatestBlockMinus` margins.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e74a94978ed3dddd1405da6a8f9958be3e440d14. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/upstream/upstream.go
+++ b/upstream/upstream.go
@@ -999,21 +999,36 @@ func (u *Upstream) EvmAssertBlockAvailability(ctx context.Context, forMethod str
 		return false, nil
 	}
 	if maxBound != math.MaxInt64 && blockNumber > maxBound {
-		telemetry.MetricUpstreamStaleUpperBound.WithLabelValues(
-			u.ProjectId,
-			u.VendorName(),
-			u.NetworkLabel(),
-			u.Id(),
-			forMethod,
-			confidence.String(),
-		).Inc()
-		u.logger.Debug().
-			Int64("blockNumber", blockNumber).
-			Int64("maxBound", maxBound).
-			Str("method", forMethod).
-			Str("upstreamId", u.config.Id).
-			Msg("block rejected: above upper availability bound")
-		return false, nil
+		// Freshness escape: a configured upper bound derived from the state
+		// poller's last-known latest can lag the network tip on fast chains (the
+		// bound refreshes only every statePollerInterval). When the caller asks
+		// for fresh data, force-poll the tip once and recompute before rejecting
+		// — mirroring the legacy blockHead path below. Only the upper bound is
+		// re-evaluated; lower bounds can only tighten as latest advances.
+		if forceFreshIfStale && u.evmStatePoller != nil && !u.evmStatePoller.IsObjectNull() {
+			if freshLatest, err := u.evmStatePoller.PollLatestBlockNumber(ctx); err == nil && freshLatest > 0 {
+				if freshMin, freshMax := u.resolveAvailabilityBoundsWithLatest(freshLatest); freshMax != maxBound {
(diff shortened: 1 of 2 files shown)


aramalipoor:
Wouldn't this basically moot the concept of debounce? Probably lower debounce would be more reasonable in these scenarios?

Issue #991, opened 2026-07-11
Vendor proposal: Satelink (pay-per-call Polygon RPC, x402 + instant keys) — erpc deployments already using us


Satelink-Protocol:
Hi — we operate **Satelink** (`rpc.satelink.network`), a pay-per-call RPC gateway on Polygon. We're proposing Satelink as an erpc vendor, and offering founding-customer terms to erpc users. Two facts first, since this list rightly runs on evidence:

**erpc deployments already route through us.** Our production traffic shows three erpc user-agents consuming our public Polygon endpoint (~1,100 calls/day sustained over weeks) — presumably via custom upstream URLs, since we're not in `thirdparty/vendors_registry.go`. They hit our anonymous free tier (500/day/IP, then HTTP 402), which erpc's failover handles but which means those users silently lose an upstream mid-day.

**Our auth model fits erpc's vendor pattern:**
- `X-API-Key` header auth; keys are self-service: `POST /v1/machine/register {"mode":"instant"}` returns a key in one call (no wallet/email), or wallet-signature registration for prepaid accounts
- Pay-per-call: $0.00003/call USDT prepaid, or a $0.10 USDC x402 payment for a 1,000-call bundle — no subscriptions, so it prices well as a fallback-tier upstream that's idle most of the time
- Machine-readable service manifest: https://rpc.satelink.network/.well-known/satelink.json · live pricing/market data: https://rpc.satelink.network/v1/compare
- Measured p50 ~44ms, Polygon PoS (137) today; chain coverage expanding

**Ask:** would a `CreateSatelinkVendor()` PR be welcome? We're happy to write it to match the existing vendor implementations.

**Offer to erpc and its users:** founding-customer terms — a $1 verification payment gets a dedicated high-quota key (100× the free tier), direct operator support, and founding-infrastructure-partner status. We're early and optimizing for real integrations over revenue. Reply here or infra@satelink.network.

Satelink-Protocol:
Closing — this duplicated our own PR #981 (approved, mergeable), which delivers exactly this. Apologies for the double-filing; all discussion → #981. The founding-customer offer for erpc users stands as described there.

PR #981, opened 2026-07-05
feat(vendor): add Satelink native provider — satelink:// URL scheme

Satelink-Protocol:
# PR: feat: add Satelink third-party provider

> Submit at: https://github.com/erpc/erpc/compare — base `erpc:main`, from a
> fork branch containing `thirdparty/satelink.go` (this repo's
> `docs/erpc-adapter/satelink_provider.go`), the `vendors_registry.go`
> registration line, the `common/defaults.go` `buildProviderSettings` case,
> a `thirdparty/satelink_test.go`, and a providers.mdx docs entry — the same
> file set as #861.

---

## Summary

- Adds `SatelinkVendor` to `thirdparty/` — implements `SupportsNetwork`,
  `GenerateConfigs`, `GetVendorSpecificErrorIfAny`, and `OwnsUpstream`
  (same shape as the Blockdaemon vendor in #861)
- Registers the vendor in `thirdparty/vendors_registry.go` and adds the
  `satelink` case to `buildProviderSettings` in `common/defaults.go`
- Adds a configuration example to `docs/pages/config/projects/providers.mdx`
- Follow-up to #960, where we introduced the endpoint and offered this adapter

## What Satelink is

[Satelink](https://rpc.satelink.network/.well-known/satelink.json) is a
fault-tolerant Polygon PoS (chainId 137) RPC gateway: 16 upstream providers
behind latency-sorted routing with EMA weighting and 3-state circuit-breaker
failover. It has a keyless free tier (500 calls/day per IP) and a prepaid
paid tier ($0.00003/call, settled in USDT on Polygon) with fully autonomous
machine onboarding — registration, funding, and key issuance are all API-only,
no dashboard or human step. We already see erpc instances routing through the
plain HTTPS endpoint; this PR makes it a first-class provider.

## What this PR adds

**URL scheme** (`satelink://…`):

| Endpoint | Meaning |
|---|---|
| `satelink://<api_key>@polygon` | Paid tier — key sent as `X-API-Key` header |
| `satelink://free@polygon` | Keyless free tier (500 calls/day per IP) |

**Supported networks**

| Chain | ID | Resolved endpoint |
|---|---|---|
| Polygon PoS mainnet | 137 | `https://rpc.satelink.network/rpc/polygon` |

**Error mapping**

| Gateway response | erpc error |
|---|---|
[...]

PR #992, opened 2026-07-11
feat: add Satelink vendor (Polygon pay-per-call RPC)

Satelink-Protocol:
Adds Satelink (`rpc.satelink.network`) as a vendor, per the proposal in #991.

**Why:** three erpc deployments already route through our public Polygon endpoint via custom upstream URLs (~1,100 calls/day sustained — visible in our traffic as `erpc` user-agents). As anonymous traffic they hit our 500/day/IP free tier and silently lose the upstream mid-day; with a vendor + apiKey they get their own quota.

**Implementation** (modeled on `ankr.go` for structure and `blockdaemon.go` for header auth):
- Networks: Polygon PoS (137), Polygon Amoy (80002)
- Endpoint: `https://rpc.satelink.network/rpc/{network}`
- Auth: `X-API-Key` header via `upstream.JsonRpc.Headers` (keys are self-service — `POST /v1/machine/register {"mode":"instant"}` returns a key in one call, no signup; prepaid pay-per-call at $0.00003/call beyond the free tier, no subscription)
- `OwnsUpstream`: `satelink://` prefix or `rpc.satelink.network` substring

Config example:
```yaml
upstreams:
  - vendorName: satelink
    settings:
      apiKey: sk_...
    evm:
      chainId: 137
```

Service details (machine-readable): https://rpc.satelink.network/.well-known/satelink.json — happy to adjust anything to fit conventions, add tests mirroring `ankr`'s if wanted, and we're offering founding-customer terms (higher quotas, direct operator support) to erpc users per #991.

### DIFF
--- a/thirdparty/satelink.go
+++ b/thirdparty/satelink.go
@@ -1,0 +1,104 @@
+package thirdparty
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"strconv"
+	"strings"
+
+	"github.com/erpc/erpc/common"
+	"github.com/rs/zerolog"
+)
+
+var satelinkNetworkNames = map[int64]string{
+	137:   "polygon",
+	80002: "amoy",
+}
+
+type SatelinkVendor struct {
+	common.Vendor
+}
+
+func CreateSatelinkVendor() common.Vendor {
+	return &SatelinkVendor{}
+}
+
+func (v *SatelinkVendor) Name() string {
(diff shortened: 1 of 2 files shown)


Satelink-Protocol:
Closing — duplicate of #981, which is more complete (defaults case, tests, docs entry) and already approved. Apologies for the noise: different operator session on our side; consolidating everything into #981.

PR #997, opened 2026-07-14
fix(evm): gate major out-of-band head suggestions behind the chain-identity check

kasrakhosravi:
## Problem

`EvmStatePoller.PollLatestBlockNumber` already routes a *major* head move — a jump beyond `DefaultToleratedBlockHeadRollback` — through `verifyChainIdOnMajorHeadMove`, which makes a fresh `eth_chainId` call and cordons an upstream that turns out to be answering for a different chain.

The out-of-band `SuggestLatestBlock` / `SuggestFinalizedBlock` paths — fed by response enrichment and halted-chain resumes — were deliberately left ungated, on the assumption that a bogus suggestion self-heals within one poll cycle.

That assumption breaks when the suggesting upstream never returns a correct low sample afterwards: a cross-wired endpoint that starts answering with another chain's (much higher) block height and then only errors or keeps serving the wrong chain. Its bogus head enters the shared counter via the suggestion path, is persisted, and skews every lag-based routing decision derived from the network head until an operator intervenes.

## Change

Route a **major forward jump** from either suggestion path through the same `verifyChainIdOnMajorHeadMove` gate, off the hot path:

- **`SuggestLatestBlock`**: a jump beyond the tolerance is handed to a single background verification (`verifyThenSuggestLatestBlock`); at most one runs at a time per poller, and it applies the suggestion only if the chain-identity check passes. Small keep-fresh advances still apply inline with zero added latency.
- **`SuggestFinalizedBlock`**: already runs in its own goroutine, so the gate is added inline there.

A proven cross-wired endpoint is cordoned by the check itself; an unverifiable one (a transient `eth_chainId` failure) simply drops the suggestion and is re-observed on the next suggestion or verified poll. Non-major suggestions are unchanged, so the hot path stays non-blocking.

**Observable semantics change:** a major forward suggestion no longer takes effect synchronously — it lands after the background `eth_chainId` round-trip (typically one probe latency). Small advances, halted-chain resume bursts within the tolerance, and the verified poll path are unaffected. This deliberately trades a moment of tip latency on >tolerance out-of-band jumps for keeping a never-self-correcting cross-wired endpoint out of the shared counter.

## Tests

`architecture/evm/evm_state_poller_suggest_gate_test.go` covers:

- small advance applies inline and synchronously;
- major jump with a matching chain id is applied;
- major jump with a mismatching chain id is dropped **and** the upstream cordoned;
- major jump with a transient probe error is dropped **without** cordoning;
- major jump surfaced as a typed chain-id-mismatch error cordons;
- unpinned chain id (`ChainId <= 0`) skips the gate (cold start);
- the finalized-axis equivalents.

`erpc/networks_served_tip_test.go` — `TestServedTip_NoNetworkLevelStickiness` drives a >tolerance jump (9_000 → 11_000), so its assertion now waits (`require.Eventually`) for the verified suggestion to apply; the invariant it pins (zero network-level stickiness — the pick follows the heads) is unchanged. The sibling halted-chain-resume test stays synchronous: its 500-block burst is within tolerance and still applies inline.

---

> [!NOTE]
> **Medium Risk**
> Changes hot-path-adjacent head tracking and routing inputs (shared counters, cordoning) for legitimate large catch-up jumps, with async verification that can briefly delay major suggests; behavior is heavily tested but affects core upstream selection semantics.
> 
> **Overview**
> **Out-of-band block head suggestions** from response enrichment and similar paths now go through the same **`verifyChainIdOnMajorHeadMove`** protection as verified polls when the jump is **major** (forward by more than `DefaultToleratedBlockHeadRollback`). That blocks cross-wired upstreams from pinning a wrong-chain height in shared counters and skewing lag-based routing when the bad endpoint never self-corrects.
> 
> For **`SuggestLatestBlock`**, major jumps are handled by **`verifyThenSuggestLatestBlock`**: a single background goroutine per poller runs `eth_chainId`; small keep-fresh advances still **`TryUpdate` inline** with no extra latency. **`SuggestFinalizedBlock`** adds the same gate inside its existing async update path before **`TryUpdate`**.
> 
> Proven chain mismatch still **cordons** the upstream; transient `eth_chainId` failures **drop** the suggestion without cordoning. Docs on **`verifyChainIdOnMajorHeadMove`** are updated to match (no longer claiming suggest paths are ungated).
> 
> New **`evm_state_poller_suggest_gate_test.go`** covers latest/finalized suggest behavior; **`networks_served_tip_test`** waits asynchronously for a major verified suggest to apply.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 67b431dd59f544057ea413b6ebafa170d6bc342f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/architecture/evm/evm_state_poller.go
+++ b/architecture/evm/evm_state_poller.go
@@ -91,6 +91,13 @@ type EvmStatePoller struct {
 	// Track if updates are in progress to avoid goroutine pile-up
 	finalizedUpdateInProgress sync.Mutex
 
+	// Serializes the off-hot-path chain-identity verification for a MAJOR
+	// forward jump suggested out-of-band via SuggestLatestBlock. At most one
+	// verification is in flight per poller; a concurrent major suggestion is
+	// dropped and re-observed on the next suggestion (or verified poll). Small
+	// keep-fresh advances never touch this.
+	latestMajorVerifyInProgress sync.Mutex
+
 	// Earliest per probe tracking
 	earliestByProbe              map[common.EvmAvailabilityProbeType]data.CounterInt64SharedVariable
 	earliestSchedulerStarted     map[common.EvmAvailabilityProbeType]bool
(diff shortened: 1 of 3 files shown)


hawkaa (commented):
lgtm!
but very surprised this is even necessary... 😬

PR #998, opened 2026-07-14
fix(health): anchor the block-head lag reference to the corroborated freshest head

kasrakhosravi:
## Problem

Per-upstream block-head lag — the value `blockNumberLagAbove` / `blockSecondsLagAbove` read to exclude stale upstreams — is computed in the tracker as `networkMax - upstreamHead`, where `networkMax` is the single most-ahead upstream's head.

That lets one wrongly-ahead upstream define the tip. A cross-wired or otherwise misbehaving endpoint that briefly reports another chain's (much higher) height sets the max millions of blocks ahead of the real chain. Every *healthy* upstream then appears to lag by that margin, trips the lag-exclusion predicate, and is dropped — so the selection policy evicts the entire healthy majority and routing collapses onto the one bad upstream.

## Change

The lag reference is now the **corroborated freshest** head: the second-highest positive head across the network's upstreams — the same statistic the served-tip subsystem already designates as its lag reference (`evm.ServedTipPick.Freshest`), so both layers agree on what the trustworthy tip is. No single rogue can move a value that at least one other upstream corroborates. With fewer than 3 positive heads there is nothing to corroborate, so the raw max is kept (preserving one/two-upstream laggard detection).

It is computed in the tracker, over **all** observed upstreams, deliberately:

- the reference *feeds* the selection policy, so deriving it from the policy's own post-exclusion eligible set would be circular (an eviction would remove a head from the input and freeze the state that caused it);
- it covers default max-mode networks identically (served-tip serving is opt-in);
- `health` cannot import `architecture/evm`, and the serve-side clamps (guaranteed-methods floor, selector partitions) are correct for "what do we advertise" but would under-measure "how stale is this node";
- it stays a zero-allocation streaming scan on the per-sample hot path.

Because the second-highest can move while the network max stands still (the corroborating upstream catches up to the leader), a change of the reference now triggers the same global lag recompute a max advance does — otherwise other upstreams would hold lags computed against a stale reference until their own next sample.

**Semantics shift, pinned by tests:** lag now measures distance to the corroborated tip rather than to the single most-ahead node — heads `[1000, 998, 995]` read lags `0/0/3` instead of `0/2/5`. A lone leader's small lead no longer counts against its peers.

**Known residual, also pinned:** two or more upstreams agreeing on a bogus far-future head corroborate each other — inherent to any height-only corroboration. The absolute per-upstream maxima remain observable via `erpc_upstream_latest_block_number`.

## Tests

- `health/tracker_lag_reference_test.go` — the incident shape (one poisoned far-future head vs. an honest majority), genuine laggard still excluded, outlier + laggard together, single/two-upstream fallback, finalized axis, and the pure order-statistic across fan-outs and ties.
- `health/tracker_blockhead_rollback_test.go` — rollback re-derivation now asserts corroborated-freshest lags after correction.
- `erpc/networks_selection_policy_realpoll_test.go` — end-to-end real-poll path: laggard excluded, within-tolerance topology keeps all upstreams with the new lag values.
- `internal/policy/stdlib` observability leaf-attribution passes with the reference-motion recompute (previously an upstream could hold a stale lag when the reference moved under a stable max).

Full `./health/` (with `-race`), `./internal/policy/...`, and `./erpc/` suites pass.

---

> [!NOTE]
> **Medium Risk**
> Changes how lag feeds `blockNumberLagAbove` / routing exclusions on every poll; behavior is well-tested but alters production selection semantics and has known limits below 3 upstreams or with multiple agreeing bad heads.
> 
> **Overview**
> Fixes a routing failure where one far-ahead (poisoned) upstream made every honest peer look massively stale against `networkMax`, tripping `blockNumberLagAbove` and evicting the healthy majority.
> 
> **Lag reference** for `BlockHeadLag` / `FinalizationLag` (and thus selection-policy predicates) is now the **corroborated freshest** head—the **second-highest** positive head when there are at least **3** upstreams—aligned with served-tip semantics. Below that quorum, behavior stays on the raw max. Per-upstream lag is **clamped to ≥ 0**; when the reference moves without the network max changing, stored `evmLatestLagRef` / `evmFinalizedLagRef` trigger a **network-wide lag recompute**.
> 
> **Semantics shift:** lag is distance to the corroborated tip, not the lone leader (e.g. heads `1000/998/995` → lags `0/0/3`). Raw network max metrics are unchanged.
> 
> New and updated tests cover the incident shape, genuine laggards, finalized axis, order statistics, rollback, and real-poll policy integration.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d332dcfeba2ad08687b2b35087e131fa569d78b3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/erpc/networks_selection_policy_realpoll_test.go
+++ b/erpc/networks_selection_policy_realpoll_test.go
@@ -279,6 +279,10 @@ func TestNetworkPolicy_RealPoll_LaggingUpstreamExcluded(t *testing.T) {
 // TestNetworkPolicy_RealPoll_WithinTolerance_NoneExcluded guards the other
 // direction: small, normal lag (a few blocks behind the tip) must NOT trip
 // blockNumberLagAbove(16). Same real-poll path, no exclusion expected.
+//
+// Lag is measured against the CORROBORATED freshest head (the second-highest,
+// 998 here — see health.Tracker networkLagReference), not the single most-ahead
+// node: u1's small lead does not count against its peers.
 func TestNetworkPolicy_RealPoll_WithinTolerance_NoneExcluded(t *testing.T) {
 	util.ResetGock()
 	defer util.ResetGock()
@@ -288,13 +292,13 @@ func TestNetworkPolicy_RealPoll_WithinTolerance_NoneExcluded(t *testing.T) {
 
 	network := setupRealPollLagNetwork(t, ctx, []realPollFixture{
 		{id: "u1", latest: 1000},
-		{id: "u2", latest: 998}, // 2 behind
-		{id: "u3", latest: 995}, // 5 behind — still within the 16-block tolerance
+		{id: "u2", latest: 998}, // at the corroborated tip
+		{id: "u3", latest: 995}, // 3 behind it — well within the 16-block tolerance
 	})
 
 	require.EqualValues(t, 0, blockHeadLagOf(t, network, "u1"))
-	require.EqualValues(t, 2, blockHeadLagOf(t, network, "u2"))
-	require.EqualValues(t, 5, blockHeadLagOf(t, network, "u3"))
+	require.EqualValues(t, 0, blockHeadLagOf(t, network, "u2"))
+	require.EqualValues(t, 3, blockHeadLagOf(t, network, "u3"))
 
 	order, excluded := policy.LatestDecisionOutputForTest(network.policyEngine, network.networkId, "*")
 	assert.NotContains(t, excluded, "u1")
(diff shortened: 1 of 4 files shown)


aramalipoor:
@kasrakhosravi should the source of highest latest block come from "served-tip" subsystem?

kasrakhosravi:
heads up: served-tip would change the normal case too — for [1000, 998, 995] this PR keeps max (lags 0/2/5), but served-tip anchors to the majority 998 (lags 0/0/3). so it's a semantics shift (distance-to-furthest-node → distance-to-majority-tip), and RealPoll_WithinTolerance + the observability split test would need updating.

PR #1002, opened 2026-07-16
fix(server): drain keep-alive connections during waitBeforeShutdown grace window

aramalipoor:
## Problem

On SIGTERM the server sleeps `waitBeforeShutdown`, then calls `http.Server.Shutdown()`. During that grace window **keep-alives stay enabled**, so nothing tells connected clients to move.

Behind load balancers that preserve established TCP flows (e.g. AWS NLB), pooled HTTP clients (gateways, sidecars, aggregators) keep pushing requests over kept-alive connections through the entire grace window — the sleep migrates nothing. `Shutdown()` then stops the listener and closes connections under live traffic: requests racing the close see resets, which surface as 502s at the client on **every deploy**.

We measured this in production: ~150 client-facing 502/504s per regional rolling deploy at ~1k rps through an NLB (fast 0.2–0.6s 502s = resets on pooled connections; 504s = in-flight work caught by the shutdown context deadline), even with `waitBeforeShutdown: 30s`, ECS `stopTimeout: 120` and NLB `deregistration_delay: 120` all correctly aligned.

## Fix

Disable keep-alives at drain start (`SetKeepAlivesEnabled(false)` on both v4/v6 servers) before the `waitBeforeShutdown` sleep:

- every HTTP/1.1 response during the grace window carries `Connection: close`, so a high-rate pooled client migrates its entire pool to healthy instances within seconds;
- HTTP/2 connections tracked by the server get GOAWAY from `Shutdown()` as before;
- by the time `Shutdown()` runs, the instance holds no pooled connections and exits clean.

`waitBeforeShutdown` thus becomes an *active drain* window instead of a passive sleep. No config or behavior change for k8s users — readiness-probe-based draining works as before, this only adds connection migration on top.

## Test

`TestHttpServer_DrainStampsConnectionClose`: steady-state responses keep connections alive; after ctx cancellation (drain window) the server still serves but every response carries `Connection: close`. Verified the test **fails without the fix**.

### DIFF
--- a/erpc/http_server.go
+++ b/erpc/http_server.go
@@ -209,6 +209,19 @@ func NewHttpServer(
 
 	go func() {
 		<-ctx.Done()
+		// Actively drain keep-alive connections during the grace window: stamp
+		// `Connection: close` on every HTTP/1.1 response (and let Shutdown GOAWAY
+		// tracked HTTP/2 conns) so pooled clients migrate to healthy instances
+		// BEFORE Shutdown starts closing connections. Load balancers that preserve
+		// established flows (e.g. AWS NLB) never break these pools on their own —
+		// without this, a client pushing traffic over kept-alive connections rides
+		// them straight into Shutdown and sees resets (502s) on every deploy.
+		if srv.serverV4 != nil {
+			srv.serverV4.SetKeepAlivesEnabled(false)
+		}
+		if srv.serverV6 != nil {
+			srv.serverV6.SetKeepAlivesEnabled(false)
+		}
 		// wait for readiness probe to mark the pod NotReady
 		// ideally (period_seconds * failure_threshold) + safety margin (1s)
 		if srv.serverCfg.WaitBeforeShutdown != nil {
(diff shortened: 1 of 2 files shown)


aramalipoor:
**On the `units` failure** (`TestConsensusPolicy/prefer_non_empty_when_both_groups_above_threshold`, expected `"0x33"` got `[]`): this is a pre-existing flake, not caused by this PR.

Evidence:
- The failing test uses per-test isolated HTTP mock servers (`startConsensusMockServers`, no shared gock), and `make test-fast` runs `SHARD_CONSENSUS` and `SHARD_HTTPSERVER` as **separate processes** — this PR's test can't interact with it.
- Reproduced with CI's exact shard invocation (`-test.run "^(TestConsensusPolicy$|...)" -test.parallel 4`) on **pristine `origin/main` (245e43ff): fails 1 in 3 runs** with the identical assertion diff. Same 1-in-3 rate on this branch.
- In isolation (`-count=10`) it passes 10/10 on both main and this branch — the flake only appears under parallel load, which is why it surfaces intermittently in CI.
- `SHARD_HTTPSERVER` (where this PR's test runs): 0 failures across repeated runs on this branch.

Also pushed a hardening commit (5dc36fc) making the new drain test fully self-contained: its erpc instance's state poller now gets dedicated gock mocks (it polls in the background and gock's registry is process-global within a shard), and the request targets an unknown project so the assertion — which is purely about `Connection: close` on the wire — doesn't depend on upstream mocks at all.

PR #1036, opened 2026-08-03
feat(evm): add skipSyncingCheck upstream config to bypass eth_syncing polling

andreclaro:
## Summary

- Adds `skipSyncingCheck` boolean field to `EvmUpstreamConfig`
- When `true`, skips `eth_syncing` polling for that upstream and hard-sets `syncingState` to `NotSyncing`
- Wires `Copy()` to deep-copy the new pointer field
- `SetDefaults` propagates `skipSyncingCheck` from `upstreamDefaults` (defaults to `false`)
- Diagnostics (`GetDiagnostics`) now reflect operator-configured skip, not just the auto-set internal flag
- Adds unit test asserting `syncingState = NotSyncing` and zero `eth_syncing` calls when flag is set
- Docs: adds `skipSyncingCheck` to the upstream config table and block-tracking reference

## Motivation

Some chains (e.g. Pharos/Antora by Ant Group) always return a syncing object from `eth_syncing` even when fully caught up. Root cause: peer advertisements keep `sync_target` one block ahead of `my_best` in the node's sync logic, so `IsSyncing()` is permanently `true`.

This permanently traps eRPC in `EvmSyncingStateSyncing`. In that state, every emptyish RPC response (`null`, `[]`, `0x`) — valid for calls like `eth_getTransactionByHash` on an unknown tx or `eth_getLogs` with no events — counts as a circuit breaker failure, causing constant CB cycling (~200–400 state changes/hr).

The existing internal `skipSyncingCheck` runtime field auto-sets after 10 consecutive `eth_syncing` **errors** — but nodes like Pharos succeed on the first call (returning a syncing object rather than an error), which sets `syncingSuccessfulOnce = true` and permanently blocks the auto-set path. This config field is the only way to pre-set the behavior from outside for such nodes.

> **Note on duck-typing alternative:** The reviewer suggested extending the object-based syncing response parser instead of adding a config flag. The Pharos bug is in their sync logic — `sync_target` is always `my_best + 1` due to peer gossip timing — so the syncing object always has `currentBlock == highestBlock - 1`. There is no reliable structural signal in the response to distinguish "genuinely syncing" from "falsely syncing" for this chain. A config escape hatch is the correct fix here; Pharos is implementing a proper server-side fix on their end.

## Usage

```yaml
upstreams:
  - endpoint: http://pharos-mainnet-node:18100
    evm:
      skipSyncingCheck: true
```

## Test plan

- [x] `go build ./...` passes
- [x] `TestEvmStatePoller_IgnoreSyncingCheck_ImmediatelyNotSyncing` passes — asserts `syncingState = NotSyncing` and zero `eth_syncing` calls
- [x] Existing state poller tests unaffected


common/config.go
```diff
@@ -1280,6 +1259,10 @@ type EvmUpstreamConfig struct {
 	// the feature.
 	TraceFilterAutoSplittingRangeThreshold int64                    `yaml:"traceFilterAutoSplittingRangeThreshold,omitempty" json:"traceFilterAutoSplittingRangeThreshold"`
 	SkipWhenSyncing                        *bool                    `yaml:"skipWhenSyncing,omitempty" json:"skipWhenSyncing"`
+	// SkipSyncingCheck disables eth_syncing polling for this upstream, treating it as always synced.
+	// Use for nodes that always return a syncing object (e.g. Pharos/Antora) where the response is
+	// misleading and causes circuit breaker false positives.
```
aramalipoor, line 1264: We already have a branch logic that handles object-based syncing responses, is it possible to add support for these chain's specific structure? Ideally in duck-typing approach (as generic as possible)

aramalipoor (approved):

PR #1038, opened 2026-08-03
feat(svm): serve Solana getBlock from the BDS gRPC cache

kasrakhosravi:
> **Reading this diff.** It targets `main` and therefore carries all of #799 — 100 files. Only these 8 are mine:
>
> ```text
> clients/grpc_bds_client.go       +215   dispatch, handler, encoding guard, error mapping
> clients/grpc_bds_resilience.go     +6   svm client on the pooled conn
> clients/grpc_bds_svm_test.go     +387   tests
> common/config.go                  +15   GrpcConnectorConfig.NetworkId
> data/grpc.go                     +104   allowlist, config binding, poller skip
> data/grpc_svm_test.go            +180   tests
> go.mod / go.sum                    +12   manifesto bump
> ```
>
> `git diff feature/solana-phase1` gives exactly that. It originally targeted `feature/solana-phase1`, but GitHub resolves that to a stale copy of the branch on this repo (`0e0a1239`) rather than the PR head, which made it permanently conflicting.

Stacked on #799 — review that first.

Wires prism (the Solana BDS reader) in as a read-through cache for `getBlock`. Five changes in the cache path plus one config field.

## Measured against the live reader

Ran this build against the deployed mainnet reader, port-forwarded, with two projects over identical upstreams — only one has the connector in front. Every number below is from that rig, not a unit test.

**Correctness — prism vs a live upstream, same slot:**

```text
slot 436798900   7 keys, 0 differing -> IDENTICAL
slot 436798901   7 keys, 0 differing -> IDENTICAL
```

**Encoding guard** (hit counter must only advance for `json`):

```text
encoding=json          -> SERVED BY PRISM
encoding=jsonParsed    -> fell through to upstream
encoding=base58        -> fell through to upstream
encoding=base64        -> fell through to upstream
encoding=base64+zstd   -> fell through to upstream
transactionDetails=accounts -> fell through to upstream
```

That guard is the correctness-critical part of this PR, so here is why it exists. Same slot, same node, two encodings:

```text
json        instruction keys: [accounts, data, programIdIndex, stackHeight]
jsonParsed  instruction keys: [parsed, program, programId, stackHeight]
```

Prism returns a structured block, which renders as `json` and nothing else — `jsonParsed` needs per-program decoders (SPL Token, System, Stake, Vote, ATA…) to turn opaque instruction data into named fields. Serving `json` to a `jsonParsed` caller hands them numeric account indexes where they expect base58 pubkeys, with no `parsed`/`program` fields — **a wrong answer that still deserialises**. So anything but `json` (and absent, which means `json`) falls through.

## What changed

**Network binding by config, not probe.** The bootstrap maps server → network by probing `eth_chainId`; Solana has no numeric chain id and answers `Unimplemented`, so no client was ever registered. Probing `GetGenesisHash` instead is not available either — the reader deliberately does not publish one until it can verify its own. New `GrpcConnectorConfig.NetworkId` pins it statically. EVM keeps the probe: it is what arms the per-request chainId assertion that catches an endpoint cross-wired *later*, and a static binding cannot. That gap is noted at the assignment site.

**`getBlock` joins `supportedMethods`** — Solana method names are unprefixed so they cannot collide with `eth_*`; one allowlist serves both.

**Dispatch** maps `getBlock` onto `bds.svm.RPCQueryService/GetBlock`, including the `rewards` inversion the proto asks for (`ExcludeRewards` is false when Agave's `rewards` is true).

**Error mapping.** `OutOfRange` / `Unimplemented` / `InvalidArgument` are benign misses that fall through — some other upstream has it, or a live upstream renders the proper JSON-RPC error. Everything else stays a real error and scores against the connector. A skipped slot also falls through: it is a permanent, cacheable fact, but Agave reports it as `-32007` and a cache connector can only return a *result*, not an error.

**Head poller skips non-EVM networks.** It issues `eth_getBlockByNumber` tag reads that a Solana server answers `Unimplemented` — three wasted round trips per interval, forever. The consequence is documented at the skip: no head state, so **only cache SVM under a `finalized` policy** until there is a slot-aware poller.

## Operator note on finality

`getBlock` is slot-pinned, so `GetFinality` keys off the *requested commitment*. A `commitment: confirmed` request classifies `unfinalized` and skips a `finalized` policy — correct, since confirmed data can still be reorged. In practice that means the cache engages for `commitment: finalized` reads. Verified both directions on the rig.

## Dependency

Importing `manifesto/evm` and `manifesto/svm` in one binary panicked at init on a duplicate `models.proto` descriptor path — both were generated with a bare include root, and nothing had imported both before. Fixed upstream in blockchain-data-standards/manifesto#15.

**Merged and pinned.** `go.mod` points at the published module — no `replace` directive.

## Tests

14 tests / 59 subtests, authored against the contracts above. **19/19 source mutations killed** — including a swapped `pre_balances`/`post_balances`, an inverted rewards flag, and a head poller that skips the wrong architecture.

```text
go test ./clients/... ./data/... ./common/...   ok (no env workarounds)
go vet   clean
gofmt    clean on all changed files
```


kasrakhosravi:
Folded into #799 (`e66c54c6`) — closing this.

Splitting it out was the wrong call. My reason was "reviewable in isolation", but that evaporated the moment I retargeted to `main`: it showed 100 files with a note explaining that only 8 were mine, which is strictly worse than just being in #799. And the staging deployment pins a single image built from this branch, so stage would have been testing a commit #799 did not contain — two PRs for one thing that ships as one artifact.

The eight files are unchanged, squashed into one commit on `feature/solana-phase1` with the temporary manifesto `replace` already removed (blockchain-data-standards/manifesto#15 merged).

PR #1046, opened 2026-08-06
fix(deps): pin postcss to 8.5.18 to fix sourceMappingURL path traversal

andreclaro:
## Summary
- Next.js bundles its own exact `postcss` version internally (`8.4.31` as of `next@15.5.19`), so bumping `next` alone doesn't change it
- Adds a pnpm `overrides` entry pinning `postcss` to `8.5.18` in `docs/package.json`, fixing [GHSA-r28c-9q8g-f849](https://github.com/postcss/postcss/security/advisories/GHSA-r28c-9q8g-f849): path traversal in previous source map auto-loading via `sourceMappingURL`, which could disclose the contents of arbitrary `.map` files reachable via `../` traversal (or an absolute path when `from` is unset)

## Test plan
- [x] `pnpm install` in `docs/` regenerates the lockfile with `postcss@8.5.18` resolved
- [x] `pnpm build` in `docs/` completes successfully

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -27,6 +27,9 @@
 	"pnpm": {
 		"ignoredBuiltDependencies": [
 			"sharp"
-		]
+		],
+		"overrides": {
+			"postcss": "8.5.18"
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


andreclaro:
Superseded by #1048, which consolidates all these dependency fixes into a single PR.

PR #1047, opened 2026-08-06
fix(deps): pin svgo to 3.3.4 to fix removeScripts sanitization bypass

andreclaro:
## Summary
- `svgo` is a transitive dependency in `docs/` via `@svgr/webpack` -> `@svgr/plugin-svgo`, which doesn't declare a compatible newer `svgo` range on its own
- Adds a pnpm `overrides` entry pinning `svgo` to `3.3.4` in `docs/package.json`, fixing [GHSA-2p49-hgcm-8545](https://github.com/svg/svgo/security/advisories/GHSA-2p49-hgcm-8545): the `removeScripts` plugin left namespaced `<svg:script>` elements and case-mismatched `javascript:` URIs intact, which could enable XSS for consumers relying on it for SVG sanitization

## Test plan
- [x] `pnpm install` in `docs/` regenerates the lockfile with `svgo@3.3.4` resolved
- [x] `pnpm build` in `docs/` completes successfully

### DIFF
--- a/docs/package.json
+++ b/docs/package.json
@@ -27,6 +27,9 @@
 	"pnpm": {
 		"ignoredBuiltDependencies": [
 			"sharp"
-		]
+		],
+		"overrides": {
+			"svgo": "3.3.4"
+		}
 	}
 }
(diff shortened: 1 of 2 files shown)


andreclaro:
Superseded by #1048, which consolidates all these dependency fixes into a single PR.

PR #1063, opened 2026-08-11
fix(evm): served tip — bounds don't vote, regression guard, trajectory referee

aramalipoor:
# Served tip: bounds don't vote, regressions are guarded, corroborated fresh minorities can win

Eleven commits, one coherent change to how the served tip is computed and defended: three that build it, then eight that apply the findings of three adversarial reviews and of an independent verification pass over the result. Every finding is mapped to its resolving commit in the comments below.

## Incident this fixes

2026-08-11 12:36–12:41 UTC, network `celo` (evm:42220): the served-tip pick collapsed from 74,550,946 to exactly **21,615,999** — the `blockAvailability.upper.exactBlock` of two pre-Gingerbread "legacy" upstreams. One healthy upstream (`prism-celo`) froze for ~30s and was lag-excluded; the ballot became `[live 74,550,978, live 74,550,974, cap 21,615,999, cap 21,615,999]`; `PickServedTip` takes `heads[len/2]` (3rd of 4) = the cap. The `latest` tag was interpolated to that height, the availability gate then rejected the healthy upstreams (their lower bound is 21,616,000) and admitted only the legacy pair — which correctly served September-2023 state: successful `eth_getBalance` = `0x0` for funded accounts (accepted via `emptyResultAccept`), `eth_call` reverts, empty logs/receipts. Recovery was self-healing when the frozen upstream returned.

## 1 — Availability-bounded upstreams no longer vote a cap as a head observation

A `blockAvailability.upper.exactBlock` bound is a serving-range declaration, not a chain-head observation. `EvmEffectiveLatestBlock()` returns `min(pollerHead, upperBound)`, so a bound-capped upstream voted its config constant into the tip ballot. `evmTipBallot` now drops such inputs **when at least one live-head input exists** in the same candidate set.

**The classification is static, from the CONFIG** (review finding M3). An upstream is a non-voting cap only when its configured `upper.exactBlock` sits below the axis-appropriate head (`LatestBlock` for latest, `FinalizedBlock` for finalized). Every other upper form is *derived* from the upstream's own head — `latestBlockMinus: N` above all — so it tracks the chain at a fixed lag and remains a live observation: an upstream that serves up to `head-5` still knows exactly where the head is. The first version tested the *resolved* bound against the head, which is true for every `latestBlockMinus` upstream, so it disenfranchised them all: on a 4-node fleet with three `latestBlockMinus: 5` nodes, a tip all four could serve became one only the single unbounded node could. Reading the configured constant also removes the second `resolveAvailabilityBounds` call per upstream per evaluation (the largest per-request cost this feature added) and the torn two-load snapshot it implied.

Decided edge cases:
- Mixed pool (≥1 live head): capped upstreams are not voters. They keep serving their range; routing is unchanged.
- All-capped pool: unchanged behavior — a network of only historical/frozen upstreams legitimately serves its cap as `latest`.
- The filter may shrink a ballot to ONE live upstream. That is deliberate and tested: one live witness is thin evidence, but the alternative is advertising a serving range that stopped tracking the chain years ago, and the regression guard bounds how far the tip may move from what this process last saw corroborated.
- **The guaranteed-method floor uses the same rule** (review finding M4). It previously built its ballot from raw effective heads and ran *after* the guard, so with `servedTip.guaranteedMethods` configured the celo pair's 21,615,999 came back as a "floor" and was served end to end. Both ballots now come from one function, with the same per-set mixed-pool rule: capped upstreams drop out of a method's ballot only when some live upstream also supports that method. When only capped upstreams support it, they still set the floor — that is the floor's purpose.

## 2 — Regression guard against corroborated live evidence

Physics backstop, independent of which voters are honest: `latest` cannot fall far below where the chain demonstrably is in one tick. If `pick < reference − maxRegressionBlocks` (new `servedTip.maxRegressionBlocks`, default 1024 = `DefaultToleratedBlockHeadRollback`, `-1` disables), the guard serves the last corroborated pick for a 60s TTL, then **fails open and disarms**. State lives on the existing `servedTipAnchor`, in-process only, no persistence, no shared state.

The reference is split into the two roles it was conflating (review findings m5, M2):

- **Detection** measures against the higher of the *second-highest* live head and the last live-corroborated pick. Against the raw max, one far-future/wrong-chain upstream made every honest pick look like a regression — and because a corroborated value is only recorded on the healthy branch, the guard then never recorded one and sat **permanently disarmed, emitting no metric and no log at all**. The last-corroborated term is needed too: the second-highest live head is itself the poison in the guard's headline shape (one live witness left, two stuck upstreams outvoting it).
- **Ceiling** on a held value is the highest live head, so the guard still cannot advertise a block no live upstream has. A ceiling a rogue can only *raise* is harmless; a ceiling of the second-highest head would collapse the hold back onto the poisoned pick.

**No live head is no longer a reason to stand down** (M2). If every live upstream leaves the eligible set while capped ones remain, the ballot correctly falls back to the caps — and both history-aware layers used to stand down there, on the ballot shape closest to the incident itself. With no live head the guard now measures against its own last corroborated pick, holds it *unclamped* for the TTL, then fails open to the capped pick. Genuinely all-capped networks never armed the guard, so they still pass straight through.

Coverage boundary, documented in the code: the remembered value's freshness comes from **request traffic**, so a network idle longer than the TTL fails open on its first poisoned evaluation after the quiet period, exactly as a cold process does. A memory that survives silence is the frozen-tip absorbing state this design forbids.

## 3 — Trajectory referee: corroborated fresh minorities beat stalled majorities

Every other mechanism is a snapshot; none knows where the chain was ten minutes ago. If a majority of eligible upstreams stalls (inside the lag-exclusion threshold, or during a fleet-wide shared-counter freeze that stalls the lag metric itself), the median follows the stale majority and two genuinely-fresh nodes are outvoted.

`TipTrajectory` samples the median of live heads (never the pick — no feedback loop) at most every 5s into a per-network/axis/lane in-memory ring, fits a robust velocity (median of half-window-lagged slopes; one poisoned sample touches ≤2 of ~90 slopes), and projects expected-now `E`. Tolerance = `max(6 × median-absolute-residual, maxRegressionBlocks)` (≈4σ) — bursty chains widen their own gate.

It participates only when **all** confidence clauses hold: window span ≥ `servedTip.trajectoryWindow` (default 10m; `0` disables; otherwise validated to `[1m, 68m16s]`), ≥30 samples, freshest sample ≤60s old, **no gap larger than 60s anywhere inside the window**, positive finite velocity. Then: cluster live heads (width `max(16, 10s × v)`), candidate groups need **≥2 members**, elect the group closest to `E` within tolerance, serve the group **minimum** (every member can serve it).

**Corroboration is earned over time, not in an instant** (review finding M1 — two confirmed exploits of the first version):

- a **chain halt** makes `E` sweep upward through every fixed offset, so a *static* pair (fork, wrong-chain pair, frozen counter) at +2000 was elected on 74 of 240 halt evaluations;
- a **routine 15s divergence** of 2-of-4 upstreams produced an override that lasted ~4m50s, advertising a block only half the fleet had.

A group may now win only after being the elected group, continuously within tolerance, for **≥30s** (6 samples), *and* having advanced its own head over that stretch by at least **0.5 × v × dwellSeconds**. The dwell is keyed on the group's member SET, so replacing a member restarts it. The velocity condition makes the halt sweep structurally impossible rather than unlikely: a static group's advance is 0 and the threshold is strictly positive whenever the referee is confident at all. **Minimum group size stays 2 at any fleet size** — scaling it with the fleet hands a stalled majority a veto in exactly the topology the referee exists for; what makes two witnesses trustworthy is the dwell and the velocity agreement, not the count.

The referee remains **upward-only**: it can raise the pick toward a corroborated fresh group, never lower it. Any uncertainty → exact plain-median behavior, byte-identical.

## Why the incident value cannot recur

`PickServedTip` is unchanged, so the *ordering* that produced 21,615,999 is still what it always was. Four independent gates now stand between it and a client:

1. a static cap is not on the ballot while any live head is (commit 1);
2. the guaranteed-method floor cannot re-admit one — **without this the claim was false end to end whenever `guaranteedMethods` was configured**;
3. losing every live upstream no longer disables the guard — **without this the claim was false whenever the eligible set collapsed onto the capped pair**, which is precisely how the incident began;
4. and (2) and (3) no longer defeat each other. Each was correct alone and they **composed straight back onto the incident value**: the guard correctly refused the cap-only ballot and held 74,550,974, and then the floor — which runs after the guard — recomputed that same collapsed instant and dragged the served value to 21,615,999. A cap-only method ballot may now set a floor only when the **registered** configuration has no live upstream serving that method; when live upstreams serve it and are merely absent, the floor is a transient lie and the guard's hold (with its bounded fail-open) governs. Pinned by `TestServedTip_GuaranteedMethodFloor_DoesNotUndoTheGuardsHold`, which drives the eligibility loss the way production does — the live upstreams stay registered.

The first version of this PR claimed (1) alone was sufficient; the second closed (2) and (3) separately. All four are needed, and all four are tested.

## Observability

`erpc_network_served_tip_regression_total{outcome=held|failed_open}` and `erpc_network_served_tip_trajectory_total{outcome=override|fallback}` count **state transitions, not evaluations** (review finding m12). The served tip is computed several times per request, so the previous per-evaluation counters scaled with RPS and reported one sustained regression as thousands — which made the help text's "any nonzero rate is alert-worthy" unusable. One sustained hold is now `+1 held`, then `+1 failed_open` if it never recovers. An override the guaranteed-method floor nullifies is not counted or logged at all: it changed nothing. `erpc_network_served_tip_lag_blocks` is clamped at 0 (it read negative whenever the guard held above the corroborated freshest view).

## Known boundary: the guard tolerates one Byzantine head, not two

The regression guard's detection reference is the second-highest live head, which neutralises a single far-future/wrong-chain upstream. **Two** upstreams reporting the same far-future block are indistinguishable, in one evaluation, from two upstreams that are simply ahead — so they can still leave the guard unarmed, and silently, since a corroborated value is only recorded on the healthy branch. Nothing available in a snapshot separates those cases. What bounds the damage is stated in the code where the reference is computed: the majority pick still needs a majority to move, the guard's memory refuses to commit an uncorroborated leap, and the referee's dwell refuses to elect a group that has not tracked the chain for half a minute.

## Two findings deliberately left as-is

- **A 2-upstream network in deep divergence holds for up to 60s.** With two upstreams far apart, the guard holds its last corroborated pick until the TTL rather than following the lower head immediately. That is the better answer: the alternative is `main`'s behaviour of serving, say, a devnet's post-reset block 1 as `latest` the moment one node is re-genesised. The hold is bounded, counted and fails open.
- **The served tip can drift between the several calls of one request.** `EvmHighestLatestBlockNumber` is called independently by block-tag interpolation, the availability gate and the empty-result guard, and nothing pins one value for the request's lifetime. This predates the branch and is unchanged by it; noted here so it is not read as introduced.

## Verification

- `go build ./...`, `gofmt` clean on every touched file (the pre-existing offenders on `main` are untouched).
- `go test ./architecture/... ./erpc/... ./upstream/... ./common/... -count=1 -timeout 40m` green (`erpc` alone runs ~14 minutes, so the default 10-minute timeout panics — pass `-timeout 30m` or more).
- Served-tip subset re-run with `-race`.
- Mutation-checked where a constant could silently un-pin a test: with `tipDwellDuration = 0` two of the three dwell tests fail; with the guard memory's leap rule disabled the fantasy-recovery test fails.
- Reviewer repro material ported into real regression tests: halt-sweep + static pair, gap amnesia (both the "one evaluation later" and "halt inside the gap" shapes), transient 2-of-4 divergence, permanently split fleet, `latestBlockMinus` fleet servability, the guaranteed-method floor re-admission, the all-live-excluded hold, single-rogue disarm, the guard's clamp branch, transition-counted metrics, and the finalized axis end to end for all three mechanisms.
- `typescript/config/src/generated.ts` updated for the changed config docs (relevant fields only; the repo regenerates types at release time).

### DIFF
--- a/architecture/evm/served_tip.go
+++ b/architecture/evm/served_tip.go
@@ -1,6 +1,12 @@
 package evm
 
-import "sort"
+import (
+	"math"
+	"sort"
+	"sync"
+	"sync/atomic"
+	"time"
+)
 
 // ServedTipInput is a single observation: an upstream's last-known tip block.
 // Callers are responsible for excluding syncing or cordoned upstreams BEFORE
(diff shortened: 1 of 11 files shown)


aramalipoor:
## Review findings → resolution

Four commits on top of the original three, applying the consolidated findings of three adversarial reviews. Every reviewer repro is ported as a real regression test in-tree (nothing is left in a scratch module).

| # | Finding | Resolution | Commit |
|---|---------|-----------|--------|
| **M1** | Referee election was an INSTANT test. Two confirmed exploits: during a chain halt the expected head sweeps upward through every fixed offset, so a static pair at +2000 won 74 of 240 halt evaluations; a routine 15s divergence of 2-of-4 upstreams produced a ~4m50s override. | Corroboration must be earned: the elected group must hold its place, continuously within tolerance, for **≥30s** (6 samples), **and** have advanced its own head by ≥ `0.5 × v × dwellSeconds`. Dwell is keyed on the group's member SET (order-independent fold of member ids), reset on membership change, on a miss, and on loss of confidence. A static group's advance is 0 against a strictly-positive threshold, so the halt sweep is structurally impossible, not merely unlikely. **Min group size stays 2 at any fleet size** — documented in `tipMinGroupSize`: two fresh witnesses must be able to outvote a stalled majority, and dwell+velocity is what makes 2 trustworthy. Tests: halt+static pair at +2000/+6000/+20000 → 0 overrides (log shows 88 evaluations where the sweep *did* put it on trajectory); 2-of-4 pause 15s → no override; genuine stall → still overrides, and only after the dwell; permanently split fleet → silent; override self-expires once the elected group stops advancing. | `068c72c5` |
| **M2** | All-live-excluded door: with `maxLiveHead == 0` the ballot fell back to caps AND both the referee and the guard stood down → the incident value served again. | With no live head the guard measures against its own last corroborated pick and holds it **unclamped** for the TTL, then fails open to the capped pick. Boot with nothing remembered, and genuinely all-capped networks, still pass straight through. Test: celo-shaped fixture (3 live + 2 capped at 21,615,999), all three live removed from eligibility → holds ~74.5M with `outcome=held`, then serves 21,615,999 with `outcome=failed_open`. | `12f6330a` |
| **M3** | Capped-voter classification was dynamic (`resolved upper < head`), which disenfranchised every `upper.latestBlockMinus` upstream (a 4/4-servable tip became 1/4-servable), took a torn two-load snapshot, and doubled `resolveAvailabilityBounds` cost. | Classification is **static, from the config**: a non-voting cap requires `blockAvailability.upper.exactBlock != nil` and `*ExactBlock < head` for the axis (`LatestBlock` / `FinalizedBlock`, per-axis correctness preserved). `latestBlockMinus` of any N is head-derived and always votes. No second bounds resolution. Tests: `latestBlockMinus: 5` fleet keeps a majority-servable tip (`upperLatestMinus` added to `servedTipFixture`); the celo `exactBlock` pair is still filtered; a ballot shrunk to one live upstream is exercised and documented. | `9015c19b` |
| **M4** | `guaranteedMethodFloor` re-admitted the cap: its ballot used raw `EvmEffectiveLatestBlock` and ran after the guard, so with `guaranteedMethods` set the network served 21,615,999 end to end. | The floor builds each method's ballot with the same `evmTipBallot` rule, including the per-set mixed-pool rule: caps drop out of a method's ballot only when a live upstream also supports that method; when only caps support it they still set the floor. Tests: the reviewer's celo + `guaranteedMethods: ["eth_getLogs"]` repro, plus the inverse (caps alone on a method still floor the tip). | `9015c19b` |
| **m5** | One far-future rogue froze `lastGood = 0` forever, silently disarming the guard. | Detection now measures against the higher of the **second-highest live head** and the last live-corroborated pick; the held value's ceiling stays the **highest** live head (a ceiling a rogue can only raise is harmless, while a second-highest ceiling would collapse the hold back onto the poisoned pick). Test: 3 honest + 1 rogue at 100M → the guard arms, `lastGood` tracks honest picks, and it still holds when the ballot collapses. | `12f6330a` |
| **m6** | Guard TTL used `UnixMilli` wall time. | `lastGoodAt` is an `atomic.Pointer[time.Time]` (monotonic), exercised through the existing `servedTipClock` seam. | `12f6330a` |
| **m7** | Gap amnesia: the gap check looked only at the newest sample's age, so it blocked exactly ONE evaluation; 1ms later the fit extrapolated across the hole, and a halt inside an unobserved gap lifted `expected` to exactly where a rogue pair would wait. | `refit` computes the largest inter-sample interval over the window (`tipFit.maxGapMs`) and `confident()` rejects a fit that straddles one. Tests: both reviewer repros — the evaluation after the gap is inert too, and the halt-in-gap pair (sitting exactly on the inflated expectation) does not win; the fit is refused outright. | `068c72c5` |
| **m8** | Double `resolveAvailabilityBounds` + torn snapshot. | Absorbed by M3: `evmTipObservation` reads the configured constant and one head load. There is no second bounds resolution left on the path. | `9015c19b` |
| **m9** | Per-evaluation unconditional `Swap`/`Store` on the anchor. | Load-first everywhere: `observe`, the guard's state transition, the referee's state transition, and `lastGoodValue`. `lastGoodAt` keeps its freshness semantics (still written on every corroborated pick). | `12f6330a` |
| **m10** | Trajectory recorded/refit on ballots that can never produce a decision — 155 single-upstream chains, and 16 partitions × 7KB rings. | `Observe` returns before recording when the ballot holds fewer than `tipMinGroupSize` valid inputs: no ring, no fit, two comparisons. Test: 200 single-input evaluations → 0 samples, nil fit. | `068c72c5` |
| **m11** | `trajectoryWindow` unvalidated. | Accepted values are exactly 0 (disabled) or `[1m, 68m16s]` (= `tipSampleInterval × tipMaxBufferCapacity × 0.8`; both constants moved to `common` so the bound cannot drift from the mechanism). The error explicitly warns that a bare YAML number is parsed as **MILLISECONDS**. Tests: range table incl. exact bounds and 5ms; YAML round-trip (`"10m"` works, bare `5` never becomes 5 minutes). | `6227bc21` |
| **m12** | Counters incremented per evaluation, and `EvmHighestLatestBlockNumber` runs several times per request, so they scaled with RPS. | Both counters fire on state ENTRY only (enter-hold, enter-fail-open, enter-override, enter-fallback); the referee's state is now a tri-state on the anchor. The override metric/log is emitted **after the guaranteed-method floor**, so an override the floor nullifies is not reported at all. Help text updated ("counts state ENTRIES, not evaluations; any nonzero rate is alert-worthy"). Tests: 20 evaluations of one regression → +1 `held`, +1 `failed_open`; nullified override → 0. | `6227bc21` |
| **m13** | `..._lag_blocks` read negative while the guard held above `Freshest`. | Clamped at ≥0, with the reason in a comment (the two counters are what report those states). Test asserts the gauge while a hold is in progress. | `12f6330a` |
| **m14** | No off switch for the guard. | `maxRegressionBlocks: -1` disables it (validation accepts `-1`, rejects anything below; docstring + `generated.ts` updated), symmetric with `trajectoryWindow: 0`. For the referee it only removes the floor under the tolerance. Test: `-1` → the raw pick is served and nothing is recorded. | `12f6330a` |
| **m15** | Guard coverage boundary undocumented. | Documented on `guardServedTipRegression` and in the PR body: the remembered value's freshness comes from request traffic, so a network idle > TTL fails open on its first poisoned evaluation after the quiet period — deliberate, because a memory that survives silence is the absorbing state this design forbids. | `12f6330a` |
| **m16** | The ballot was sorted twice per evaluation. | `PickServedTip` exposes its own descending working slice (`ServedTipPick.Sorted`) and the referee clusters that. It also carries the upstream ids M1's dwell identity needs, so the second sort is gone rather than moved. | `068c72c5` |
| **m17** | Coverage gaps. | Added: the finalized axis end to end for all three mechanisms (cap filter, referee, guard); the guard's clamp branch reached the way production reaches it (the upstream backing the remembered pick leaves and the best remaining live head is below it); repeated `EvmHighestLatestBlockNumber` calls → transition-counted metrics. | `6227bc21` |

### Deliberately unchanged

- **The 2-upstream deep-divergence 60s hold is intentional.** Better than `main` serving a devnet's post-reset block 1 as `latest`; the hold is bounded, counted, and fails open. One sentence added to the PR body.
- **Value drift between the several `EvmHighestLatestBlockNumber` calls of one request predates this branch** and is unchanged by it. Noted in the PR body so it is not read as introduced here.

### Spec deviations

- M1 specifies the dwell's group identity as "hash of sorted member upstream ids". It is implemented as an **order-independent** fold (sum of per-id FNV-1a hashes) — the same set identity, without sorting or allocating on the request path.
- m5 specifies that the `min(lastGood, reference)` clamp use the same second-highest reference. Detection uses it; the **clamp keeps the highest live head**, because a second-highest ceiling collapses the hold back onto the poisoned pick in the guard's own headline shape (one live witness left, two stuck upstreams outvoting it) — it would have broken `TestServedTip_RegressionGuard_HoldsLastCorroboratedPick`, i.e. the guard's core purpose. Both roles are named and documented on `servedTipReference`.

aramalipoor:
## Verification-pass findings → resolution

An independent verification pass over `b9b7571e` returned NOT MERGE-READY: one blocker where two of the earlier fixes composed back onto the incident value, one test-weakening of the same severity, two moderates and three minors. All are applied.

| # | Finding | Resolution | Commit |
|---|---------|-----------|--------|
| **F1** | **BLOCKER — M2 × M4 compose back onto the incident value.** With `guaranteedMethods: ["eth_getLogs"]` and every live upstream out of the eligible set, the guard correctly held 74,550,974 and then `guaranteedMethodFloor` — which runs *after* the guard — recomputed the same collapsed instant, hit the all-capped fallback, and dragged the served value to 21,615,999. Reproduced at that head. | A cap-only method ballot may set a floor only when the **registered** configuration has no live (non-`exactBlock`-capped) upstream serving that method. Eligibility churn cannot change the registered set, so "live upstreams serve this method, they are merely absent" is now distinguishable from "only capped upstreams ever serve it": the former skips the floor and lets the guard's hold (and bounded fail-open) govern, the latter still floors to the cap. An upstream with no head yet counts as live — "cannot tell" falls on the side of skipping. The check runs only when a method ballot came back cap-only. Tests: `…_DoesNotUndoTheGuardsHold` (holds 74,550,974 through the hold, fails open to the cap past the TTL) and the permanent case re-asserted after the live upstreams leave. | `1fdb6efe` |
| **F-A** | **Dwell pinned by nothing** — zeroing `tipDwellDuration` left the whole suite green, because the tests asserted in terms of the constant they were meant to pin. | Rewritten against **absolute literals**: (1) a pair on the trajectory, corroborating, advancing at chain velocity and separated from the majority *from the first evaluation* must not have overridden at 0/5/15/25s and must have by 45s; (2) the transient two-of-four split is now separated by more than one cluster width from the first sample, so only elapsed time can prevent the override; (3) the membership test asserts the dwell's **reset** branch (existing stretch, changed group id, start moved forward), not the nil branch. **Acceptance: with `tipDwellDuration = 0`, (1) and (2) fail.** | `0e21a85e` |
| **F2** | **Group identity collided structurally.** The commutative fold (sum of per-id FNV-1a hashes) is suffix-swap invariant for prefix-sharing ids: `{prism-celo-1, nirvana-celo-2}` = `{prism-celo-2, nirvana-celo-1}`. A 4-vendor × 3-replica fleet contains **19** colliding pairs, and a collision lets a different set inherit a dwell it never earned. | Ids are sorted and hashed as one separated stream, so the identity is a property of the set with the hash's avalanche intact. The sort runs on a stack-array copy (escape analysis: no allocation) and the request path stays lock-free. Test: the demonstrated colliding shape has distinct identities, order-independence still holds, and the suffix-swapped pair does **not** inherit the earned dwell. | `0e21a85e` |
| **F3** | **`lastGood` could commit a fantasy.** A momentarily majority-far-future eligible set makes the pick itself the fantasy; the guard remembered it and then held it against the honest fleet for the full 60s TTL — worse than `main`, which recovers on the next evaluation. | `rememberServedTipPick` gates the store on: a live ballot; `pick <= ref.Corroborated` (two live witnesses at-or-above, single-live fallback unchanged); and **no uncorroborated leap** — a still-fresh memory may not be raised by more than the regression tolerance in one step. The leap rule constrains only what is *remembered*, never what is *served*, so it cannot hold or freeze a tip; a legitimate burst leaves the memory behind for at most one TTL and the detection reference ignores it (it takes the higher of memory and live corroborated head). Test: fantasy window → memory stays honest → recovery is immediate. **Mutation-checked**: disabling the leap rule fails the test with `100000000` vs `10000`. | `13c73090` |
| **F5** | `lastGoodAt.Store` allocated and stored on every healthy evaluation. | Refreshed at most once per second (load-compare first). A 60s TTL is exact to within its own granularity. | `13c73090` |
| **F6** | The ring was sized once at first use while `Window` is re-read per evaluation, so **raising `trajectoryWindow` on a live config reload stood the referee down forever**. | A window change rebuilds the ring and clears the fit and the dwell; the tracker re-warms from live observations — the same inert-then-confident path every process takes after a deploy. Documented on the new `TipTrajectory.window` field. Test: raised window → track restarts, no stale fit or dwell survives, and re-warming reaches the new span. | `13c73090` |
| **F4** | Two coordinated far-future upstreams can still silently disarm the guard (the detection reference is 1-Byzantine). | Documented only, as specified: a comment where the reference is computed and a section in the PR body naming the boundary and what bounds the damage (majority pick needs a majority, memory refuses leaps, dwell refuses untracked groups). | `13c73090` |

### Notes

- **PR body**: the verification command now carries `-timeout 40m` (the `erpc` package alone runs ~14 minutes, so the default 10-minute timeout panics rather than fails), the "cannot recur" section states the F1 composition explicitly and cites its test, and the 1-Byzantine boundary has its own section.
- **Spec deviations in this round** (both defended in-code): F2 sorts into a stack array rather than a scratch buffer on the tracker — a shared buffer would have to be taken under the tracker's mutex, serialising every evaluation to save an allocation that escape analysis says does not happen. F6 keeps the window-derived ring size instead of always allocating the validation maximum: the fit is computed over the whole buffer, so a 1024-sample ring would silently turn every network's fit into an 85-minute one and change the meaning of `trajectoryWindow`; the rebuild-on-change is what actually fixes the reported bug.
- **F1's test drives eligibility loss with the syncing state**, not the order-pinning helper: `PinUpstreamOrderForTest` replaces the registry outright, so it cannot express "absent from the ballot but still configured" — which is what every production cause (lag exclusion, cordon, restart) actually is.
- **CI on `b9b7571e`**: the `units` job failed on `TestConsensusPolicy…prefer_non_empty`, in code this branch does not touch; it passed locally 3/3. This push re-triggers CI.

aramalipoor:
Round 3 (verification round 2 findings) — commit 5398d4f9:

| Finding | Resolution |
|---|---|
| R2-1 (blocking) — every fail-open disarms the guard's memory, and an unarmed guard committed whatever pick appeared; one majority-rogue evaluation right after a fail-open seeded the memory with a fantasy value which was then held for a full TTL after the honest fleet returned (main recovers immediately in that shape) | Arming now requires continuity: a value may enter the memory only when it is within the regression tolerance above the last value this process already trusted — the armed memory while armed, otherwise the previously served value. A true first evaluation (nothing served yet) arms unconditionally, which equals main's exposure since main has no guard at all. Serving is never constrained by any arming rule. New tests: post-fail-open rogue window (recovery on the next evaluation, not TTL expiry) and a cold-boot variant. |
| Hold symmetry (verifier-implied) | Pinned: a hold may only ever raise the served value toward the corroborated past, never lower it below the current raw pick (`AHoldNeverLowersTheTip`). |
| R2-2 (minor) — the throttle comment claimed an allocation win that did not happen (the timestamp escaped to the heap; 1.00 allocs/op) | Clock read hoisted inside the throttle branch; an allocation regression test now pins 0 allocs/op on the healthy steady path. |

Mutation evidence (recorded in the commit message): removing the continuity gate fails `AfterFailOpenDisarm`; removing the hold-raise guard fails `AHoldNeverLowersTheTip` (serves 100 where the pick was 5,000); restoring the unthrottled clock fails the allocation test at 1 alloc/op.

Independent re-verification of this head is running; its verdict lands here next.

PR #1064, opened 2026-08-11
feat(evm): trajectory referee for the served-tip pick — corroborated fresh minorities beat stalled majorities

aramalipoor:
> **Stacked on #1063** (`fix/served-tip-live-heads`). This PR targets that
> branch, so its diff is only the trajectory referee. Merge #1063 first; this
> retargets to `main` on its own afterwards.

## The gap #1063 does not close

#1063 fixes two ways a served-tip ballot gets poisoned by values that are not
head observations. Both fixes — and the majority order statistic itself — are
**snapshots of one instant**. That is exactly what a majority *stall* defeats:

* N upstreams freeze while staying inside the lag-exclusion threshold, or a
  fleet-wide shared-counter freeze stalls the lag metric itself;
* the strict majority is now the **stale** group;
* the 2 upstreams that genuinely hold the chain's head are outvoted;
* every instantaneous rule agrees with the stale majority, because "most
  upstreams say block H" and "the chain stopped at H" are the same observation.

The operator requirement, verbatim: *in an N vs 2 split where the 2 sit at a
higher block, still use the 2 — but governed by a long-term in-memory
latest-block track recorded over at least 10-15 minutes; pick the group closest
to that long-term trajectory; be robust and strict enough that the tracking
never creates more issues than it solves.*

The last clause is the hard part, and it has a name: commit `7dee07c0` removed
the previous stateful tip pipeline (velocity gate + persisted monotonic clamp)
because it reached an **absorbing state** — every live tip rejected, the served
tip frozen fleet-wide across 25+ chain/region pairs. Anything stateful here has
to be structurally incapable of that, not merely tuned away from it.

## Mechanism

`evm.TipTrajectory` — per network, per axis, per pod, in memory only.

**Record** (throttled to one sample per 5s, on the tip evaluation itself): a
monotonic-clock timestamp and the plain **median of the live heads**. Never the
served value, never the referee's own output, never a bound-capped observation.

**Refit** (on insertion, never per request):

| quantity | estimator |
|---|---|
| velocity | median of **half-window-lagged slopes** — sample `i` against sample `i+n/2`, median of the ~n/2 of them |
| intercept | median residual (the standard robust intercept) |
| spread | median absolute residual around that line |

Two properties matter and full Theil-Sen buys neither: every slope spans half
the window, so per-sample noise is divided by a long baseline; and one poisoned
sample enters at most two of the ~90 slopes, so the median never moves.
Theil-Sen's n²/2 pairs (≈16k at n=180) would cost a 128 KB sort per fit for the
same answer. `TestTipTrajectory_PoisonedSampleDoesNotMoveTheFit` pins it.

**Decide** (O(1) read per request): cluster the live heads by single linkage at
`max(16 blocks, 10s × velocity)`, elect the group of **≥2** whose max sits
closest to the extrapolated expected head, and require that distance to be
within `max(6 × spread, maxRegressionBlocks)`. The pick is the group's
**minimum** — a block every member can serve.

Derived constants, all justified in comments next to them:

| constant | value | why |
|---|---|---|
| sample throttle | 5s | 120 samples in the default window; the O(n log n) refit runs at most 0.2 Hz per network+axis |
| cluster width | `max(16, 10s × v)` | 10s of chain progress covers poller skew and propagation between healthy providers; the floor keeps sub-second-block chains from splitting a healthy fleet into singletons |
| tolerance | `max(6 × spread, maxRegressionBlocks)` | a median absolute residual is ≈0.67σ, so k=6 is ≈4σ — a chain that pauses and bursts widens its own gate, a metronomic chain keeps the configured floor |
| min group size | 2 | one witness is not corroboration, however perfectly it matches |
| min samples / max gap | 30 / 60s | below either, "trajectory" is noise or guesswork |

## Confidence gate — the feature is inert unless every clause holds

`span ≥ trajectoryWindow` (new optional config field: unset → 10m, explicit
`0` → disabled) **and** `≥30 samples` **and** the freshest sample is `≤60s` old
*as of the start of the evaluation* (an evaluation may not paper over the gap it
just noticed) **and** a finite `velocity > 0`.

Otherwise the served tip is **byte-identical** to this branch today — including
on every pod for the first ten minutes after every deploy.

## Why it cannot recreate the `7dee07c0` frozen tip

* **Advisory, never generative.** It only ranks the groups the live upstreams
  offer in this very evaluation. It cannot invent, remember or extrapolate a
  servable block; the trajectory is used to *choose*, never to *produce*.
* **Upward-only.** It may raise the pick to a corroborated fresher group; it
  can never lower or hold one. *A referee that cannot hold a value back cannot
  freeze a tip.* Erring low is already the majority picker's job, and the
  regression guard's — so this is also the weakest rule that satisfies the
  requirement. (Deviation from the spec I was given, which allowed a downward
  override too; it bought nothing the majority pick doesn't already do, and it
  was the only direction from which a frozen tip was reachable.)
* **Self-limiting evidence.** Because samples are the plain median, a majority
  stall keeps feeding the tracker the stalled median: the fitted velocity
  decays and after roughly half a window the referee stands *itself* down. The
  intervention has a built-in expiry no configuration can extend.
* **Self-calibrating.** The fit tracks the median's own history, so a fleet
  permanently split around its median (flashblocks-style groups) elects the
  median's group — no permanently-offset group is ever closer to the
  trajectory. Only a group that leaves its own long-term track loses.
* **No shared state.** Per-pod, in-memory, rebuilt from live observations after
  a restart. Deliberate: a shared head counter is how the 2026-08 celo trigger
  propagated fleet-wide in the first place.
* The existing `networks_served_tip_invariants_test.go` — the black-box
  prod-incident suite that must never be adapted to new internals — passes
  unchanged.

## Order and observability

Inside `servedTip`: bound-capped input filter (#1063 fix 1) → **trajectory or
majority pick** → regression guard (#1063 fix 2) → guaranteed-method floor. The
guard judges whatever the referee produced, unchanged; because the referee's
pick is the minimum of a group of ≥2, it never exceeds
`ServedTipPick.Freshest`, so the lag gauge cannot read negative.

`erpc_network_served_tip_trajectory_total{outcome=override|fallback}` counts
the only two evaluations worth counting — the referee served an on-trajectory
group instead of a stalled majority, or it elected a fresher group and refused
it as not close enough. Both are silent in steady state by construction (a
fleet in one cluster elects that cluster; a fleet split around its median
elects the median's group). Overrides also log once when they start and once
when they stop, never per request.

## Default-on vs default-off — your call

I made it **default-on with a strict gate** rather than opt-in, on this
reasoning: the gate makes it a no-op in every state except the one it exists
for, it can only ever move the tip *up* to a block two live upstreams already
serve, an opt-in flag would mean the networks that need it are exactly the ones
nobody enabled it on, and `trajectoryWindow: 0` is a complete, tested off
switch. If you would rather it be opt-in, say so and I will invert it — the
change is one line in `servedTipTrajectoryWindow()` plus the config doc, and
none of the tests move.

## Test evidence

```
$ go test ./architecture/... ./common/... -count=1
ok  	github.com/erpc/erpc/architecture/evm	3.596s
ok  	github.com/erpc/erpc/architecture/evm/integrity	0.547s
ok  	github.com/erpc/erpc/architecture/svm	2.990s
ok  	github.com/erpc/erpc/common	3.587s
ok  	github.com/erpc/erpc/common/legacy	2.085s

$ go test ./upstream/... -count=1
ok  	github.com/erpc/erpc/upstream	7.647s

$ go test ./erpc/... -count=1 -timeout 25m
ok  	github.com/erpc/erpc/erpc	824.339s

$ go test ./erpc/ -run TestServedTip -count=1 -race -v   # 35/35 pass, incl. the
                                                        # untouched prod-incident
                                                        # invariants
--- PASS: TestServedTip_TrajectoryReferee_StalledMajorityLosesToFreshMinority (1.14s)
--- PASS: TestServedTip_TrajectoryReferee_RunawayMinorityLoses (1.14s)
--- PASS: TestServedTip_TrajectoryReferee_ColdProcessIsInert (1.13s)
--- PASS: TestServedTip_TrajectoryReferee_StaleTrackFailsOpen (2.27s)
--- PASS: TestServedTip_TrajectoryReferee_PauseThenBurstDoesNotFalsePositive (1.12s)
--- PASS: TestServedTip_TrajectoryReferee_SingletonIsNotCorroboration (1.13s)
--- PASS: TestServedTip_TrajectoryReferee_ComposesWithRegressionGuard (1.14s)
--- PASS: TestServedTip_TrajectoryReferee_DisabledByZeroWindow (1.13s)
ok  	github.com/erpc/erpc/erpc	45.107s

$ go test ./architecture/evm/ -run 'TestTipTrajectory|TestTipClusterWidth' -count=1 -race -v
--- PASS: TestTipTrajectory_FitRecoversVelocityAndExpectedHead (0.00s)
--- PASS: TestTipTrajectory_PoisonedSampleDoesNotMoveTheFit (0.00s)
--- PASS: TestTipTrajectory_ConfidenceGate (0.00s)
    --- PASS: .../ColdBufferIsInert   .../ShortWindowIsInert   .../StaleTrackIsInert
    --- PASS: .../HaltedChainIsInert  .../DisabledRecordsNothing
--- PASS: TestTipTrajectory_GroupElection (0.01s)
    --- PASS: .../StalledMajorityLosesToCorroboratedFreshPair
    --- PASS: .../SingletonIsNeverCorroboration  .../NearMissIsDeclinedNotOverridden
    --- PASS: .../RunawayPairIsOutsideTolerance .../UpwardOnly
    --- PASS: .../HealthyFleetIsOneGroupAndNeverIntervenes
--- PASS: TestTipTrajectory_BurstyChainWidensItsOwnTolerance (0.00s)
--- PASS: TestTipTrajectory_SampleThrottleAndBufferBound (0.07s)
--- PASS: TestTipTrajectory_ConcurrentObserve (0.00s)
--- PASS: TestTipClusterWidth (0.00s)
ok  	github.com/erpc/erpc/architecture/evm	2.030s
```

New coverage:

* `architecture/evm/served_tip_test.go` — the fit recovers a known velocity;
  one poisoned sample does not move it; every clause of the confidence gate
  (cold buffer, short span, stale track, halted chain, disabled); the election
  (stalled majority loses, singleton is not corroboration, near-miss is
  declined not overridden, runaway is outside tolerance, upward-only, healthy
  fleet never intervenes); a bursty chain widens its own tolerance while a
  metronomic one keeps the floor; the throttle and the ring bound.
* `erpc/networks_served_tip_test.go` — the same behaviour end-to-end through
  real `EvmHighestLatestBlockNumber` evaluations on the pinned served-tip
  clock: the headline 3-stalled-vs-2-fresh case (with the plain majority pick
  computed alongside to prove an override happened), the runaway minority, a
  cold process, the stale-track pair (identical evidence, one sample past the
  gap), pause-then-burst, the singleton, composition with the regression guard,
  and the `trajectoryWindow: 0` off switch.

### DIFF
--- a/architecture/evm/served_tip.go
+++ b/architecture/evm/served_tip.go
@@ -1,6 +1,12 @@
 package evm
 
-import "sort"
+import (
+	"math"
+	"sort"
+	"sync"
+	"sync/atomic"
+	"time"
+)
 
 // ServedTipInput is a single observation: an upstream's last-known tip block.
 // Callers are responsible for excluding syncing or cordoned upstreams BEFORE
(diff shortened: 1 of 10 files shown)


aramalipoor:
Note: this shows as "merged" only because its base branch fix/served-tip-live-heads was fast-forwarded to include these commits. The reviewable PR against main is #1063, which now carries all three commits.

PR #1080, opened 2026-08-13
feat(evm): network-level blockAvailability lower bound with JSON-RPC -32001 response

andreclaro:
## Summary

- Adds `blockAvailability` to `EvmNetworkConfig` so operators can define a network-level lower (and optionally upper) block availability bound — independent of per-upstream bounds.
- When a request's block number falls below the lower bound, eRPC short-circuits **before upstream selection** and returns a standards-compliant JSON-RPC error (`-32001`, matching what pruned EVM nodes return) instead of a generic HTTP 503/412.
- Evaluated **after** the cache layer, so cached responses are still served.
- Fail-open on unresolvable block numbers (by-hash, tag-only params, unknown head).

## Config example

```yaml
networks:
  - id: evm:130  # unichain-mainnet
    evm:
      chainId: 130
      blockAvailability:
        lower:
          latestBlockMinus: 604800  # ~7 days; matches RETH_RETAIN_BLOCKS
```

## Motivation

Internal nodes on unichain-mainnet prune history older than ~7 days (`RETH_RETAIN_BLOCKS: 604800`). With `lowParticipantsBehavior: returnError`, pruned-block requests degenerate into `ErrConsensusLowParticipants` (HTTP 412) — not a valid JSON-RPC response. Clients that expect a `{"error": {...}}` envelope mishandle it.

This gate short-circuits those requests before they reach consensus and returns a proper `-32001` error that EVM clients already know how to handle.

## Changes

- `common/config.go` — `BlockAvailability *EvmBlockAvailabilityConfig` on `EvmNetworkConfig`
- `erpc/networks.go` — `tryShortCircuitNetworkBlockUnavailable` + call site after cache, before upstream selection
- `erpc/networks_availability_test.go` — 4 tests: below `ExactBlock`, within `ExactBlock`, below `LatestBlockMinus`, by-hash fail-open

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./erpc/... -run TestNetworkBlockAvail` — all 4 pass

## Model and tooling

- AI model: Claude Sonnet 4.6 via Claude Code
- Human author: Andre Claro


andreclaro:
Superseded by #1081.

The review rounds here kept surfacing findings in the same place — the new network-level gate had to re-derive block references, resolve tags, special-case range methods, and recompute bounds, each of which duplicated logic that already exists per-upstream. Rather than keep patching that surface, #1081 drops the gate and closes the two actual gaps instead:

1. `projects[].upstreamDefaults` is already a full `UpstreamConfig`, but `EvmUpstreamConfig.SetDefaults` inherited `MaxAvailableRecentBlocks` and never `BlockAvailability` — so project-level config needed no new field, just the missing inheritance.
2. `ErrUpstreamBlockUnavailable` had no entry in `TranslateToJsonRpcException`, so it surfaced as the generic `-32603`. Mapping it to `-32014` fixes the response for per-upstream bounds too, not only the project-level case.

Net: 60 lines added instead of 331, no new config field, and enforcement stays on the existing per-upstream path — which already handles the range-method and block-tag cases that findings 1 and 2 on this PR were about.

PR #1083, opened 2026-08-18
fix(evm): correct upper-bound fail-open, skipped metric, and consensus error for block availability

andreclaro:
## Problem

Four instrumentation and gating bugs surfaced after shipping project-level \`blockAvailability\` (#1081):

### 1. Upper bound fails open when \`latestBlock\` is unknown

\`resolveAvailabilityBounds\` returned \`math.MaxInt64\` for the upper bound whenever \`latestBlock == 0\` (state poller not yet run, or sustaining transport failures). This made \`checkUpstreamBlockAvailability\` treat both bounds as unbounded and skip gating entirely. On a node with ongoing transport failures the result was \`ErrEndpointTransportFailure\`, which broke the unanimity requirement in \`isBlockUnavailableVerdict\` — leaving a block-unavailable pool producing \`consensus_on_error\` instead of \`-32014\`.

### 2. Pre-check block-unavailable skips not counted toward \`-32014\` verdict

\`isBlockUnavailableVerdict\` treated all \`ErrUpstreamRequestSkipped\` causes as neutral, so a pool where every upstream was pre-check skipped due to block-unavailable (lower or upper bound exceeded) returned \`agreed=0\` and fell through to \`-32603\` instead of \`-32014\`.

\`handleBlockSkip\` wraps non-retryable bound violations in \`ErrUpstreamRequestSkipped\` so the failsafe skip path can distinguish them from wire failures. These skips carry a verdict — the upstream was consulted via the bound check, just not over the wire — and must count toward unanimity.

### 3. Block-availability pre-check skips counted as upstream request errors

\`handleBlockSkip\` was incrementing \`MetricUpstreamErrorTotal\` for block availability pre-check skips. No wire call is made in this path, so error metrics were inflated with expected, by-design gate outcomes.

### 4. Unanimous block-unavailable consensus counted as \`consensus_on_error\`

When all upstreams unanimously agree the requested block is outside the availability window (\`hasConsensus=true\` + \`ErrUpstreamBlockUnavailable\`), the consensus records \`outcome=consensus_on_error\`. \`ErrUpstreamBlockUnavailable\` is \`SeverityWarning\` (correct for a single lagging upstream), which passed the severity guard and inflated \`MetricConsensusErrors\` with deterministic client behaviour.

## Fix

**\`upstream/upstream.go\`** — For the upper-bound branch, compute \`val = 0 - latestBlockMinus\` when the state poller has not delivered a head yet. For \`latestBlockMinus: 0\` this resolves to \`maxBound = 0\`, so any block > 0 exceeds the bound and the upstream is skipped pre-request with \`ErrUpstreamBlockUnavailable\`. Track \`upperFromZeroLatest\` to prevent the invalid-range safety net (\`min > max → fail-open\`) from undoing this intentional restriction.

**\`common/json_rpc.go\`** — In \`isBlockUnavailableVerdict\`, distinguish two kinds of \`ErrUpstreamRequestSkipped\` causes: those wrapping \`ErrUpstreamBlockUnavailable\` (pre-check verdict — count toward unanimity) and all others such as method filter, cordon, or selector skips (neutral — neither agree nor disagree, so a pool of pure neutral skips cannot claim \`-32014\`).

**\`erpc/networks.go\`** — Switch \`handleBlockSkip\` from \`MetricUpstreamErrorTotal\` to \`MetricUpstreamSkippedTotal\`. Pre-check skips are gate decisions, not wire failures.

**\`consensus/executor.go\`** — Guard \`MetricConsensusErrors\` with \`blockUnavailableConsensus\` (\`hasConsensus && HasErrorCode(result.Error, ErrCodeUpstreamBlockUnavailable)\`). Unanimous block-unavailable is a deterministic gate outcome — the client requested a block no upstream will serve — and must not inflate infra error counters.

## Tests

- \`upstream\`: existing \`TestEvmAssertBlockAvailability\` suite covers the new upper-bound behaviour; new \`upperFromZeroLatest\` flag is exercised by the invalid-range guard path.
- \`common\`: two new sub-tests in \`TestTranslateToJsonRpcException_BlockUnavailableIsMissingData\` — \`all pre-check skipped due to block unavailable\` (must be \`-32014\`) and \`all skipped for non-block-unavailable reason\` (must not be \`-32014\`).
- \`consensus\`: new sub-test \`block-unavailable unanimous consensus skips consensus_errors_total\` in \`TestRecordMetricsAndTracing_InfoSeverityNotCountedAsConsensusError\`.
- All four affected packages pass \`go test ./... -count=1\`.

### DIFF
--- a/erpc/networks.go
+++ b/erpc/networks.go
@@ -2752,7 +2752,7 @@ func (n *Network) handleBlockSkip(
 		attribute.Bool("skip_retryable", isRetryable),
 	)
 	finality := req.Finality(ctx)
-	telemetry.MetricUpstreamSkippedTotal.WithLabelValues(
+	telemetry.CounterHandle(telemetry.MetricUpstreamSkippedTotal,
 		n.projectId, u.VendorName(), n.Label(), u.Id(), method,
 		finality.String(), req.UserId(), req.AgentName()).Inc()
 	errToStore := skipErr
--- a/erpc/networks.go
+++ b/erpc/networks.go
@@ -2752,11 +2752,9 @@ func (n *Network) handleBlockSkip(
 		attribute.Bool("skip_retryable", isRetryable),
 	)
 	finality := req.Finality(ctx)
-	telemetry.CounterHandle(telemetry.MetricUpstreamErrorTotal,
+	telemetry.MetricUpstreamSkippedTotal.WithLabelValues(
 		n.projectId, u.VendorName(), n.Label(), u.Id(), method,
-		common.ErrorFingerprint(skipErr), string(common.SeverityInfo),
-		req.CompositeType(), finality.String(),
-		req.UserId(), req.AgentName()).Inc()
+		finality.String(), req.UserId(), req.AgentName()).Inc()
 	errToStore := skipErr
 	if !isRetryable {
 		errToStore = common.NewErrUpstreamRequestSkipped(skipErr, u.Id())
--- a/common/block_availability_project_test.go
+++ b/common/block_availability_project_test.go
@@ -115,6 +115,44 @@ func TestTranslateToJsonRpcException_BlockUnavailableIsMissingData(t *testing.T)
 				NewErrFailsafeRetryExceeded(ScopeNetwork, exhausted, nil))))
 	})
 
+	// All upstreams pre-check skipped due to block-unavailable (e.g. every upstream
+	// exceeded its lower/upper availability bound). The skip carries a verdict —
+	// the upstream was consulted, just not over the wire — so all-skipped is
+	// still unanimity and must return -32014.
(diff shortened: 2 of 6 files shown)


erpc/networks.go
```diff
@@ -2752,11 +2752,9 @@ func (n *Network) handleBlockSkip(
 		attribute.Bool("skip_retryable", isRetryable),
 	)
 	finality := req.Finality(ctx)
-	telemetry.CounterHandle(telemetry.MetricUpstreamErrorTotal,
+	telemetry.MetricUpstreamSkippedTotal.WithLabelValues(
```
aramalipoor: Why aren't we using "telemetry.CounterHandle"?

andreclaro: good catch. we should use it! updated PR!

PR #1089, opened 2026-08-19
docs(specs): unified metrics manager with metrics.customizations

andreclaro:
## Summary

Spec-only PR for a **unified metrics manager** — one owner in `telemetry/` for how every metric is defined, registered, and (for labeled kinds) projected — with a single operator surface, `metrics.customizations`, covering family exposure, label projection, and per-family histogram buckets.

- `specs/unified-metrics-manager/feature.md` — manager + `MetricPolicy` design (`Define*` at init, `Configure` after config), `customizations` semantics (`keep`/`drop`, specificity precedence, `buckets: []float64`), stock-collector scrape filter, legacy label-knob deprecation
- `specs/unified-metrics-manager/plan.md` — Phase 1 manager + customizations (no call-site changes) → Phase 2 unified handles + no-op path → Phase 3 allocation optimization

Key design points:
- Today's three registration paths (deferred counters, deferred histograms, ~87 `promauto`) make any global policy a per-kind bolt-on; the manager collapses them into one define → configure → register path
- Operator surface is **one list**: `metrics.customizations` with `subject` + optional `action` / `labels` / `buckets` — not parallel `exposeMetrics`/`dropMetrics` + kind-split label knobs
- Allowlists are explicit: `subject: "*"`, `action: drop`, then `action: keep` carve-outs; overlapping subjects resolve by **specificity** (exact > longer prefix > `*`), not list order
- Dropped eRPC families are **never registered** (scrape + collection cost gone); `FilteredGatherer` only covers stock `go_*`/`process_*`/`promhttp_*` collectors
- Gauges are never label-projected; all histograms go through `DefineLabeledHistogram` so label and bucket rules apply
- Legacy `histogramDropLabels` / `counterDropLabels` (+ overrides) stay as deprecated aliases with kind masks; Phase 1 changes no call sites; default config stays scrape-compatible

## Test plan

- [ ] Spec review only — no code changes in this PR
- [ ] Implementation phases have their own acceptance criteria in `plan.md`


specs/unified-metrics-manager/feature.md
```diff
@@ -0,0 +1,223 @@
...
+
+Consequences of the split:
+
+- **No single gate.** "Don't expose family X" today requires three different mechanisms (skip a register call, skip another register call, unregister after the fact) — and the third requires a name → collector mapping that doesn't exist.
+- **Policy logic is per-kind.** Label projection is implemented twice (`HistogramLabelFilter`, `CounterLabelFilter`) and is unavailable for gauges.
+- **No no-op path.** Every call site pays full Prometheus cost even for families an operator will never scrape.
+
+### Why scrape-time filtering alone is insufficient
+
+A `FilteredGatherer` wrapping `prometheus.DefaultGatherer` (filtering `[]*dto.MetricFamily` after `Gather()`) is cheap to build and does shrink the HTTP response, but it still:
```
aramalipoor: Wwe would still benefit if we can actually make the final metrics page smaller (in our prod is around 160mb where it would be super useful to reduce it).

Maybe the current design already makes it that way, but if not then having a custom gatherer to eliminate metrics/labels that we don't want is a good idea.

specs/unified-metrics-manager/feature.md
```diff
@@ -0,0 +1,223 @@
...
+
+  # Allowlist — when non-empty, ONLY these families are exposed
+  exposeMetrics:
+    - upstream_*                       # whole subsystem (trailing prefix)
+    - network_request_duration_seconds # single family
+
+  # Denylist — remove families from exposure (applied after exposeMetrics)
+  dropMetrics:
+    - consensus_*                      # whole subsystem
+    - network_evm_block_range_requested_total
```
aramalipoor: Maybe `include:`/`exclude:` would be cleaner vs repeating "metrics" word.

It would be great to also include labels? Perhaps the elements here could be an object:
```
metrics:
  // ...
  exclude:
    - subject: "consensus_*"
       keepLabels: ["user", "category"]
       dropLabels: ["agent_name"]
```

---

Another way to design it could be:

```
metrics:
  // ...
  customizations:
    - subject: "consensus_*"
       action: drop
       labels:
           - subject: "agent_*"
              action: drop
           - subject: "agent_name"
              action: keep
```

specs/unified-metrics-manager/feature.md
andreclaro: Yes, that's the goal of the drop metrics features... we have a lot of resource load on our metrics agents due to the amount of metrics that are exposed... even not scrapping them all.

I will make it clear in the design as well.

specs/unified-metrics-manager/feature.md
andreclaro: I love this idea of dropping labels as well, because we also don't want all the labels for each metric.

@aramalipoor, any preference between option 1 and 2? 🤔

aramalipoor: I'm leaned more towards "customizations" option, 2.

andreclaro:
Additional comments:Aram: "it would be great to have a solution for Histograms too. ... basically a generic way to override histograms buckets for some metrics ... i think it's already in the plan but good to emphasize to remove/deprecate any old way of dropping metrics... we could have backward-compat only on config level, but not necessary if complex"

aramalipoor:
@andreclaro let me know when ready to merge, also let's make sure it's updated with our latest features and we also have a unified config for metrics/label dropping (I think it's already done but just in case we have left-overs)

andreclaro:
@aramalipoor ready to merge from my side.

Specs are updated to the agreed `metrics.customizations` surface (`keep`/`drop`, specificity precedence, per-family `buckets`) — one list for family exposure + label projection + histogram bucket overrides. The old `exposeMetrics`/`dropMetrics` draft is gone; the four legacy label knobs stay as deprecated aliases only.

Also refreshed against current main (open-ended family set / no stale catalog assumptions). Leftover dual-knob design is called out explicitly in the deprecations section.

andreclaro:
@aramalipoor CI isn't red on test failures — the workflows (`test`, CodeQL, Dependency Review, xray) are sitting in `action_required` after the force-push from my fork. Specs-only change; once you approve the pending workflow runs they should go green the same way they did on the previous tip.

https://github.com/erpc/erpc/actions/runs/33500794880

andreclaro:
implemented here: https://github.com/erpc/erpc/pull/1128

PR #1093, opened 2026-08-20
fix(shadow): gate block-hash selectors by the primary's resolved number

aramalipoor:
## The gap

The shadow availability gate (#1090) extracts a block number from the **request** — so a block-**hash** selector, which carries no number, always failed open and was mirrored regardless of whether the shadow upstream could hold the block yet.

That combination is exactly the tip-follow pattern: indexers select blocks by hash for reorg safety, at the chain head. A replica trailing the head by seconds cannot have that block at the moment the mirror fires. Measured on a live shadow stream: **~23/s of guaranteed "unknown hash" errors** — the single largest error class the stream produced, invisible to request-side extraction by construction.

## The fix

The primary has already answered by the time the shadow runs, so its response carries the resolved block number (`resp.EvmBlockNumber()`). When request extraction yields nothing, the gate falls back to that.

Fail-open shape unchanged: a tag, an unparseable ref, a non-EVM upstream, an errored assertion, or a response without a number all mirror exactly as before. Only a concrete height the upstream states it does not have is skipped.

### DIFF
--- a/erpc/shadow.go
+++ b/erpc/shadow.go
@@ -80,7 +80,26 @@ func (p *PreparedProject) executeShadowRequests(ctx context.Context, network *Ne
 		// unparseable ref, a non-EVM upstream or an errored assertion all
 		// mirror as before. Only a CONCRETE height the upstream states it
 		// does not have is skipped.
-		if _, blockNumber, err := evm.ExtractBlockReferenceFromRequest(ctx, origReq); err == nil && blockNumber > 0 {
+		blockNumber := int64(0)
+		if _, bn, err := evm.ExtractBlockReferenceFromRequest(ctx, origReq); err == nil && bn > 0 {
+			blockNumber = bn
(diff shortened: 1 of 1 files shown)


aramalipoor:
Withdrawing this after tracing the mechanism through the codebase: a response's EvmBlockNumber is populated from the response body via each method's RespRefs (extractRefFromJsonRpcResponse), and the eth_call-family responses — the bulk of what shadow mirrors — carry no block reference, so the fallback added here can never fire for exactly the traffic it was written for. Where it can fire (block/receipt-shaped responses), gating a HASH selector by number availability is branch-blind: having height H does not imply having that hash. The shadow upstream now briefly parks unknown hashes instead of failing fast, and with its tip lag reduced the guaranteed-miss class this tried to gate measures ~0 — so there is nothing left for this gate to remove. If a residual reappears, the right shape is classifying the upstream's typed not-yet refusal as skip-comparison rather than guessing block numbers on the request side.

PR #1094, opened 2026-08-20
chore(evm): classify mempool priority rejections as ExecutionException

andreclaro:
## Summary

- Reclassify specific `eth_sendRawTransaction` upstream mempool rejections from `ErrEndpointServerSideException` to `ErrEndpointExecutionException` so circuit breakers do not trip on deterministic client-side pool policy failures.
- Matched messages (observed in production on EVM chains):
  - `An existing transaction had higher priority`
  - `A newer transaction had higher priority`
  - bare `rejected`
- No network retry for sendRaw — upstreams typically agree on these rejections.

## Problem

Some upstreams return JSON-RPC `-32603` with mempool policy messages that do not match existing normalizer patterns (reverts, nonce, insufficient funds, etc.). These fall through to the `ServerSideException` fallback, which counts toward circuit breaker failure thresholds even though the upstream is healthy and answered correctly.

## Test plan

- [x] `go test ./architecture/evm/ -run TestExtractJsonRpcError_MempoolPriorityRejections`
- [x] Negative case: `"try a higher priority fee"` is not reclassified

### DIFF
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -379,6 +379,20 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 			)
 		}
 
+		if strings.Contains(ml, "existing transaction had higher priority") ||
+			strings.Contains(ml, "newer transaction had higher priority") ||
+			strings.TrimSpace(ml) == "rejected" {
+			return common.NewErrEndpointExecutionException(
+				common.NewErrJsonRpcExceptionInternal(
+					int(code),
+					common.JsonRpcErrorTransactionRejected,
+					err.Message,
+					nil,
+					details,
+				),
+			)
+		}
+
 		//----------------------------------------------------------------
 		// "Insufficient funds / balance" errors
 		// Note: This comes AFTER nonce/duplicate detection to avoid masking those errors.
--- a/architecture/evm/error_normalizer.go
+++ b/architecture/evm/error_normalizer.go
@@ -56,11 +56,16 @@ func ExtractJsonRpcError(r *http.Response, nr *common.NormalizedResponse, jr *co
 				msg += " Data: " + s
 			}
 		default:
-			// passthrough error data as is
-			details["data"] = err.Data
+			// Absent "data" unmarshals to a nil interface; appending it would
+			// turn e.g. Monad's bare "rejected" into "rejected Data: <nil>" and
+			// silently break every whole-message (equality/suffix) check below.
+			if err.Data != nil {
+				// passthrough error data as is
+				details["data"] = err.Data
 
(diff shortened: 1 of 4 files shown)


kasrakhosravi:
Pushed two changes onto this branch (rebased over #1095 as well):

**1. The bare `"rejected"` match never fired in production.** The normalizer appends ` Data: <nil>` to the matching buffer whenever the upstream error has no `data` field — which is exactly Monad's wire shape (`JsonRpcError::custom` uses `skip_serializing_if`). So `strings.TrimSpace(ml) == "rejected"` compared `"rejected data: <nil>"` and always fell through to `ServerSideException`, tripping the breaker anyway. The unit test passed because it modeled data as `""` (empty string), which skips the append. Fixed at the root (skip the append for nil data) plus the equality now checks the pristine `err.Message`. Reproduced and verified end-to-end with a live eRPC against mock upstreams returning `{"code":-32603,"message":"rejected"}`.

**2. Kept eth_sendRawTransaction failover; scoped the change to breaker/scoring.** These messages come from Monad's [`EthTxPoolDropReason`](https://github.com/category-labs/monad-bft/blob/master/monad-eth-txpool-types/src/lib.rs) (`ExistingHigherPriority` / `ReplacedByHigherPriority`; bare `rejected` is the `TxStatus::Evicted` path in [`monad-rpc`](https://github.com/category-labs/monad-bft/blob/master/monad-rpc/src/handlers/eth/txn.rs)). Pool contents and min-fee policies are node-local, and pre-PR these errors (as `ServerSideException`) did fail over — so returning a terminal error changed two behaviors at once. The branch now sets `WithRetryableTowardNetwork(true)` for `eth_sendRawTransaction`, identical to the adjacent `-32003` branch; the net change vs `main` is only that rejections stop feeding the circuit breaker. Also added `"Transaction fee too low"` (`FeeTooLow` from the same enum — by far the most common pool rejection), which otherwise still opened breakers.

**E2E verification** (live erpc, 2 mock upstreams, breaker 3-of-5, network retry 3):
| scenario | before | after |
|---|---|---|
| priority rejection | **both breakers open after 1 tx; reads dead** (`eth_blockNumber` → breaker open) | failover attempted, clean `-32003` answer, breakers closed |
| `Transaction fee too low` | both breakers open | breakers closed |
| bare `rejected` (no `data` field) | both breakers open (match never fired) | breakers closed |
| up1 rejects → up2 accepts | — | client gets tx hash |
| `try a higher priority fee` (control) | breaker counts | breaker still counts ✓ |

Tests: rewrote the unit test with wire-shape (nil-data) cases + a non-sendRaw terminal case, added two network-level tests in `networks_sendrawtx_test.go` (failover-succeeds, all-reject-surfaces-ExecutionException), documented rule 12b in `docs/pages/reference/errors.mdx`.

PR #1096, opened 2026-08-20
fix(erpc): enable h2c without shared gRPC

### LINKED ISSUE #965: bug: cleartext http2 not enabled by default
**Describe the bug**
Cleartext http2 should be enabled by defualt, but right now it only gets enabled with either grpc or tls integration.

**To Reproduce**
Start eRPC without tls or grpc, then try:

`curl -i --http2-prior-knowledge http://localhost:4000/main/evm/8453`

**Expected behavior**
Connection should work, but doesn't.

Parikalp-Bhardwaj:
## Summary

Enable cleartext HTTP/2 (h2c) on the non-TLS IPv4 listener even when shared gRPC is disabled.

Previously, `h2c.NewHandler` was only installed inside the `grpcSharesHttpV4` path. As a result, HTTP/2 prior-knowledge clients could not connect when running without TLS and shared gRPC.

This change decouples h2c support from gRPC while preserving the existing shared-gRPC behavior.

Closes #965

## Changes

- Enable h2c whenever the IPv4 listener runs without TLS
- Add a transport-level regression test using a real cleartext HTTP/2 connection
- Verify HTTP/1.1 continues to work
- Preserve the existing shared-gRPC behavior
- Document the h2c behavior

## Before

With TLS disabled and shared gRPC disabled, normal HTTP/1.1 requests worked:

```bash
curl -v --http1.1 \
  http://127.0.0.1:4000/main/evm/8453
```

Result:

```text
< HTTP/1.1 200 OK
< Content-Type: application/json
< X-Erpc-Version: dev
```

However, a cleartext HTTP/2 prior-knowledge request failed:

```bash
curl -v --http2-prior-knowledge \
  http://127.0.0.1:4000/main/evm/8453
```

Result:

```text
> GET /main/evm/8453 HTTP/2

Remote peer returned unexpected data while we expected SETTINGS frame.
Perhaps, peer does not support HTTP/2 properly.

curl: (16) Remote peer returned unexpected data while we expected SETTINGS frame.
```

The TCP connection itself succeeded, but the cleartext HTTP listener did not speak HTTP/2.

The reason was that `h2c.NewHandler` was only installed inside the shared-gRPC branch:

```go
if grpcSharesHttpV4(cfg) {
    // shared gRPC dispatch

    if cfg.TLS == nil || !cfg.TLS.Enabled {
        handlerV4 = h2c.NewHandler(handlerV4, &http2.Server{})
    }
}
```

When gRPC was disabled, `grpcSharesHttpV4(cfg)` returned false and the h2c wrapper was never installed.

As a result, the non-TLS listener continued serving HTTP/1.x only.

## After

The h2c wrapper is now applied independently of shared gRPC whenever TLS is disabled:

```go
if grpcSharesHttpV4(cfg) {
    // shared gRPC dispatch
}

if cfg.TLS == nil || !cfg.TLS.Enabled {
    handlerV4 = h2c.NewHandler(handlerV4, &http2.Server{})
}
```

With this change, the same HTTP/2 prior-knowledge request succeeds:

```bash
curl -v --http2-prior-knowledge \
  http://127.0.0.1:4000/main/evm/8453
```

Result:

```text
> GET /main/evm/8453 HTTP/2

< HTTP/2 200
< content-type: application/json
< x-erpc-version: dev
```

Response:

```json
{"main":[{"id":"evm:8453","state":"OK"}]}
```

Cleartext HTTP/2 also succeeds through `localhost`:

```bash
curl -v --http2-prior-knowledge \
  http://localhost:4000/main/evm/8453
```

Result:

```text
> GET /main/evm/8453 HTTP/2
< HTTP/2 200
```

Normal HTTP/1.1 requests continue to work:

```bash
curl -v --http1.1 \
  http://127.0.0.1:4000/main/evm/8453
```

Result:

```text
< HTTP/1.1 200 OK
```

So the behavior changes from:

```text
Before:

HTTP/1.1                  ✅ works
HTTP/2 prior knowledge    ❌ fails
```

to:

```text
After:

HTTP/1.1                  ✅ works
HTTP/2 prior knowledge    ✅ works
```

## Root Cause

Cleartext HTTP/2 support was coupled to shared gRPC configuration.

The server only installed:

```go
h2c.NewHandler(handlerV4, &http2.Server{})
```

when:

```go
grpcSharesHttpV4(cfg)
```

was true.

That condition requires gRPC to be enabled and configured to share the same IPv4 host and port as the HTTP server.

Therefore, with the default non-TLS configuration and gRPC disabled, h2c was never enabled.

A plain Go `http.Server` using `ListenAndServe()` serves HTTP/1.x over a cleartext connection unless h2c support is explicitly added.

The fix separates the two concerns:

- Shared gRPC controls gRPC request dispatch
- TLS state controls whether the cleartext HTTP handler needs h2c

## Testing

Added:

```text
TestHttpServer_H2C_WithoutSharedGrpc
```

The test uses a real listener and a real cleartext HTTP/2 client rather than directly invoking the HTTP handler.

It verifies that:

- TLS is disabled
- Shared gRPC is disabled
- The server accepts a real h2c connection
- `resp.ProtoMajor == 2`
- HTTP/1.1 continues to work

The regression test fails on `main` with an HTTP/2 protocol error because the server responds as HTTP/1.x instead of sending an HTTP/2 SETTINGS frame.

It passes with this change.

Existing shared-gRPC behavior was also verified with:

```text
TestHttpServer_CanSharePortWithGrpc
```
## Scope

This PR changes the IPv4 HTTP handler path covered by #965.

It keeps the existing:

- HTTP/1.1 behavior
- shared-gRPC behavior
- TLS behavior

unchanged.

The change is intentionally limited to decoupling cleartext HTTP/2 support from the shared-gRPC configuration.


Parikalp-Bhardwaj:
Hi @aramalipoor, When you get a chance, could you please review this PR? I’ve added the fix and regression coverage for the h2c issue, thanks

aramalipoor:
Great contribution @Parikalp-Bhardwaj 🚀

PR #1097, opened 2026-08-21
spec(served-tip): group-aware served tip

andreclaro:
This PR proposes the design for group-aware served tip.

## Problem
Mixed provider pools (e.g. `type:external` + `type:internal`) can have one group consistently lag the other. The current global strict-majority tip can advance to a block only the fast group has, breaking consensus for requests that later need participation from each group.

## Proposal
Add `networks[].evm.servedTip.requiredGroups`: a list of upstream selectors. When configured, the network-wide latest/finalized tip becomes the minimum of the strict-majority tips computed independently inside each selector-matched group. This guarantees the advertised block is servable by at least one upstream from every required group.

## Deliverables
- `specs/servedtip-group-aware/feature.md` — full specification: config schema, algorithm, interactions with existing guards/referee/metrics, test scenarios, backward compatibility.
- `specs/servedtip-group-aware/plan.md` — phased implementation plan.

## Key design decisions
- Group selectors reuse existing `use-upstream` selector syntax.
- Trajectory referee is disabled for the network-wide lane when groups are active (raising the tip above the slowest group would violate the guarantee).
- Regression guard uses `min(per-group corroborated/max)` as reference so the deliberate fast/slow gap is not treated as a regression.
- Empty group selectors are warned and ignored (fail-open) in v1.
- Per-group metrics expose the bottleneck group.

## Implementation reference
A reference implementation exists on `andreclaro/erpc#feat/servedtip-group-aware` (not part of this PR) for reviewers who want to see how the spec maps to code. It will be proposed separately once the design is settled.

Looking for feedback on the design before moving forward.

### DIFF
--- a/specs/servedtip-group-aware/feature.md
+++ b/specs/servedtip-group-aware/feature.md
@@ -1,0 +1,242 @@
+# Group-Aware Served Tip — Specification
+
+> Let operators declare upstream groups (via existing tag/id selectors) and
+> compute the network-wide served `latest`/`finalized` tip as the **intersection
+> of group-majorities**: the highest block that every required group can serve.
+> This guarantees that later requests routed through consensus requiring one
+> participant from each group can always be served by at least one upstream from
+> every group.
+
+## 1. Problem
+
+eRPC's served tip is today a strict majority order statistic over all eligible
+upstreams (`evm.PickServedTip`), or over a single selector-scoped subset for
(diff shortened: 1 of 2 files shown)


andreclaro:
something smilar was implemented here: https://github.com/erpc/erpc/pull/1166

Issue #1034, opened 2026-08-01
bug: CacheHash maps distinct valid requests to the same key


imarshallwidjaja:
**Describe the bug**

`JsonRpcRequest.CacheHash` (`common/json_rpc.go`) builds the cache/multiplex key by writing each param into one SHA-256 stream with no type tags, lengths, separators, or container boundaries, and formats `float64` with `%f`. Two requests with different params can feed identical bytes into the hash and land on the same key. That key drives in-flight multiplexing and the persistent cache range key, so the second request gets the first request's response.

**To Reproduce**

1. Use erpc version `0.1.1` (`45656d6`). Current `main` (`e06be39`) has the same code.

2. Any config. Multiplexing is on by default; a cache policy covering the method makes the crossover durable.

3. Two valid `eth_getLogs` filters that differ only in topic structure:

```bash
# topics[0] = A or B, topic[1] unconstrained
curl --location 'http://localhost:4000/main/evm/1' \
--header 'Content-Type: application/json' \
--data '{"method":"eth_getLogs","params":[{"fromBlock":"0x64","toBlock":"0x64","address":"0x0000000000000000000000000000000000000001","topics":[["0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"]]}],"id":1,"jsonrpc":"2.0"}'

# topics[0] = A and topics[1] = B, a different and narrower result set
curl --location 'http://localhost:4000/main/evm/1' \
--header 'Content-Type: application/json' \
--data '{"method":"eth_getLogs","params":[{"fromBlock":"0x64","toBlock":"0x64","address":"0x0000000000000000000000000000000000000001","topics":["0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa","0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"]}],"id":1,"jsonrpc":"2.0"}'
```

Both flatten to the same bytes and produce the same key:

```
[...]

PR #1098, opened 2026-08-22
fix(common): prevent JSON-RPC CacheHash aliasing

Parikalp-Bhardwaj:
## Summary

Fix `JsonRpcRequest.CacheHash` so distinct valid JSON-RPC requests cannot map to the same cache/multiplex key due to ambiguous parameter encoding.

Previously, request parameters were recursively written into the hash stream without preserving enough structural information such as type, container boundaries, or element boundaries. `float64` values were also formatted using `%f`, which could lose precision.

As a result, semantically different requests could produce identical bytes before SHA-256 hashing and therefore share the same `CacheHash`.

This affects both persistent caching and in-flight request multiplexing.

Related #1034

## Root Cause

The previous `hashValue` implementation flattened parameter values directly into a single hash stream.

For example, these two different topic structures:

```json
[
  [
    "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
    "0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
  ]
]
```

and:

```json
[
  "0xaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa",
  "0xbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb"
]
```

could produce the same input bytes because array/container boundaries were not encoded.

The same problem also existed for values such as:

```json
[25]
```

and:

```json
[25.0000001]
```

because `%f` could serialize both as:

```text
25.000000
```

This is not a SHA-256 collision. The problem was the ambiguous serialization before hashing.

## Changes

- Make JSON-RPC request hashing structure-aware
- Preserve JSON value types and container boundaries
- Preserve array ordering and element boundaries
- Keep object hashing deterministic by sorting keys
- Preserve full floating-point precision
- Prevent ambiguous concatenation of request values
- Add regression tests for the collisions reported in #1034

## Before

Two semantically different `eth_getLogs` requests could produce the same cache key:

```text
topics = [["A", "B"]]

topics = ["A", "B"]
```

Both could be flattened into equivalent bytes and therefore produce the same `CacheHash`.

Similarly:

```json
["0x10", "0x64", [25]]
```

and:

```json
["0x10", "0x64", [25.0000001]]
```

could produce the same key because float formatting lost precision.

This meant:

```text
Request A
    ↓
CacheHash X

Request B
    ↓
CacheHash X
```

and the requests could incorrectly share a multiplexed or cached response.

## After

The request encoding now preserves the information required to distinguish request identity.

The same examples now produce different hashes:

```text
topics = [["A", "B"]]
        ↓
CacheHash A

topics = ["A", "B"]
        ↓
CacheHash B

CacheHash A != CacheHash B
```

and:

```text
25
        ↓
CacheHash A

25.0000001
        ↓
CacheHash B

CacheHash A != CacheHash B
```

## Compatibility

Changing the request encoding changes the generated `CacheHash` values.

Existing persistent cache entries created with the old hashing scheme will not be reused by the new implementation and will effectively become cold/stale entries until they expire or are evicted.

No cache migration is required because the old entries are not referenced by the new keys.

## Scope

This PR is intentionally limited to JSON-RPC request identity hashing and its regression tests.

It does not change:

- cache storage behavior
- request routing
- multiplexing logic
- upstream selection
- response canonicalization
- unrelated SVM behavior

### DIFF
--- a/common/json_rpc.go
+++ b/common/json_rpc.go
@@ -1478,24 +1478,63 @@ func (r *JsonRpcRequest) PeekByPath(path ...interface{}) (interface{}, error) {
 	return current, nil
 }
 
+// hashValue writes a deterministic, structure-preserving encoding of v into h,
+// used to derive request-identity keys (CacheHash / in-flight multiplexing).
+//
+// The old encoding recursively flattened params into the hash stream with no
+// type tags, lengths, or container markers, so structurally distinct requests
+// collapsed to the same byte stream and therefore the same key (#1034):
+// topics [[A,B]] and [A,B] were indistinguishable, "ab"+"c" aliased "a"+"bc",
+// and %f rounded 25.0000001 to "25.000000". Because that key drives both the
+// persistent cache and in-flight multiplexing, a follower request could be
+// handed the leader's response.
+//
+// This encoding is injective across JSON value shapes: every value carries a
+// one-byte type tag, every string/number/container is length- or count-
+// delimited, and object keys are emitted in sorted order. Two self-delimiting
+// tokens cannot concatenate into a third, so different params can never produce
+// the same bytes.
+//
+//	null    -> "N"
+//	bool    -> "B0" | "B1"
+//	number  -> "F" <len> ":" <shortest-round-tripping decimal>
+//	string  -> "S" <len> ":" <lowercased bytes>
+//	array   -> "A" <count> ":" <encoded element>...
+//	object  -> "O" <count> ":" ("K" <len> ":" <key>) <encoded value> ...
+//
+// String params are lowercased, preserving the long-standing EVM hex
+// normalization: a checksummed address and its lowercase form are the same
+// request and share a cache entry / multiplex group. The type tags make that
+// safe against cross-type aliasing that lowercasing alone would create — the
+// string "true" (S4:true) and the boolean true (B1) no longer collide. (SVM
+// needs the opposite, case-preserving key; see svmRequestKey, which is why it
+// does not route through this function.)
 func hashValue(h io.Writer, v interface{}) error {
 	switch t := v.(type) {
+	case nil:
+		_, err := io.WriteString(h, "N")
+		return err
 	case bool:
-		_, err := io.WriteString(h, fmt.Sprintf("%t", t))
+		s := "B0"
+		if t {
+			s = "B1"
+		}
+		_, err := io.WriteString(h, s)
 		return err
 	case int:
-		_, err := io.WriteString(h, fmt.Sprintf("%d", t))
-		return err
+		return writeHashToken(h, 'F', strconv.FormatInt(int64(t), 10))
 	case float64:
-		_, err := io.WriteString(h, fmt.Sprintf("%f", t))
-		return err
+		// 'g'/-1 emits the shortest decimal that round-trips to the same
+		// float64, preserving full precision (25 vs 25.0000001) unlike %f.
+		return writeHashToken(h, 'F', strconv.FormatFloat(t, 'g', -1, 64))
 	case string:
-		_, err := io.WriteString(h, strings.ToLower(t))
-		return err
+		return writeHashToken(h, 'S', strings.ToLower(t))
 	case []interface{}:
+		if err := writeHashHeader(h, 'A', len(t)); err != nil {
+			return err
+		}
 		for _, i := range t {
-			err := hashValue(h, i)
-			if err != nil {
+			if err := hashValue(h, i); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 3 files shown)


aramalipoor:
@Parikalp-Bhardwaj thanks for the PR can you run benchmarks of before/after? This is a very sensitive method and we shouldn't impact the allocations nor the ops/sec otherwise it might hinder production workloads and cause CPU spikes.

Parikalp-Bhardwaj:
@aramalipoor thanks for pointing that out. I’ll add a focused `CacheHash` benchmark and compare the current main implementation against this PR for `ns/op`, `B/op`, and `allocs/op`. I’ll share the before/after results here and adjust the implementation if there’s any meaningful regression.

Parikalp-Bhardwaj:
@aramalipoor thanks, I ran a focused before/after benchmark for `JsonRpcRequest.CacheHash` using the same benchmark implementation and environment.

The benchmark calls `InvalidateCacheHash()` on every iteration, so it measures actual hash generation rather than the memoized fast path. Request construction/parsing is outside the timed section.

Command:

```bash
go test ./common -run '^$' -bench '^BenchmarkCacheHash_' -benchmem -count=10
```

Environment:
```bash
goos: darwin
goarch: arm64
pkg: github.com/erpc/erpc/common
cpu: Apple M3 Pro
BenchmarkCacheHash_Simple-12             3536488               324.9 ns/op           176 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3591993               329.9 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3629407               330.3 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3484622               342.2 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3509943               344.8 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3037644               334.9 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3637490               331.1 ns/op           176 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3574176               336.9 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3528998               349.8 ns/op           176 B/op          5 allocs/op
BenchmarkCacheHash_Simple-12             3520762               338.8 ns/op           177 B/op          5 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1652978               726.2 ns/op           241 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1647519               718.6 ns/op           241 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1648586               757.2 ns/op           240 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1639220               727.9 ns/op           240 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1660053               737.5 ns/op           241 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1665720               721.2 ns/op           241 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1634395               752.4 ns/op           240 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1628260               729.8 ns/op           241 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1606330               735.6 ns/op           241 B/op          6 allocs/op
BenchmarkCacheHash_EthGetLogs-12         1664241               720.7 ns/op           240 B/op          6 allocs/op
PASS
ok      github.com/erpc/erpc/common     35.203s
```

All measured deltas were statistically significant with `n=10`.

So this version does not introduce a performance or allocation regression in these workloads, it improves both execution time and allocation pressure while preserving the structure-aware hashing fix for #1034.

I also added the benchmark to the PR so the comparison can be reproduced.

Parikalp-Bhardwaj:
@aramalipoor This appears to be a CodeQL false positive.

`CacheHash` is not used for password storage, authentication, or credential verification. It derives a deterministic internal identity for JSON-RPC requests used by caching and in-flight multiplexing.

The hashed input is the JSON-RPC method/params from the request body, not a password or API credential. I also traced the `CacheHash` consumers and confirmed the result is only used as an internal cache/multiplex key.

SHA-256 already existed in the previous `CacheHash` implementation. This PR changes the request serialization and buffer strategy, which made the dataflow visible to CodeQL.

A password KDF such as bcrypt/Argon2 would be inappropriate for this per-request hot path and would introduce a significant performance regression.

Requesting dismissal as a false positive.

aramalipoor:
Thanks for finding and the work on this @Parikalp-Bhardwaj 🙌

PR #1100, opened 2026-08-22
feat(shadow): pinBlockTag — evaluate mirrored tag selectors at the primary's height

aramalipoor:
## Problem

Shadow mirrors forward a byte-identical copy, so a `latest` selector is re-evaluated at the shadow upstream's own head — wall-clock-later than the primary's answer. Any read of volatile state (pool reserves, oracle prices, most visibly large `aggregate3` multicalls) then diverges **by construction**, and `shadow_response_mismatch_total` reports a comparison artifact rather than a correctness signal. On a live mirror this class runs at a steady ~10/min floor that no upstream-side fix can remove.

## Change

`shadow.pinBlockTag` (default `false` = byte-identical mirroring, unchanged) rewrites `latest` in the mirrored copy to a concrete height before forwarding:

1. the primary's resolved block number when its response carries one (same source as the block-hash availability gating), else
2. the network's tracked head at mirror time — the closest available approximation for methods like `eth_call` whose responses carry no number.

State methods only (`eth_call`, `eth_getBalance`, `eth_getCode`, `eth_getTransactionCount`, `eth_getStorageAt`); for `eth_getBlockByNumber` pinning would change the question, not the evaluation height. Fail-open on every gap: parse trouble, a non-`latest` selector, or no resolvable height forwards the copy untouched.

- `common/config.go`: `ShadowUpstreamConfig.PinBlockTag`.
- `erpc/shadow.go`: the rewrite + wiring.
- `erpc/shadow_pin_test.go`: rewrite-helper coverage.
- TS types regenerated.

### DIFF
--- a/common/config.go
+++ b/common/config.go
@@ -1153,6 +1153,14 @@ type ShadowUpstreamConfig struct {
 	Enabled      bool                `yaml:"enabled" json:"enabled"`
 	SampleRate   *float64            `yaml:"sampleRate,omitempty" json:"sampleRate,omitempty"`
 	IgnoreFields map[string][]string `yaml:"ignoreFields,omitempty" json:"ignoreFields"`
+	// PinBlockTag rewrites a tag block selector ("latest"/"pending") in the mirrored copy to a
+	// concrete height before forwarding: the primary's resolved block number when its response
+	// carries one, else the network's tracked head at mirror time. Without it the shadow executes
+	// the tag at its OWN head wall-clock-later than the primary did, so any read of volatile state
+	// (pool reserves, oracle prices — most visibly large multicalls) diverges by construction and
+	// reports a mismatch that says nothing about correctness. Default false = mirror byte-identical
+	// requests, exactly as before.
+	PinBlockTag bool `yaml:"pinBlockTag,omitempty" json:"pinBlockTag,omitempty"`
 }
 
 // Deprecated: UpstreamIntegrityConfig is a non-functional legacy stub (never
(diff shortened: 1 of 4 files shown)


aramalipoor:
CI note: `TestConsensusPolicy` failed on the units run but passes locally on this branch (`go test ./erpc/ -run TestConsensusPolicy -count=1` → ok, 5.9s). The test lives in the consensus suite (`networks_consensus_quota_test.go` / `consensus_dsl_test.go`) with no file overlap with this diff (shadow.go, config.go, generated types, shadow tests). Re-ran the failed job.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aramalipoor:
CI note, second rerun: this failure is a DIFFERENT test than the first (`TestInitializer_MarkAsFailedUnblocksWaitThenRetries`, while `TestConsensusPolicy` passed this time). It also passes locally on this branch (`-count=3` → ok) and lives in `util/initializer_test.go` — no overlap with this diff. Two distinct timing-sensitive tests across two runs on loaded runners; re-running once more.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

aramalipoor:
CI note, third spin: a third DISTINCT test this time (`TestNetwork_EvmGetLogs/UseCacheWhenOneOfSubRequestsIsAlreadyCached`; the previous two passed). Also passes locally on this branch at `-count=3`. Three different timing-sensitive tests across three runs, all local-green, none sharing a file with this diff (shadow.go/config.go/types/tests) — runner-load flake, not a regression from this change. One more rerun queued; leaving the consolidated evidence here rather than spinning further.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1103, opened 2026-08-24
fix(shadow): pin latest-tag to the serving upstream's polled head

aramalipoor:
## Problem

#1100 added `pinBlockTag`: mirrored `latest`-tag selectors are rewritten to a concrete height so the shadow evaluates the same state the primary did. When the primary's response carried no block number (e.g. `eth_call`), the pin fell back to the network's highest tracked head at mirror time.

That fallback is a MAX over ALL upstreams. Whenever any other vendor's head leads the serving upstream — i.e. on essentially every block boundary — the pin lands on a height the primary never evaluated. The shadow then reads state one block ahead of the primary's answer, and every volatile read reports a mismatch that says nothing about correctness. Enabling the feature raised the mismatch rate ~25x in production measurement; it is currently disabled in config because of this.

## Fix

- The pin height comes from the response's own resolved number when present, else from the **serving** upstream's state poller head.
- That head is read **once at mirror entry** inside `executeShadowRequests` (the clone already carries the serving upstream), keeping all pin logic encapsulated in shadow.go. Entry is separated from the primary's answer by only a goroutine spawn and the response hash — µs-to-ms against a poller that updates ~1/s — while a read deeper in the per-upstream fan-out would drift further for no gain.
- No usable height → no pin (fail-open, matching every other guard on this path).

Residual skew is bounded by the primary's own poll interval (the poller can be at most one poll behind the node's evaluation head), instead of unbounded by whichever vendor happens to lead the network.

## Testing

- `go build ./...`, `go test ./erpc/ -run TestPinBlockTagInBody -count=1` pass.
- `pinBlockTagInBody` itself is unchanged; this PR only changes where the height comes from.

### DIFF
--- a/erpc/projects.go
+++ b/erpc/projects.go
@@ -193,21 +193,7 @@ func (p *PreparedProject) Forward(ctx context.Context, networkId string, nq *com
 					cloneResp.SetHedges(resp.Hedges())
 					cloneResp.SetEvmBlockRef(resp.EvmBlockRef())
 					cloneResp.SetEvmBlockNumber(resp.EvmBlockNumber())
-					// Capture the SERVING upstream's own polled head now,
-					// synchronously: this is the closest observable proxy for
-					// the height at which that upstream just evaluated a
-					// "latest" tag. The shadow goroutine may run arbitrarily
-					// later (semaphore queueing), so reading any head inside
-					// it would drift past the height the primary answered at.
-					var primaryHead int64
-					if ups := resp.Upstream(); ups != nil {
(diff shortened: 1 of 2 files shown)


aramalipoor:
CI note: the `security` job failed with a gosec **loader** error (`could not import github.com/klauspost/compress/zstd` in `architecture/svm/json_rpc_cache.go` — a file this PR does not touch) while reporting `Issues: 0`. Same workflow passes on current main; rerunning the failed job.

Issue #883, opened 2026-05-13
feat: add Prometheus metric for circuit breaker state


andreclaro:
## Problem

Circuit breaker state transitions (closed → open → half-open) in `upstream/failsafe.go`
are only written to zerolog. There is no way to observe circuit breaker activity in
dashboards.

## Proposed metric

| Metric | Type | Labels |
|---|---|---|
| `erpc_upstream_circuit_breaker_state` | gauge | `project`, `vendor`, `network`, `upstream` |

Values: `0`=closed, `1`=open, `2`=half-open.

DavidGoodman101:
Hi, I'd like to take this one 🙌

**One tweak** to the proposal before I open a PR. Breakers are per-(matchMethod, matchFinality), so a single per-upstream gauge would collapse a method-specific outage (e.g. only trace_* breaking) into a "healthy" 0 and hide it. I'd suggest adding method (and finality) labels and emitting from every breaker:

 | | Metric | Type | Labels | Values | 
|---|---|---|---|---|
  | **Originally proposed** | `erpc_upstream_circuit_breaker_state` | Gauge | `project`, `vendor`, `network`, `upstream`  | `0` closed · `1` open · `2` half-open |
| **Suggested** | `erpc_upstream_circuit_breaker_state` | Gauge | `project`, `vendor`, `network`, `upstream`, **`method`**, **`finality`** | `0` closed · `1` open · `2` half-open |

Parikalp-Bhardwaj:
Hi, I'd like to work on this.

I noticed `main` already has `MetricUpstreamBreakerStateChange`, which records breaker transitions, but there still doesn't appear to be a gauge for the current breaker state.

Is `erpc_upstream_circuit_breaker_state` still desired in addition to the transition counter?

If yes, should it keep the original labels:

`project`, `vendor`, `network`, `upstream`

or follow the existing counter's lower-cardinality labels?

PR #1106, opened 2026-08-25
feat(telemetry): expose upstream circuit breaker state

Parikalp-Bhardwaj:
## Summary

Adds `erpc_upstream_circuit_breaker_state`, a Prometheus gauge that exposes the **current state** of each configured upstream circuit breaker.

## Problem

eRPC already exposes:

```text
erpc_upstream_breaker_state_change_total
```

which records circuit-breaker **state transitions**.

That metric is useful for answering questions such as:

> How often is this breaker opening or recovering?

But it does not directly answer:

> What state is the breaker in right now?

Operators currently have to reconstruct the current state from transition counters, which is unreliable across process restarts and counter resets and makes direct alerting unnecessarily difficult.

## Implementation

This PR adds:

```text
erpc_upstream_circuit_breaker_state
```

as a Prometheus `GaugeVec`.

The metric reports:

| Value | State |
| ---: | --- |
| `0` | closed |
| `1` | open |
| `2` | half-open |

Labels:

```text
project
vendor
network
upstream
category
finality
```

`category` and `finality` identify the owning failsafe entry's `matchMethod` / `matchFinality`.

This is necessary because one upstream can have multiple independent circuit breakers for different method/finality scopes. Collapsing them into a single per-upstream series would allow independent breakers to overwrite each other's state.

## Design

Circuit-breaker semantics remain unchanged.

The generic `failsafe` package stays telemetry-independent:

```text
failsafe.Breaker
        ↓
OnTransition
        ↓
upstream breakerStateReporter
        ↓
Prometheus transition counter + state gauge
```

The existing transition hook is extended through a small upstream telemetry adapter rather than adding Prometheus dependencies to the breaker state machine.

The exported numeric values use an explicit state-to-value mapping instead of relying on the internal `failsafe.State` enum ordering.

### Initial state

The gauge is seeded when `SetNetworkConfig` runs, after the upstream's metric identity is complete.

This ensures an untripped breaker is visible immediately as:

```text
state = 0
```

without generating a fake transition or incrementing the existing transition counter.

### Async transition callbacks

`Breaker.OnTransition` callbacks are asynchronous.

The reporter therefore publishes the breaker's current `State()` rather than blindly trusting the callback's `to` argument, so delayed transition callbacks converge on the actual current breaker state.

## Existing metric compatibility

The existing transition metric remains unchanged:

```text
erpc_upstream_breaker_state_change_total
```

Its name, type, labels, and semantics are preserved.

The new gauge complements it:

```text
transition counter -> how often did the breaker change state?
state gauge        -> what state is the breaker in now?
```

For example:

```promql
erpc_upstream_circuit_breaker_state == 1
```

can now be used directly to detect currently open breakers.

## Scope

This change intentionally covers **upstream circuit breakers only**.

Cache-connector breakers are not included because #883 is upstream-scoped and cache connectors do not have the same project/vendor/network/upstream metric identity.

## Compatibility

- no circuit-breaker state-machine changes
- no threshold changes
- no selection-policy changes
- existing transition metric unchanged
- no request-derived or unbounded-cardinality labels added

Related #883.


upstream/upstream.go
```diff
@@ -271,6 +253,12 @@ func NewUpstream(
 	}
 	pup.networkLabel.Store("n/a")
 
+	// Wire circuit-breaker telemetry. Side-effect-only, so failsafe/ stays
+	// telemetry-free. Wired here (not while building the executors) because
```
aramalipoor, line 257: Any reason why not to have this in failsafe code so that everything is in 1 place?

Parikalp-Bhardwaj: I kept it here because the metric needs upstream-specific labels that aren’t available inside failsafe. The breaker still owns the state transitions and exposes them through OnTransition the upstream layer only translates those transitions into labeled telemetry, moving it into failsafe would require passing upstream-specific metadata into an otherwise generic package.

Parikalp-Bhardwaj: @aramalipoor That was my reasoning, but I am happy to move it if you prefer everything in one place. I will wait for your confirmation on the approach before changing it. 

I also noticed the branch now conflicts with the latest main, so I will resolve that before pushing the next update.

aramalipoor: I would say if we have access to the Upstream object in the failsafe module for CB let's extract the necessary labels there, but if there's no access to Upstream and passing it down would create bad/cyclic dependencies this can work.

Parikalp-Bhardwaj: Yes, that’s the case here, failsafe doesn’t have access to the upstream object, and passing it down would create a circular dependency, will keep the current approach, thanks for confirming

PR #1128, opened 2026-09-04
feat(metrics): unified metrics manager with metrics.customizations

andreclaro:
## Summary

Unifies metric registration behind one manager and adds `metrics.customizations` for exposure, per-family label, and per-family histogram-bucket control.

Implements the design in the spec PR (#1089).

All 141 families now `Define*` themselves unregistered and are registered once by `telemetry.Configure` from `erpc.Init`. Prometheus freezes a family's label-set hash on first registration (and keeps it after `Unregister`), so a label-set change — or dropping a family entirely — has to happen before that first registration. Registering at package init (the old promauto path) made that impossible; the define-then-register split is what lets a customization reach every family.

`metrics.customizations` is one ordered list of `subject` + optional `action`/`labels`/`buckets` entries. Matching is by specificity (exact > longer prefix > shorter prefix > `*`), not list order, so a broad drop plus a narrow keep carve-out reads the way an operator expects regardless of ordering. A dropped family is never registered, so it costs no series at all; the post-gather `FilteredGatherer` is retained only for the stock `go_`/`process_`/`promhttp_` collectors the manager doesn't own. The four legacy `histogram*`/`counter*` label knobs are deprecated but still work — they desugar into the same rule set.

### Notable follow-ups on this branch

- **Concurrency:** `rebuildInPlace` previously replaced `vec` and `activeIdx` with two unsynchronized field writes while readers accessed them locklessly. Both fields are now bundled into an immutable state published through a single `atomic.Pointer` store, with race tests for counter and histogram rebuild under concurrent observe+collect.
- Rebased onto current `main` (through the `eth_getLogs` response-size work).

## Test plan

- [ ] `go build ./...` and `go vet ./telemetry/ ./common/`
- [ ] `go test ./telemetry/ ./common/ -count=1` (includes rebuild race tests)
- [ ] CI race-enabled test job covers the concurrency tests (`-race` / cgo)
- [ ] Spot-check docs: `docs/pages/reference/metrics.mdx`, `operation/monitoring.mdx`, `operation/cli.mdx`
- [ ] Confirm TypeScript config types in `typescript/config/src/generated.ts` match the new schema


aramalipoor:
@andreclaro can you do some before/after bechmarks for critical Forward flow (the core proxy logic) for success/failures cases to make sure the new flow is not degredating performance (ops/s and allocations/s)? Especially the atomic/mutex locks and state tracking, to not cause memory churn/leak, or cause overhead per-request in various scenarios.

andreclaro:
> @andreclaro can you do some before/after bechmarks for critical Forward flow (the core proxy logic) for success/failures cases to make sure the new flow is not degredating performance (ops/s and allocations/s)? Especially the atomic/mutex locks and state tracking, to not cause memory churn/leak, or cause overhead per-request in various scenarios.

I compared `main` vs `feat/unified-metrics-manager` with the same benches (using `-count=5`):
- Recording APIs whose internals change (`WithLabelValues`, `CounterHandle`, `histogram Observe`): no extra allocs, ns/op in the noise.
- Forward success and failure (core proxy path): 330 / 499 mallocs/req on both sides. Parallel throughput with not a significant delta. Failure serial was a bit faster on the feature branch (−2.8%), probably noise.

Results: https://gist.github.com/andreclaro/be9b7c3334d0c77cc59183951c245708

aramalipoor (approved):

PR #1135, opened 2026-09-07
feat(evm): add servedTip lag cushion (lagBlocks / lag)

andreclaro:
Mixed eRPC pools (`type:internal` + `type:external`) often have internals a few blocks behind the strict-majority served tip. The advertised `latest` then lands on a block internals cannot serve, breaking consensus (requiredParticipants / composition disputes). This adds an optional lag cushion so the advertised tip can be pulled back to a height the slower members can still serve — without a hard per-group floor that a stuck cluster would use to pin the whole network.

## What changed

- `common/config.go` — `EvmServedTipConfig` gains optional `lagBlocks` (int64, default 0 = today's behaviour) and `lag` (duration, converted to blocks via EMA block time). A positive `lagBlocks` wins when both are set, matching the 0-means-unset convention of the other numeric servedTip fields.
- `erpc/networks.go` — after the existing pipeline (`PickServedTip` → trajectory referee → `guardServedTipRegression` → guaranteed-method floor) the advertised tip is clamped: `advertised = max(0, pick.Tip - lagBlocks)`. The subtract runs **last**, never before the regression guard, so a cushion can't look like a poisoned ballot, and it's applied after the referee's accounting so it neither masks nor triggers a referee alert. Applies to tags in `enabledFor`; no effect in max mode / when served tip is unset.
- `common/validation.go` — reject negative `lagBlocks` / `lag`.
- `docs/pages/reference/evm/block-tracking.mdx` `<AISection>` — schema/default, edge cases (clamp-to-0, block-time-unknown, lagBlocks precedence), observability. Neutral placeholders only.
- `typescript/config/src/generated.ts` — regenerated additive optional fields (no schema break).

## Verification

- `go test ./common/...` and `go test ./erpc/ -run ServedTip` — unit tests cover: unset = no-op; `lagBlocks: N` advertises majority − N; pick < N clamps to 0 (no negative / zero-poison); a lagging minority within N can serve the advertised block; a minority stalled beyond N does not pin the tip to that stall; and the `lag` duration → blocks path via the block-time override.

## AI assistance

- **Model:** anthropic/claude-opus-4.8
- **Harness / skills:** Claude Code (Symphony); Cursor (this upstream PR)
- **Human-reviewed:** no


aramalipoor:
Investigated this locally against `main`. The goal is right; the mechanism already exists in the codebase, keyed on method instead of tag. Draft alternative: #1166.

## 1. `guaranteedMethods` already clamps the tip to a subset's own majority

`guaranteedFloor` (was `guaranteedMethodFloor`, `erpc/networks.go:1505`) computes a per-subset ballot and folds the minimum into the advertised tip. It runs at the same point in the pipeline this PR inserts the subtraction. Only the membership predicate is method-shaped.

Verified on unmodified `main` — 3 externals at 1000, 2 internals at 993, `guaranteedMethods` pointed at a sentinel the externals `ignoreMethods`:

```
advertised = 993   (network majority is 1000)
```

Abusing `ignoreMethods` is a hack, not a recommendation. It is evidence that the clamp is the right shape and already built.

## 2. A fixed cushion is wrong in both directions

| internals behind | `lagBlocks: 7` | group majority |
|---|---|---|
| 2 | tip−7, **5 wasted** | tip−2 |
| 0 | tip−7, **7 wasted** | tip |
| 40 | tip−7, **still broken** | tip−40 |

`lagBlocks` also applies to the whole network on every axis in `enabledFor`, not only to requests the group must answer.

## 3. The stated objection to a per-group floor does not hold

> without a hard per-group minimum that a single stuck upstream could use to pin the whole network

The existing floor is the subset's **majority**, not its **minimum**. Verified — 3 internals, two at 993 and one frozen at 400:

```
advertised = 993   (the 400 is outvoted inside its own group)
```

A group of one is the real caveat, and it is the same caveat `guaranteedMethods` already carries for a single-supporting-upstream method.

## 4. `lag` is a no-op on cold start

`servedTipBlockTime()` → `EvmBlockTime()` returns 0 until the EMA warms, so `lag: "12s"` resolves to a **0-block cushion** exactly when a fresh pod starts serving — the moment the cushion is most needed. Verified: `EvmBlockTime()` is 0 on a freshly bootstrapped network.

Related: the only test covering that path sets `servedTipBlockTimeOverride`, which is dead on `main` (declared at `erpc/networks.go:79`, never read; its doc still refers to `buildServedTipConfig`, removed in 7dee07c0). So the production conversion is untested.

## 5. Alternative

#1166 — `servedTip.guaranteedFor: ["type:internal"]`. Reuses `guaranteedFloor` verbatim, swaps the predicate to an id/tag selector, same vocabulary `consensus.requiredParticipants` and `use-upstream` already use. No constant to size, tracks the real lag, scoped to the named group. CI green including the race suite.

Happy to be wrong on any of this — pointers above are all runnable.

andreclaro:
closed in favor of https://github.com/erpc/erpc/pull/1166

PR #1136, opened 2026-09-07
feat(cordon): persist cordon state via shared state

andreclaro:
## Summary

- Extends the shared-state backend so `erpc_cordonUpstream` / `erpc_uncordonUpstream` propagate cluster-wide and survive pod restarts / rolling deployments
- Cordon state stored as a single JSON map blob per upstream (`{clusterKey}/cordon-map/{projectId}/{upstreamId}:methods`) with a distributed lock on writes
- Cross-replica propagation reuses the existing counter pub/sub pipeline; replicas also run a ~30s periodic reconcile when pub/sub is unavailable
- Admin-driven and automatic cordons (consensus, health, identity) use **separate bits** in `TrackedMetrics` so neither path clobbers the other
- Selection policy `.removeCordoned()` reads the **effective** cordon reason (admin OR automatic), so admin cordons actually drop upstreams from routing
- Backward-compatible: gated on `sharedState` being configured; without it, behavior is unchanged

## Test plan

- [x] Unit tests: shared-state cordon registry, upstream admin persist/reconcile, tracker admin bits, `RemoveCordoned` for admin cordons
- [ ] Deploy with Redis shared state and verify cordon on replica A appears on replica B
- [ ] Restart a pod and confirm admin cordon restores from shared state
- [ ] Confirm `erpc_listCordoned` + `erpc_upstream_cordoned` + `erpc_selection_position=-1` after eval tick
- [ ] Confirm `X-ERPC-Use-Upstream` cannot force a cordoned upstream
- [ ] Deploy without shared state and confirm in-memory-only behavior


kasrakhosravi:
Reviewed and pushed a fix-forward commit (`c73bf34`) rather than a list of change requests — the problem is real and the plumbing was the part that needed rework. Diff is now +853/−137 (was +1247/−129).

**Why the problem is worth solving:** our prod fleet runs 4–10 proxy pods behind an HPA. `erpc_cordonUpstream` reaches one of them, and the cordon dies on the next scale-up or rollout. No config knob covers it.

**Regression that had to go first**

- `UncordonAdmin` cleared only the admin bit. The EVM chain-identity cordon (`architecture/evm/evm_state_poller.go:735`) has no auto-uncordon path, so after the change an operator could no longer lift it at all — on `main` they can.
- Now `erpc_uncordonUpstream` is an explicit override: it lifts every owner on the cell.

**What changed**

| Concern | Before | Now |
|---|---|---|
| Tracker state | `Cordoned` + `AdminCordoned` bit pairs, 7 fields, 7 methods, `adminCordonedMethods` | one `map[owner]CordonEntry` per cell; cordoned = `len > 0` |
| Storage | one key per upstream | one record per project: `<clusterKey>/cordons/<projectId>` |
| Propagation | counter pub/sub + one 30s goroutine per upstream | one 15s poll per project (`upstream/cordon_sync.go`) |
| Write failure | apply locally, keep a pending-mutation retry queue | persist first, fail closed, error to the operator |
| Memory driver | persisted | no store; in-process only |

**Why the pub/sub path was dropped:** the notify counter went through `GetCounterInt64(...).TryUpdate(time.Now().UnixNano())`, and `counterInt64.processNewState` rejects any remote state with `UpdatedAt <= currentTs`. Pod clock skew silently drops the notification, so the periodic reconcile was doing the real work. Polling one key per project also cuts the read load: ~620 upstreams × 10 pods was ~200 Redis GET/s for state that changes a few times a month.

**Why writes now fail closed:** applying a cordon locally after a failed write is what created the need for `pendingCordonMutations` — the queue existed only to stop the next reconcile undoing it. Persisting first means a pod never holds a cordon its peers will not, and the sync loop has nothing to undo.

**Why the memory driver no longer persists:** ristretto `Set` is asynchronous and may drop, so a sync loop reading it could clear a live cordon with nothing gained — there are no peers to share with.

**Verification**

- Unit: `health`, `data` (two registries over miniredis), `upstream` (two replicas, restart restore, auto-owner isolation, fail-closed), `internal/policy`, `consensus`, `architecture/*`, and the full `erpc` suite — green.
- End-to-end, two binaries + Redis + two fake upstreams: cordon on replica 1 appears on replica 2 within one 15s tick and shifts traffic (`erpc_selection_position=-1`); restarting a pod restores the cordon; uncordon from either replica clears it everywhere and emits the duration histogram; with Redis stopped `erpc_cordonUpstream` returns an error and applies nothing.

Docs updated for the owner model, the 15s convergence bound, and the fail-closed behaviour. Happy to split the tracker refactor into its own commit if you'd rather review it separately.

PR #1137, opened 2026-09-08
feat(cordon): share the operator cordon across replicas as a shared counter

kasrakhosravi:
## Summary

The whole-upstream operator cordon (`erpc_cordonUpstream`, `method: "*"`) becomes one shared counter per upstream — the same `CounterInt64SharedVariable` that already shares block heights — so it reaches every replica and survives restarts. Shared state is untouched. Supersedes #1136.

```
erpc_cordonUpstream ──> upstream.operatorCordon.TryUpdate(now)   (0 = lifted)
                                    │
             shared state: pub/sub + bootstrap fetch, as for latestBlock
                                    │
                        OnValue ──> tracker.Cordon/Uncordon(up, "*")   (existing cell)
```

- `upstream/upstream.go`: `operatorCordonVar()` (key `v2/operatorCordon/<project>/<upstream>/<hash>`, rollback threshold `0`), `CordonAdmin`, `UncordonAdmin`; `Uncordon("*")` refuses while the operator holds the shared cordon
- `internal/policy/eval.go`: `cordonedReason` read via `tracker.CordonedReason` (wildcard first), not the slot's metrics bucket
- `erpc/admin.go`: `"*"` → shared cordon; any other method → this replica only

Health tracker and shared state untouched. Non-test, non-doc delta: 4 files, +83 / −14.

## Behavior

| | before | after |
|---|---|---|
| Operator cordon on pod A | pod A only, lost on restart | every replica (ms on Redis), restored at bootstrap |
| Consensus penalty ending on a pod | cleared the operator cordon too | operator cordon untouched |
| Wildcard cordon under `evalScope: network-method` | ignored by `removeCordoned()` | excluded on every method slot |
| Re-cordon from another pod | n/a | keeps the shared value |
| `erpc_uncordonUpstream` (`"*"`) | this pod | every replica, automatic wildcard cordons included (one cell) |

Deliberately pod-local (an `int64` cannot carry them): method-scoped cordons, the reason string (peers show `"operator cordon (set on another replica)"`).

## Verification

- `upstream`: two replicas over miniredis converge in <100 ms, restart restores, detectors cannot lift an operator cordon, memory driver stays local
- `internal/policy/stdlib`: wildcard + method cordons excluded at all four `evalScope` values
- `erpc`: three full instances over one Redis; admin RPC on one, routing observed on the others via `eth_getBalance`; restart restore
- Live smoke: two `erpc` binaries + `redis-server` + mock upstreams, `curl`-driven — peer converged in <1 s, restarted pod restored before its first request

## Risk

- Shared counters push on change only, with no retry. If shared state is unreachable at the moment of the call, the cordon applies on that pod and is **not** re-pushed when the store returns (verified live: pods stayed divergent). Re-issue the call once the store is back. Same property as every other shared counter.


erpc/admin.go
```diff
@@ -672,10 +674,15 @@ func (e *ERPC) handleCordonUpstream(_ context.Context, nq *common.NormalizedRequ
...
+	switch {
+	case p.Method != "*" && cordon:
 		u.Cordon(p.Method, reason)
-	} else {
+	case p.Method != "*":
 		u.Uncordon(p.Method, reason)
+	case cordon:
+		u.CordonAdmin(ctx, reason)
+	default:
+		u.UncordonAdmin(ctx, reason)
```
aramalipoor, line 685: Any reason we'd need CordonAdmin? why not this to be a transparent behavior of Un/Cordon()?

kasrakhosravi: hmm Cordon(*) is also called by the consensus sit-out timer and the pollers, which uncordon on their own. If that went to shared state, a penalty expiring on one pod would clear the operator's cordon on every pod.

andreclaro (approved):

PR #1146, opened 2026-09-14
feat(consensus): tag request upstream IDs on Consensus.Run at start

andreclaro:
## Summary
- Copy the request's existing ordered upstream IDs onto `consensus.participants` at `Consensus.Run` start (before any slot returns).
- Mirror the same slice onto the parent `Network.Forward` span. End-of-run outcome / `upstreams.*` tags are unchanged.

## Test plan
- [x] `go test -count=1 -run TestConsensus_ParticipantsTaggedBeforeAnyReturn ./consensus/`
- [ ] In STG, confirm a completed (or caller-abandoned) `Consensus.Run` has `@consensus.participants` with the request's upstream IDs.


consensus/executor.go
```diff
@@ -143,8 +143,18 @@ func (e *executor) Run(
...
 
 	labels := e.extractMetricsLabels(ctx, originalReq)
+	parentSpan := trace.SpanFromContext(ctx)
 	ctx, consensusSpan := e.startConsensusSpan(ctx, labels)
 	defer consensusSpan.End()
+	if ups := originalReq.Upstreams(); len(ups) > 0 {
+		ids := make([]string, len(ups))
+		for i, u := range ups {
+			ids[i] = u.Id()
+		}
```
aramalipoor, line 153: Let's do this only if detailed tracing is enabled to avoid costly ops on basic tracing?

andreclaro: yes, agree!

PR #1159, opened 2026-09-20
fix(consensus): hash permanent missing-data for agreement

andreclaro:
Permanent `ErrEndpointMissingData` from different vendors can carry different JSON-RPC codes for the same skipped/absent slot. Hashing those by `NormalizedCode()` made 2-of-N `returnError` dispute them. Collapse the permanent class to one consensus hash; leave transient missing-data on the wire-code hash.

## Verification

```text
go test -count=1 -run 'ErrorToConsensusHash|ClassifyAndHash_PermanentSlotSkipped' ./consensus/
ok  	github.com/erpc/erpc/consensus	(4 tests)

go test -count=1 ./consensus/
ok  	github.com/erpc/erpc/consensus	167 tests
```

Also tested against live traffic: two vendors returning different skip codes for the same absent slot now agree under `disputeBehavior: returnError` instead of disputing.

## AI assistance

- **Model:** Cursor Grok 4.6
- **Harness / skills:** Cursor
- **Human-reviewed:** no


consensus/analysis.go
```diff
@@ -444,6 +444,17 @@ func errorToConsensusHash(err error) string {
...
+	// vendors emit different JSON-RPC codes for the same skipped/absent
+	// condition (e.g. Solana Alchemy -32007 vs QuickNode -32009). Hashing by
+	// NormalizedCode() disputes them under returnError. Collapse permanent
+	// missing-data into one class; leave transient missing-data (tip lag, etc.)
+	// on the wire-code hash. Do not rewrite NormalizedCode() — the original
+	// error still reaches the client when this group wins.
+	if common.HasErrorCode(err, common.ErrCodeEndpointMissingData) &&
+		common.IsPermanentlyMissingData(err) {
+		return "ErrEndpointMissingData:permanent"
+	}
```
aramalipoor, line 457: This feels super hacky, do we need a new error class?

andreclaro: For this particular case, I don't think we need a new error class here.

On SVM `error_normalizer` we are marking those two errors codes, `-32007` and `-32009`, as `WithPermanentMissingData.`:
- https://github.com/erpc/erpc/blob/main/architecture/svm/error_normalizer.go#L438-L447
- https://github.com/erpc/erpc/blob/main/common/errors.go#L2616-L2634

A new `ErrEndpointPermanentMissingData` error class would break every place that checks ErrCodeEndpointMissingData (retry-toward-network, upstream scoring, isAgreedUponError, the SVM normalizer), with no change the client would see.

Please do correct me if I'm missing anything.

CC: @kasrakhosravi

